PR #13, opened 2024-03-30
Reconfigure using the Okta Group Owners API to be opt-in

somethingnew2-0:
As surfaced by @gabrielsroka, the Okta Group Owners API is only available on Okta organizations with IGA enabled. As this feature in Access is not necessary unless you're using Otka Group Owner in IGA, disable it by default, but allow it to be opt-in if desired.

### DIFF
--- a/README.md
+++ b/README.md
@@ -46,10 +46,10 @@ CLIENT_ORIGIN_URL=http://localhost:3000
 REACT_APP_API_SERVER_URL=http://localhost:6060
 ```
 
-Then run the following commands to set up your python virtual environment:
+Then run the following commands to set up your python virtual environment. Access can be run with Python 3.10 and above:
 
 ```
-python3.11 -m venv venv
+python3 -m venv venv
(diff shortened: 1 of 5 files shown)


README.md
```diff
@@ -46,10 +46,10 @@ CLIENT_ORIGIN_URL=http://localhost:3000
 REACT_APP_API_SERVER_URL=http://localhost:6060
 ```
 
-Then run the following commands to set up your python virtual environment:
+Then run the following commands to set up your python virtual environment. Access can be run with Python 3.10 and above:
 
 ```
-python3.11 -m venv venv
```
ilyigna, line 52: why remove this? what if the host has installed 2 different python

somethingnew2-0: Removing this so we don't forget to update it for every new version of released Python (as what happened here). If someone has multiple python versions supported they can specify the desired version using a similar alias or an absolute path. Ideally, `python3` will point to a recent install if Python >= 3.10.

PR #23, opened 2024-04-09
Update user documentation and blog post link

eguerrant:
Added a new section to the FAQ of the user docs and corrected the blog post link


.vscode/settings.json
```diff
@@ -12,8 +12,8 @@
   "[python]": {
     "editor.formatOnPaste": false,
     "editor.codeActionsOnSave": {
-      "source.fixAll.ruff": true,
-      "source.organizeImports.ruff": true
+      "source.fixAll.ruff": "explicit",
+      "source.organizeImports.ruff": "explicit"
```
barborico, line 16: TIL these are functionally equivalent, and the string value is preferred in later VS Code versions.
https://code.visualstudio.com/updates/v1_85#_code-actions-on-save-and-auto

package-lock.json
barborico, line 1: Looks like `*` is equivalent to the latest version, so no change here.
https://docs.npmjs.com/about-semantic-versioning

barborico (dismissed):
LGTM. Seems like some of the automated changes are functionally equivalent, and the new content sounds good.

.vscode/settings.json
eguerrant: Oh those files weren't supposed to make it into the PR. Only the changes to Home.tsx. Gonna delete the changes

barborico (approved):
Sorry, I didn't mean that you should undo the automatic changes. I was just curious about them and so had done some research to confirm that the functionality was staying the same; it was just stated using more recent conventions.

PR #28, opened 2024-04-12
Access request and expiring access bugs

eguerrant:
Bug fixes:

- If you created a access request from a group's page, it would not apply time limit constraints (they would still be applied when the access request was approved/rejected so no biggie but still good to fix from a UX perspective)
  - Corrected to filter the until drop down and warn that there's a time limit constraint

- If you tried to renew a role's membership/ownership to a group you own that had the 'owners can't add themselves' constraint applied but used the row's 'Renew' button on the Expiring Access page instead of the 'Bulk Renewal' button, it would allow you to select the row but then throw an error when you tried to submit.
  - Corrected to not automatically select rows you shouldn't be able to renew

### DIFF
--- a/src/pages/requests/Create.tsx
+++ b/src/pages/requests/Create.tsx
@@ -188,17 +188,49 @@ const UNTIL_JUST_NUMERIC_ID_TO_LABELS: Record<string, string> = {
 
 const UNTIL_OPTIONS = Object.entries(UNTIL_ID_TO_LABELS).map(([id, label], index) => ({id: id, label: label}));
 
+function filterUntilLabels(timeLimit: number): [string, Array<Record<string, string>>] {
+  const filteredUntil = Object.keys(UNTIL_JUST_NUMERIC_ID_TO_LABELS)
+    .filter((key) => Number(key) <= timeLimit!)
+    .reduce(
+      (obj, key) => {
+        obj[key] = UNTIL_JUST_NUMERIC_ID_TO_LABELS[key];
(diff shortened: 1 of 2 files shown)


src/pages/roles/BulkRenewal.tsx
```diff
@@ -196,6 +189,33 @@ function BulkRenewalDialog(props: BulkRenewalDialogProps) {
     return out;
   }, new Set<string>());
 
+  const [selected, setSelected] = React.useState<RoleGroupMap[]>(() =>
```
somethingnew2-0, line 192: Can we add a few comments describing the filtering logic here and below?

src/pages/requests/Create.tsx
```diff
@@ -188,18 +188,50 @@ const UNTIL_JUST_NUMERIC_ID_TO_LABELS: Record<string, string> = {
...
+  );
+
+  return [Object.keys(filteredUntil).at(-1)!, filteredLabeles];
+}
+
 function CreateRequestContainer(props: CreateRequestContainerProps) {
   const navigate = useNavigate();
 
-  const [until, setUntil] = React.useState('1209600');
+  const [timeLimit, setTimeLimit] = React.useState<number | null>(
```
somethingnew2-0, line 215: Can you add a comment here describing what is being calculated for the time limit here?

somethingnew2-0 (approved):

somethingnew2-0 (approved):

PR #34, opened 2024-04-17
Implement a conditional request plugin interface so access requests can be automatically approved or denied

somethingnew2-0:
Includes a new plugin interface `ConditionalAccessPluginSpec` for allowing custom logic for conditionally automatically approving or denying access requests.

Breaking changes:
- `NotificationPluginSpec.access_request_created()` and `NotificationPluginSpec.access_request_completed()` were incorrectly sending the `OktaGroup.name` string as the `group` parameter instead of the `OktaGroup` object itself. This has been corrected so now the `OktaGroup` object is correctly sent to notification plugins for those functions.

---
  - https://app.asana.com/0/0/1207094387950525


jonathanhle:
Just dropping a note that this is 🔥

examples/plugins/access_requests/requests.py
```diff
@@ -0,0 +1,27 @@
...
+request_hook_impl = pluggy.HookimplMarker('access_requests')
+logger = logging.getLogger(__name__)
+
+@request_hook_impl
+def access_request_created(access_request: AccessRequest,
+                           group: OktaGroup,
+                           requester: OktaUser) -> Optional[ConditionalAccessResponse]:
+    """Auto-approve memberships to the Auto-Approved-Group group"""
+
+    if not access_request.request_ownership and group.name == "Auto-Approved-Group":
```
eguerrant, line 20: Will the groups have to be called 'Auto-Approved-Group' if we want to allow auto approval..? Maybe I'm missing how this fits in with the rest of the code...

eguerrant (dismissed):

examples/plugins/access_requests/requests.py
somethingnew2-0: This is just a toy example of auto-approving membership to a single group. We can change the plugin code to approve groups by their name or some other criteria.

eguerrant (approved):

PR #41, opened 2024-04-19
Add more rows per page options

eguerrant:
Added more rows per page options.

Before: 5, 10, or 20
Now: 5, 10, 20, 50, or all

### DIFF
--- a/api/pagination.py
+++ b/api/pagination.py
@@ -23,6 +23,8 @@ def extract_pagination(
 def paginate(query: SQLAlchemy, schema: Schema) -> Dict[str, Any]:
     page, per_page, other_request_args = extract_pagination(**request.args) # type: ignore[arg-type]
     # Make pagination index 0 based instead of 1 based
+    if per_page is -1:
+        per_page=query.count()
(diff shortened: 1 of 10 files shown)


api/pagination.py
```diff
@@ -23,6 +23,8 @@ def extract_pagination(
 def paginate(query: SQLAlchemy, schema: Schema) -> Dict[str, Any]:
     page, per_page, other_request_args = extract_pagination(**request.args) # type: ignore[arg-type]
     # Make pagination index 0 based instead of 1 based
+    if per_page is -1:
```
somethingnew2-0, line 26: We could probably have it default to no pagination if per_page is 0 instead of -1 and having it issue a COUNT query

somethingnew2-0: Or maybe we should just return all results in the API if per_page is None or unspecified (instead of the default of 50)

somethingnew2-0: Ah ok, let's keep `-1` meaning "All" in the UI and API then. We might be able to simplify the `paginate()` API code a  bit so that `query.paginate(page=page + 1, per_page=per_page)` isn't called unnecessarily, but it's fine for now.

PR #59, opened 2024-05-08
Create App with Role Owners

somethingnew2-0:
Allow list of `initial_owner_role_ids` to be specified during App creation (a POST to the /api/app endpoint)


goakley (approved):
Looks lovely, thank you!

PR #22, opened 2024-04-08
[fix: Issue #11] handle 429 retry with value from X-Rate-Limit-Reset

frenchi:
A **untested** attempt at resolving https://github.com/discord/access/issues/11 "use Okta's response headers for rate limits".

I ran `tox -e test` and the linter passed, but I don't have an Okta API key handy to test the actual 429 behaviour. Happy for this to be the first attempt that someone else refactors/takes over.

Issue #12, opened 2024-03-28
retry on resp.next()


gabrielsroka:
more converstaions with @somethingnew2-0

i noticed this "pattern" 4 times (paginating thru a given api)
https://github.com/discord/access/blob/7fda4989242204591dca97718ed3af4cf8582614/api/services/okta_service.py#L96-L105
don't u need to `retry` the `next()`, too?

i use async generators for it. write it once, factor it out, reuse it everywhere. something like

```py
import asyncio
from okta.client import Client as OktaClient

async def get_applications():
    async with OktaClient() as okta_client:
        filter = {'filter': 'status eq "ACTIVE"'}
        async for app in get_objects(okta_client.list_applications(filter)):
            print(f"Application Name: {app.name}")
            print(f"Application ID: {app.id}")
            async for app_user in get_objects(okta_client.list_application_users(app.id)):
                print(app_user.id)
            print()

async def get_objects(coro):
    objects, resp, _ = await coro
    while objects:
        for object in objects:
            yield object
        objects, _ = await resp.next() if resp.has_next() else (None, None)

asyncio.run(get_applications())
```

but that might not work here. i wrote it a while ago. maybe i should rewrite it

somethingnew2-0:
Resolved in https://github.com/discord/access/pull/63

PR #63, opened 2024-05-16
Add retry logic and timeout

### LINKED ISSUE #11: use Okta's response headers for rate limits
transcript of (Gabriel's half of) a conversation with @somethingnew2-0 

why do u use exponential backoff when Okta tells you when u can retry the api call?
https://github.com/discord/access/blob/40d8c0b2e19a0783430cfdac0834c4d893356cc0/api/services/okta_service.py#L41-L66

if u get a 429 error that tells u when to retry, why not look at those headers?
eg (and this isn't perfect, but...)
https://github.com/gabrielsroka/gabrielsroka.github.io/blob/master/console/index.html#L169-L187

ie, if u reach the rate limit at 10:00:00 and it tells u to retry at 10:01:00, there's no point in retrying at 10:00:01.2, 10:00:02.4, 10:00:04.8. ur just gonna get more errors

> Okta provides three headers in each response to report on both concurrent and org-wide rate limits.
For org-wide rate limits, the three headers show the limit that is being enforced, when it resets, and how close you are to hitting the limit:
X-Rate-Limit-Limit - the rate limit ceiling that is applicable for the current request.
X-Rate-Limit-Remaining - the number of requests left for the current rate-limit window.
X-Rate-Limit-Reset - the time at which the rate limit resets, specified in UTC epoch time (in seconds).

https://developer.okta.com/docs/reference/rl-best-practices/#check-your-rate-limits-with-okta-s-rate-limit-headers

eguerrant:
Added in the code from #22 (idk how to work with forks) as well as adding a timeout resolve #11.

### DIFF
--- a/api/services/okta_service.py
+++ b/api/services/okta_service.py
@@ -112,7 +112,7 @@ class OktaService:
             assert users is not None and resp is not None
 
             while resp.has_next():
-                more_users, _ = await OktaService._retry(resp.next())
+                more_users, _ = await OktaService._retry(resp.next)
                 users.extend(more_users)
             return list(map(lambda user: User(user), users))
 
@@ -216,7 +216,7 @@ class OktaService:
             assert groups is not None and resp is not None
 
             while resp.has_next():
-                more_groups, _ = await OktaService._retry(resp.next())
+                more_groups, _ = await OktaService._retry(resp.next)
                 groups.extend(more_groups)
             return list(map(lambda group: Group(group), groups))
 
@@ -244,7 +244,7 @@ class OktaService:
             assert group_rules is not None and resp is not None
 
             while resp.has_next():
-                more_group_rules, _ = await OktaService._retry(resp.next())
+                more_group_rules, _ = await OktaService._retry(resp.next)
                 group_rules.extend(more_group_rules)
             return group_rules
 
@@ -259,7 +259,7 @@ class OktaService:
             assert users is not None and resp is not None
 
             while resp.has_next():
-                more_users, _ = await OktaService._retry(resp.next())
+                more_users, _ = await OktaService._retry(resp.next)
                 users.extend(more_users)
             return list(map(lambda user: User(user), users))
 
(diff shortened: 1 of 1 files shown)


api/services/okta_service.py
```diff
@@ -42,7 +44,7 @@ def initialize(self, okta_domain: Optional[str], okta_api_token: Optional[str],
     async def _retry(func: Callable[[Any], Any], *args: Any, **kwargs: Any) -> Any:
```
somethingnew2-0, line 44: Can we wrap any of the `await resp.next()`  with `_retry`?

https://github.com/discord/access/blob/main/api/services/okta_service.py#L104
https://github.com/discord/access/blob/main/api/services/okta_service.py#L208
https://github.com/discord/access/blob/main/api/services/okta_service.py#L236
https://github.com/discord/access/blob/main/api/services/okta_service.py#L251

That could possibly fix the disconnect errors we're seeing when listing members of large groups with multiple API response pages. That should also close out this open issue https://github.com/discord/access/issues/12

api/services/okta_service.py
```diff
@@ -42,7 +44,7 @@ def initialize(self, okta_domain: Optional[str], okta_api_token: Optional[str],
...
         """Retry Okta API requests with specific status codes using exponential backoff."""
         for attempt in range(1 + REQUEST_MAX_RETRIES):
-            result = await func(*args, **kwargs)
+            result =  await asyncio.wait_for(func(*args, **kwargs), timeout=30)
```
somethingnew2-0, line 47: Let's make the `30` a constant above `REQUEST_TIMEOUT = 30` or something similar

api/services/okta_service.py
```diff
@@ -101,7 +112,7 @@ async def _list_users() -> list[User]:
             assert users is not None and resp is not None
 
             while resp.has_next():
-                more_users, _ = await resp.next()
+                more_users, _ = await OktaService._retry(resp.next())
```
somethingnew2-0, line 115: I believe we need to pass in the function instead of it's execution to `_retry()`

```suggestion
                more_users, _ = await OktaService._retry(resp.next)
```

eguerrant: Oh good catch. Fixed it

somethingnew2-0 (approved):

PR #73, opened 2024-05-24
Require at least one entity when adding to a group

goakley:
Fixes a bug where the 'Add' button is clickable when there have been no entities selected, which for users does nothing and for roles causes the modal to get stuck in a submitting state indefinitely.

bug


somethingnew2-0 (approved):
:shipit:

PR #92, opened 2024-06-27
Replace ILIKE email lookups from IDP authentication with more specific case-insensitive match

somethingnew2-0:


barborico (approved):
Thanks for the quick fix!

PR #139, opened 2024-09-11
Add limited Okta connection pooling

barborico:
This PR introduces limited connection pooling for requests from Access to Okta. The intent is to avoid creating excessive connections, e.g. in case of Okta API degradation resulting in timeouts. Although the [aiohttp docs for `ClientSession`](https://docs.aiohttp.org/en/stable/client_reference.html#client-session) state that "it is suggested you use a single session for the lifetime of your application to benefit from connection pooling," this isn't feasible due to the synchronous nature of the Access Flask app. Instead, we introduce a session with each async event loop to call the Okta API. This should at least consolidate retries and paginated requests to a single connection.

### DIFF
--- a/api/services/okta_service.py
+++ b/api/services/okta_service.py
@@ -5,12 +5,14 @@ import logging
 from datetime import UTC, datetime
 from typing import Any, Callable, Optional
 
+import aiohttp
 import dateutil.parser
 from okta.client import Client as OktaClient
 from okta.models.group import Group as OktaGroupType
 from okta.models.group_rule import GroupRule as OktaGroupRuleType
 from okta.models.user import User as OktaUserType
 from okta.models.user_schema import UserSchema as OktaUserSchemaType
+from okta.request_executor import RequestExecutor as OktaRequestExecutor
 
 from api.models import OktaGroup, OktaUser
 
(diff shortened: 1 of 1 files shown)


api/services/okta_service.py
```diff
@@ -85,147 +87,187 @@ async def _retry(func: Callable[[Any], Any], *args: Any, **kwargs: Any) -> Any:
             await asyncio.sleep(wait_time)
 
     def get_user(self, userId: str) -> User:
-        user, _, error = asyncio.run(OktaService._retry(self.okta_client.get_user, userId))
+        async def _get_user(userId: str) -> User:
+            request_executor: OktaRequestExecutor = self.okta_client.get_request_executor()
```
somethingnew2-0, line 91: Can this block be abstracted into a separate method or moved into the `OktaService._retry()` helper to cut down a bit on the redundancy?


```
request_executor: OktaRequestExecutor = self.okta_client.get_request_executor()
async with aiohttp.ClientSession() as session:
    request_executor.set_session(session)
```

somethingnew2-0 (approved):

api/services/okta_service.py
barborico: I separated it from the `_retry` helper so that paginated operations could share a connection, but took your advice to create a context manager and helper that handles sessions.

somethingnew2-0 (approved):

PR #140, opened 2024-09-13
Update docs and change column order in Expiring Roles

eguerrant:
When the Access homepage docs were written, conditional access plugins didn't exist. Now that they do, it seemed worth adding a section on the homepage to give a brief overview of plugins and point to the examples in the Access git repo.

Screenshot 2024-09-13 at 10 15 44 AM

This pr also swaps the Group and Role columns on the Expiring Roles page to match the Expiring Groups page (entity expiring in the first column, group that the entity has membership to in the second column).


barborico (approved):

barborico (approved):

Westfargo0147 (commented):
Am I unbanned

PR #175, opened 2024-10-30
Role requests API

eguerrant:
Adding role requests (like access requests but so that role owners can request that their roles are added to a group).

This is just the API piece. The UI will be in a separate PR.

I have no clue how to fix the last failing test atm...


api/operations/approve_role_request.py
```diff
@@ -0,0 +1,151 @@
...
+        if requester is None or requester.deleted_at is not None:
+            return self.role_request
+
+        # Don't allow approving a request for an a deleted or unmanaged group
+        if self.role_request.active_requested_group is None:
+            return self.role_request
+        if not self.role_request.active_requested_group.is_managed:
+            return self.role_request
+
+        self.role_request.status = AccessRequestStatus.APPROVED
```
somethingnew2-0, line 83: Should this be done inside ModifyRoleGroups? That way whenever ModifyRoleGroups is called (eg. an admin adds the role to the group manually), the corresponding request is also automatically approved.

api/operations/create_role_request.py
```diff
@@ -0,0 +1,182 @@
...
+            .filter(OktaGroup.deleted_at.is_(None))
+            .filter(OktaGroup.id == (requested_group if isinstance(requested_group, str) else requested_group.id))
+            .first()
+        )
+
+        self.request_ownership = request_ownership
+        self.request_reason = request_reason
+        self.request_ending_at = request_ending_at
+
+        self.request_approvers = db.session.query()
```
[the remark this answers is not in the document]
somethingnew2-0: Oh, this is copied from CreateAccessRequest which is also unnecessary

api/views/role_requests_views.py
```diff
@@ -0,0 +1,22 @@
+from flask import Blueprint
+
+from api.extensions import Api, docs
+from api.views.resources import RoleRequestList, RoleRequestResource
+
+bp_name = "api-role-requests"
+bp_url_prefix = "/api/role_requests"
```
somethingnew2-0, line 7: We should use hyphens instead of underscores for URL paths

```suggestion
bp_url_prefix = "/api/role-requests"
```

tests/test_role_request.py
somethingnew2-0, line 1: Are tests still WIP?

api/operations/create_role_request.py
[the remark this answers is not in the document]
eguerrant: Oh yeah a decent chunk of this is copied from access requests then modified. Seems like the request approvers line actually isn't used at all so I removed it.

migrations/versions/6d2a03b326f9_role_requests.py
```diff
@@ -0,0 +1,130 @@
...
+    sa.Column('approval_ending_at', sa.DateTime(), nullable=True),
+    sa.Column('approved_membership_id', sa.BigInteger().with_variant(sa.Integer(), 'sqlite'), nullable=True),
+    sa.ForeignKeyConstraint(['approved_membership_id'], ['role_group_map.id'], name=op.f('fk_role_request_approved_membership_id_role_group_map')),
+    sa.ForeignKeyConstraint(['requested_group_id'], ['okta_group.id'], name=op.f('fk_role_request_requested_group_id_okta_group')),
+    sa.ForeignKeyConstraint(['requester_role_id'], ['okta_group.id'], name=op.f('fk_role_request_requester_role_id_okta_group')),
+    sa.ForeignKeyConstraint(['requester_user_id'], ['okta_user.id'], name=op.f('fk_role_request_requester_user_id_okta_user')),
+    sa.ForeignKeyConstraint(['resolver_user_id'], ['okta_user.id'], name=op.f('fk_role_request_resolver_user_id_okta_user')),
+    sa.PrimaryKeyConstraint('id', name=op.f('pk_role_request'))
+    )
+    with op.batch_alter_table('access_request', schema=None) as batch_op:
```
somethingnew2-0, line 44: Interesting, I'm not sure why all these existing columns had `nullable=True` in the initial migration https://github.com/discord/access/blob/main/migrations/versions/d6db40b0805d_initial_migration.py

somethingnew2-0:
Looks like the tests aren't passing quite yet, but it's getting close!

PR #181, opened 2024-10-31
TypeScript Improvements

chippydip:
Converted `groupBy` helper to be strongly typed. This exposed some type errors where it was being used (result was assumed to be a `Map` but was actually just a `Record`). It now uses a typed accessor function to extract the key so the unsafe `deepFind` is not longer needed.

### DIFF
--- a/src/pages/users/Read.tsx
+++ b/src/pages/users/Read.tsx
@@ -423,7 +423,7 @@ export default function ReadUser() {
   const user = data ?? ({} as OktaUser);
 
   const ownerships = groupBy(user.active_group_ownerships, (m) => m.active_user?.id);
-  const memberships = groupBy(user.active_group_memberships, (m) => m.active_user?.id);
+  const memberships = groupBy(user.active_group_memberships, (m) => m.active_group?.id);
 
   const showRemoveGroupFromRoleDialog = (removeGroup: PolymorphicGroup, fromRole: RoleGroup, owner: boolean) => {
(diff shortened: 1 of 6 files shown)


src/pages/users/Read.tsx
```diff
@@ -422,14 +422,8 @@ export default function ReadUser() {
...
 
-  const ownerships: Map<string, Array<OktaUserGroupMember>> = groupBy(
-    user.active_group_ownerships ?? [],
-    'active_group.id',
-  );
-  const memberships: Map<string, Array<OktaUserGroupMember>> = groupBy(
-    user.active_group_memberships ?? [],
-    'active_group.id',
-  );
+  const ownerships = groupBy(user.active_group_ownerships, (m) => m.active_user?.id);
```
somethingnew2-0, line 425:
```suggestion
  const ownerships = groupBy(user.active_group_ownerships, (m) => m.active_group?.id);
```

src/pages/users/Read.tsx
```diff
@@ -422,14 +422,8 @@ export default function ReadUser() {
...
+  const memberships = groupBy(user.active_group_memberships, (m) => m.active_user?.id);
```
somethingnew2-0, line 426:
```suggestion
  const memberships = groupBy(user.active_group_memberships, (m) => m.active_group?.id);
```

somethingnew2-0 (approved):

PR #182, opened 2024-11-01
Users: update role display to wrap on overflow

savathoon:
Before:
Screenshot 2024-11-01 at 11 13 30 AM

After:
Screenshot 2024-11-01 at 11 14 35 AM


somethingnew2-0:
Does this also need to be fixed on the Groups Read page?

savathoon:
I already added this fix on the groups read page, this is just taking the same styles and applying them to users. Idk where else this might happen but saw it on the users page.

somethingnew2-0 (approved):

PR #190, opened 2024-11-08
Filter 'expiring in a week' individual notifications

eguerrant:
Currently, expiring access notifications for individuals come in two forms. Access expiring tomorrow and access expiring in a week. A while ago we added filtering for the 'expiring tomorrow' notifications to give preference to access via a role (users aren't notified about expiring individual access if they have the same access via a role) but did not add the same filtering logic to 'expiring in a week' notifications. This PR corrects that oversight.

### DIFF
--- a/api/syncer.py
+++ b/api/syncer.py
@@ -458,6 +458,12 @@ def expiring_access_notifications_user() -> None:
         .all()
     )
 
+    db_memberships_expiring_next_week = [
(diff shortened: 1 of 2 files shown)


barborico (dismissed):
Good catch! It might be helpful to include a comment that explains that and why these notifications are being filtered out.

PR #195, opened 2024-11-14
Add allow_discord_access custom Okta Group profile attribute check

jonathanhle:
Add `OKTA_GROUP_PROFILE_CUSTOM_ATTR` to may specify an Okta Group Custom Profile attribute used to override `is_managed_group` logic.

For the purposes of the tests and this comment, I've set `OKTA_GROUP_PROFILE_CUSTOM_ATTR` to `allow_discord_access`.  However, it is arbitrary and optional.  We'll use this ENV VAR value with regard to the following logic in this comment:

Verify the `allow_discord_access` custom Okta Group profile attribute to determine if Discord Access can manage the group: if true, management is enabled; if false, it is not.

**Why:**
* We have some Okta Groups that we want viewable in Discord Access, but do not want them to be manageable by it.
* We have some Okta Groups with a couple small Group Rules that we still want managed by Discord Access; membership rules won't step on each other between the existing Okta Group rules and Discord Access Access Requests

To support this, we're proposing a simple check under `is_managed_group` that looks for `allow_discord_access` in a Custom Okta Group profile attribute.
- if allow_discord_access is Undefined on the Okta Group, it'll default to the current logic in is_managed_group
- if allow_discord_access is True, return True - Discord Access can manage it (even if there are Group Rules)
- if allow_discord_access is False, return False - Discord Access cannot manage it - works to exclude any Okta Groups we don't want Discord Access to manage ever.

### DIFF
--- a/api/services/okta_service.py
+++ b/api/services/okta_service.py
@@ -535,4 +535,16 @@ class Group:
 
 
 def is_managed_group(group: Group, group_ids_with_group_rules: dict[str, list[OktaGroupRuleType]]) -> bool:
+    # Check if 'allow_discord_access' attribute exists as a custom Okta Group Profile attribute and retrieve its value
+    allow_discord_access = getattr(group.profile, "allow_discord_access", None)
+
+    # If 'allow_discord_access' is explicitly set to False, the group should not be managed
+    if allow_discord_access is False:
+        return False
+
+    # If 'allow_discord_access' is True and the group type is OKTA_GROUP, it can be managed even if it has group rules
+    if allow_discord_access is True and group.type == "OKTA_GROUP":
+        return True
+
+    # By default, the group should be of type OKTA_GROUP and should not have any group rules to be managed
     return (group.type == "OKTA_GROUP") and (group.id not in group_ids_with_group_rules)
--- a/api/services/okta_service.py
+++ b/api/services/okta_service.py
@@ -14,6 +14,7 @@ from okta.models.user import User as OktaUserType
 from okta.models.user_schema import UserSchema as OktaUserSchemaType
 from okta.request_executor import RequestExecutor as OktaRequestExecutor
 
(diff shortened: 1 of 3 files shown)


api/services/okta_service.py
```diff
@@ -535,4 +535,16 @@ def update_okta_group(
 
 
 def is_managed_group(group: Group, group_ids_with_group_rules: dict[str, list[OktaGroupRuleType]]) -> bool:
+    # Check if 'allow_discord_access' attribute exists as a custom Okta Group Profile attribute and retrieve its value
+    allow_discord_access = getattr(group.profile, "allow_discord_access", None)
```
somethingnew2-0, line 539: Not to bikeshed too much on the attribute name, but could we either name it something like `managed_by_access` or `access_is_authoritative`? Or better yet, could we have the attribute name be customizable with an environment variable (added [here](https://github.com/discord/access/blob/main/api/config.py)) in case other deployments already have an attribute they want to use?

somethingnew2-0 (commented):
Thanks for this addition!

api/services/okta_service.py
jonathanhle: Made some adjustments on my local.  Just trying to get my tests to work, but new ENV VAR for this sounds fine to me.

jonathanhle: Added a new optional parameter to `is_managed_group` so that I can mock it in the added tests.  Seems to work fine and all the tests pass. Spot checking on my local, functionality of the override works too.  

Resolving this conversation but welcoming further comments if any.

jonathanhle: New ENV VAR is `OKTA_GROUP_PROFILE_CUSTOM_ATTR`

jonathanhle:
@somethingnew2-0 made some updates and I think everything is passing tests.  Let me know how it looks after your review.

somethingnew2-0 (approved):

PR #199, opened 2024-11-20
Proposing Slack Notifications example

jonathanhle:
Notes
* Changing to node:22-alpine because of build issues after latest upstream container update: Logged a ticket to the node repo with the isssue: https://github.com/nodejs/docker-node/issues/2166
* Example readme with how this might be installed in a Docker container build
* Tested and working in my local environments

### DIFF
--- a/examples/plugins/notifications_slack/notifications.py
+++ b/examples/plugins/notifications_slack/notifications.py
@@ -16,7 +16,6 @@ logger = logging.getLogger(__name__)
 
 # Initialize Slack client and signature verifier
 slack_token = os.environ["SLACK_BOT_TOKEN"]
-signing_secret = os.environ["SLACK_SIGNING_SECRET"]
 client = WebClient(token=slack_token)
 alerts_channel = os.environ.get("SLACK_ALERTS_CHANNEL")
 CLIENT_ORIGIN_URL = os.environ.get("CLIENT_ORIGIN_URL")  # e.g. "https://discord-access-instance.com"
--- a/examples/plugins/notifications_slack/notifications.py
+++ b/examples/plugins/notifications_slack/notifications.py
@@ -1,0 +1,279 @@
+from __future__ import print_function
+
+import logging
+import os
+from datetime import date, datetime, timedelta
+from typing import List, Optional
+
+import pluggy
+from slack_sdk import WebClient
+from slack_sdk.errors import SlackApiError
+from slack_sdk.signature import SignatureVerifier
+
+from api.models import AccessRequest, OktaGroup, OktaUser, RoleGroup
+
+notification_hook_impl = pluggy.HookimplMarker("access_notifications")
+logger = logging.getLogger(__name__)
+
+# Initialize Slack client and signature verifier
+slack_token = os.environ["SLACK_BOT_TOKEN"]
+signing_secret = os.environ["SLACK_SIGNING_SECRET"]
+client = WebClient(token=slack_token)
+signature_verifier = SignatureVerifier(signing_secret)
+alerts_channel = os.environ.get("SLACK_ALERTS_CHANNEL")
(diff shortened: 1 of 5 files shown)


examples/plugins/notifications_slack/notifications.py
```diff
@@ -0,0 +1,279 @@
...
+from api.models import AccessRequest, OktaGroup, OktaUser, RoleGroup
+
+notification_hook_impl = pluggy.HookimplMarker("access_notifications")
+logger = logging.getLogger(__name__)
+
+# Initialize Slack client and signature verifier
+slack_token = os.environ["SLACK_BOT_TOKEN"]
+signing_secret = os.environ["SLACK_SIGNING_SECRET"]
+client = WebClient(token=slack_token)
+signature_verifier = SignatureVerifier(signing_secret)
```
mhintz-clickup, line 22: Where do you use this? I think this is only needed if slack send you back messages

examples/plugins/notifications_slack/notifications.py
```diff
@@ -0,0 +1,279 @@
...
+from slack_sdk.signature import SignatureVerifier
+
+from api.models import AccessRequest, OktaGroup, OktaUser, RoleGroup
+
+notification_hook_impl = pluggy.HookimplMarker("access_notifications")
+logger = logging.getLogger(__name__)
+
+# Initialize Slack client and signature verifier
+slack_token = os.environ["SLACK_BOT_TOKEN"]
+signing_secret = os.environ["SLACK_SIGNING_SECRET"]
```
mhintz-clickup, line 20: Where do you use this? I think this is only needed if slack send you back messages

mhintz-clickup: i removed this line and line 22; deployment worked fine

examples/plugins/notifications_slack/notifications.py
```diff
@@ -0,0 +1,279 @@
...
+def send_slack_dm(user: OktaUser, message: str) -> None:
+    """Send a direct message to a Slack user.
+
+    Args:
+        user (OktaUser): The user to send the message to.
+        message (str): The message content.
+    """
+    user_id = get_user_id_by_email(user.email)
+    if user_id:
+        mention_message = f"<@{user_id}> {message}"
```
mhintz-clickup, line 110: This is not needed. The App will hit the user just like a DM and if they have notifications setup they should still get it. I had mine set to `Direct messages, mentions & keywords` and I got notified. 

Would then need to change line 112 so that `text=message`; I made both adjustments and it worked fine for me

examples/plugins/notifications_slack/notifications.py
jonathanhle: Removed SignatureVerifier and the related unused var.  Thanks!

examples/plugins/notifications_slack/notifications.py
jonathanhle: So on the Slack instance I'm on, they wouldn't get the ding without the `@username`.  I kind of want this to be noisy, esp. on launch.  

I definitely added `@{user_id}>` because it wasn't working the way I wanted it too.  Okay if I leave it in?  We can make it configurable in the ENVARs if you find it annoying long term.

mhintz-clickup (approved):
Makes complete sense now why you had the @ symbol. 
I can verify this works great as I cloned down the plugin and implemented.

somethingnew2-0 (approved):
LGTM, let me know if you think this is ready to merge!

jonathanhle:
> LGTM, let me know if you think this is ready to merge!

I removed an unused ENV from the readme just a second again.  This is ready to go!  Thanks ahead for the merge.

somethingnew2-0 (approved):

Issue #37, opened 2024-04-18
request to expose Group Active tags to Conditional Access Plugin


jonathanhle:
I'd like to make some conditional access decisions based on the tags that are applied to an Okta Group within Access.

I have an exploratory branch [here](https://github.com/discord/access/compare/main...jonathanhle:access:jle-conditional-access-redux?expand=1) if it's useful.

```
    # Auto Approve if "AutoApproval" is applied to Group
    if not access_request.request_ownership and "AutoApproval" in group_tags:
        logger.info(f"Auto-approving access request {access_request.id} to group {group.name}")
        return ConditionalAccessResponse(
            approved=True,
            reason="Group membership auto-approved based on AutoApproval tag",
            ending_at=access_request.request_ending_at,
        )
```

somethingnew2-0:
PR to include this feature https://github.com/discord/access/pull/38

PR #200, opened 2024-11-21
feat(conditional-access): changed this to work with tags as well as group names

mhintz-clickup:
The goal of this is to add the ability to use Tags as well as groups. I know people will customize this but I thought this might give them a better base if they don't have complex use cases.

1. Changed the code to rely on environment variable instead of hard coded group names.
2. Added ability to use a tag for auto-approvals.
3. By default, the tag `Auto-Approve` is used. This means you don't have to set any environment variables if you don't want.
4. Added a Readme.

Verified this works in my deployment. Please let me know if a test is required for plugins

### DIFF
--- a/examples/plugins/conditional_access/Readme.md
+++ b/examples/plugins/conditional_access/Readme.md
@@ -1,0 +1,31 @@
+# Conditional Access Plugin
+
+This plugin will allow you to automatically approve or deny access requests based on the group or tag membership of the group.
+
+## Installation
+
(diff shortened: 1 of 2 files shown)


jonathanhle:
Looks fine to me.

somethingnew2-0 (approved):
Makes sense as follow-up to https://github.com/discord/access/issues/37, LGTM! 

Let me know if you're ready to merge this PR.

somethingnew2-0:
Going to merge

PR #194, opened 2024-11-13
Make the default time durations configurable on User Form Interface

jonathanhle:
We'd like customize the selectable time durations--including having a different default duration--when creating an Access Request.

Updated README with new REACT_APP Environment Vars introduced:
```
**Environment Default Overrides:**

- `UNTIL_ID_TO_LABELS_OVERRIDE`: Allows overriding the default ID to labels mapping. **Example:** `'{"14400":"4 Hours","43200":"12 Hours","432000":"5 Days","1209600":"Two Weeks","2592000":"30 Days","7776000":"90 Days"}'`.
- `UNTIL_JUST_NUMERIC_ID_TO_LABELS_OVERRIDE`: Allows overriding the default numeric ID to labels mapping. **Example:** `'{"14400":"4 Hours","43200":"12 Hours","432000":"5 Days","1209600":"Two Weeks","2592000":"30 Days","7776000":"90 Days"}'`.
- `DEFAULT_ACCESS_TIME_OVERRIDE`: Allows overriding the default access time. **Example:** `43200` (in seconds).

Defaults for these overrides can be found in the `src/env-overrides.tsx` file.
```

Credit to @Jui1337 who worked on the initial logic for this during her Internship on our team, but didn't manage to get this upstream yet.

I believe we've respected the current settings by setting them as the defaults in `src/env-overrides.tsx`.  

Let me know if you'd like any changes / if this can be accepted upstream.

PR #212, opened 2024-12-11
Make the default time durations configurable on User Form Interface

rktoomey:
This is a rewrite of https://github.com/discord/access/pull/194 after the discussion in the Discord channel at https://discord.com/channels/1232815453907189922/1306560215130443827.

This PR creates a default configuration file for the access app, and allows an environment-specific override configuration file path to be supplied as a build arg to Docker. See the README for details.

During the build process, craco uses webpack Define plugin to make the access config contents globally available as part of the static build artifacts.

cc @jonathanhle


savathoon:
🙏 this is incredible work thank you so much for fleshing out this proof of concept for dynamically loaded configs. 

I have some questions about specific parts of the approach, I don't consider these blocking concerns but am curious about some specific decisions around loading the config from the static folder instead of pre-populating the js with config values. I'll add more detail around specific implementations but this round trip to retrieve config values is my biggest current open question.

craco.config.js
```diff
@@ -15,5 +17,23 @@ module.exports = {
     alias: {
       '@mui/styled-engine': '@mui/styled-engine-sc',
     },
+    plugins: [
+      new CopyWebpackPlugin({
```
savathoon, line 21: It's been awhile since I've looked at webpack plugins, but I think instead of doing a blind `copy` of the config file into the build dir it might be possible to populate config values server-side and generate a "config" module that has the right values pre-populated from something like an `*.env` file. The benefits of doing it this way would be to save a round trip and make the initial time to render faster but if there are cons to this approach happy to discuss.

rktoomey: We want to be able to configure the access time dropdowns. The first PR (https://github.com/discord/access/pull/194) used environment variables in `.env.production`. My attempt uses a config file because in the discord discussion (https://discord.com/channels/1232815453907189922/1306560215130443827) the proposed revision was to use a config file that would be copied in with webpack.

> populate config values server-side

Unfortunately, I don't understand what you want :(
I am not well-versed in React so that might be the problem.

rktoomey: @savathoon please see my latest commit.

1. removed config files and config loader
2. create access_config.py to create a typed `AccessConfig` instance, looking for env vars that override individual properties.
3. created Flask endpoint in `app.py` to serve `/access_config.js`  
4. use `public/index.html` to load the `/access_config.js` script in every page
5. updated the front-end config loader to use `window.__ACCESS_CONFIG__` as the source of the config
6. updated the callsites

Please see what you think, because I'm not sure that I understand what you want.

somethingnew2-0 (changes requested):
We shouldn't include a roundtrip in every page load to load static config. This static config should be included in the JS bundle during build time.

rktoomey:
@somethingnew2-0 @savathoon 

I believe I've made the changes you requested. In this latest iteration. Please let me know what you think.

craco.config.js
```diff
@@ -1,4 +1,46 @@
...
+const webpack = require('webpack');
+
+// Load the default config
+const defaultConfigPath = path.resolve(__dirname, 'src/config/config.default.json');
+const accessConfig = JSON.parse(fs.readFileSync(defaultConfigPath, 'utf8'));
+
+// Parse environment variables
+let ACCESS_TIME_LABELS_OVERRIDE = null;
+try {
+  ACCESS_TIME_LABELS_OVERRIDE = process.env.ACCESS_TIME_LABELS_OVERRIDE
```
somethingnew2-0, line 13: Instead of using environment variables to override the default config (like in https://github.com/discord/access/pull/194), can we read from the override config file like you had before using the define plugin below?

somethingnew2-0: The override config file path can be set by an environment variable (like before) so that we know whether or not to include it here.

rktoomey:
Ok. Now you can set a build arg for Docker to use the config override file specified at ACCESS_FILE_CONFIG_PATH.

.env.production.example
```diff
@@ -12,4 +12,7 @@ REACT_SENTRY_DSN=https://<key>@sentry.io/<project>
 CLOUDFLARE_TEAM_DOMAIN=<CLOUDFLARE_ACCESS_TEAM_DOMAIN>
 CLOUDFLARE_APPLICATION_AUDIENCE=<CLOUFLARE_ACCESS_AUDIENCE_TAG>
 SECRET_KEY=<YOUR_SECRET_KEY>
-OIDC_CLIENT_SECRETS=<YOUR_CLIENT_SECRETS>
\ No newline at end of file
+OIDC_CLIENT_SECRETS=<YOUR_CLIENT_SECRETS>
+
+# ACCESS_TIME_LABELS_OVERRIDE='{"14400":"4 Hours","43200":"12 Hours","432000":"5 Days","1209600":"Two Weeks","2592000":"30 Days","7776000":"90 Days"}'
```
somethingnew2-0, line 17: Can we clean this up?

somethingnew2-0:
This is looking much better, thanks for adding this! 

I'm going to let @savathoon give it another review when they get a chance before I approve

.env.production.example
rktoomey: Done

docker-compose.yml
somethingnew2-0, line 1: Looks like linting is failing on this file. After running `npx prettier -w .` the fix looks like https://github.com/discord/access/commit/f846094ad56438daca6a45705734fc2fadc8afaa

rktoomey: Thank you, it should be fixed

jonathanhle:
> @savathoon do you have some time to look at this PR this week?

Let us know if there's anything additional to get this merged upstream.

Dockerfile
```diff
@@ -6,14 +6,17 @@ ARG PUSH_SENTRY_RELEASE="false"
 FROM node:22-alpine AS build-step
 ARG SENTRY_RELEASE=""
 WORKDIR /app
-ENV PATH /app/node_modules/.bin:$PATH
+ENV PATH=/app/node_modules/.bin:$PATH
 COPY craco.config.js package.json package-lock.json tsconfig.json tsconfig.paths.json .env.production* ./
 COPY ./src ./src
 COPY ./public ./public
```
savathoon, line 12: 🤔 since the config folder is located within `src` couldn't we just assume the files have been copied over instead of hardcoding it in the copy statement below?

.env.production.example
savathoon, line 1: This could be a solid place to define an env variable for config override path

src/config/config.empty.json
savathoon, line 1: Is this necessary?

src/config/accessConfig.ts
savathoon, line 1: I really like the idea of this file exporting a module. Some suggestions:
- Move some of the logic in `craco.config.js` here to create its final resolved key:value pairs. You could also add a custom attribute error handling logic like:
```
class UndefinedConfigError extends Error {
  constructor(key: string, obj: object) {
    const message = `'${key}' is not a defined config value in AccessConfig`;
    super(message);
    this.name = "UndefinedConfigError";
  }
}

// Function to safely access object keys
function getConfig<T extends object, K extends keyof T>(obj: T, key: K): T[K] {
  if (key in obj) {
    return obj[key];
  } else {
    throw new UndefinedConfigError(String(key), obj);
  }
}
```
instead of the sanity check currently in the config. 
- Instead of reading in raw JSON, have the config files be `.ts` and export an object that can be imported directly in the access config module. I'm not sure what the cleanest way to manage dynamic imports would be but so long as the logic is contained to the access config module and maybe a `process.env.*` variable I'd be ok with it.

savathoon (commented):
Sorry for the late turnaround, thank you for making the previous round of changes. Most of what I suggested is in the interest of simplification/consolidation but would love to get your thoughts.

rktoomey:
> Sorry for the late turnaround, thank you for making the previous round of changes. Most of what I suggested is in the interest of simplification/consolidation but would love to get your thoughts. 

Before I do any further work on this, I'd like to confirm that @somethingnew2-0 agrees with the changes requested in this review. 

Further, I'd like us to agree that the override config file will be kept in `src/config`. This way there can be an env var in `.env.production` without the need for a build arg to Docker to ensure that the override config file gets copied to the Docker image before craco runs. 

cc @jonathanhle

Dockerfile
somethingnew2-0: Since `ACCESS_FILE_CONFIG_PATH` is a Dockerfile `ARG` it's possible the file is specified elsewhere outside of `src` by the `docker build --build-arg` command

.env.production.example
somethingnew2-0: This place is the used by the flask environment and not used during build time

src/config/config.empty.json
somethingnew2-0: Probably not entirely necessary since it's empty, but not sure offhand if the Dockerfile can conditionally copy a file if it exists or not.

src/config/accessConfig.ts
somethingnew2-0: This sounds good to me, @savathoon perhaps you could fork this branch and create a commit demonstrating what you mean?

somethingnew2-0:
> Before I do any further work on this, I'd like to confirm that @somethingnew2-0 agrees with the changes requested in this review.

I've reviewed @savathoon's comments, they suggested to me privately taking at forking and suggesting a few commits to simplify things.
 
> Further, I'd like us to agree that the override config file will be kept in `src/config`. This way there can be an env var in `.env.production` without the need for a build arg to Docker to ensure that the override config file gets copied to the Docker image before craco runs.

That sounds good to me. IMO simplicity is preferred. So if we can remove the need for a build arg in the Dockerfile, all the better.

.env.production.example
rktoomey: Added ACCESS_CONFIG_FILE to .env.production.example

Dockerfile
rktoomey: Done.

README.md
```diff
@@ -248,28 +248,69 @@ If you are using Cloudflare Access, ensure that you configure `CLOUDFLARE_TEAM_D
...
+| `DEFAULT_ACCESS_TIME` | Specifies the default time access label to use for dropdowns on the front end. Contains a string with a number of seconds corresponding to a key in the access time labels. | `"86400"`                                                      | 
+
+The front-end config is loaded in [`craco.config.js`](craco.config.js). See
+[`src/config/loadAccessConfig.js`](src/config/loadAccessConfig.js) for more details.
+
+#### Backend Configuration
+
+To override values on the back-end, modify these key-value pairs inside the `BACKEND` key in your custom config file.
+
+_There are currently no supported back-end configuration overrides but this is planned to change soon._
```
jonathanhle: I'll be working on a following on PR, where I need to override some strings on the backend related to the regex name pattern used for Okta Group names.

config/config.default.json
```diff
@@ -0,0 +1,15 @@
...
+      "432000": "5 Days",
+      "1209600": "Two Weeks",
+      "2592000": "30 Days",
+      "7776000": "90 Days",
+      "indefinite": "Indefinite",
+      "custom": "Custom"
+    },
+    "DEFAULT_ACCESS_TIME": "1209600"
+  },
+  "BACKEND": {}
```
jonathanhle: I'll be working on a follow-on PR, where I need to override some strings on the backend related to the regex name pattern used for Okta Group names.

rktoomey: Actually, I added that in https://github.com/discord/access/pull/212/commits/d1968e5f989899df41ef301d357c01c61989d6c6 to show why having the same config file for front-end and back-end is useful (much like how we use `.env.production` for env vars in both React and Flask)

rktoomey:
@somethingnew2-0 @savathoon Review requested again.

As requested, the need for the Docker build arg is removed.
`craco.config.js` is tidied up: the logic for the access config load and validation has been moved to `src/config/loadAccessConfig.js`. Why js and not ts?  Well....

I tried to make the change @savathoon suggested, to convert the config to a Typescript object. This failed for multiple reasons:
- `craco.config.js` is CommonJS and can't deal with Typescript
- attempting to type the object to `AccessConfig` in `globals.d.ts` fails because it can't import that type from `src/config/accessConfig.ts`
- trying to covert the craco config to Typescript opened up a whole bunch of other problems, starting with that `react-app-alias` doesn't publish types

Whilst I was doing that, @jonathanhle and I spoke about his need to be able to configure the Access app on both the front-end and the back-end. His use case is simple: our pre-existing Okta entities use a different pattern to the one hard-coded in the Flask and React app.

So keeping the access config file JSON gives us the flexibility to do this neatly: now you can specify a pattern and a custom error message. I also wrote an access config file loader and tests.

Please re-review.

src/config/config.empty.json
rktoomey: Removed

src/config/accessConfig.ts
rktoomey: See comments at https://github.com/discord/access/pull/212#issuecomment-2589817397

jonathanhle:
> @somethingnew2-0 @savathoon Review requested again.
> 
> As requested, the need for the Docker build arg is removed. `craco.config.js` is tidied up: the logic for the access config load and validation has been moved to `src/config/loadAccessConfig.js`. Why js and not ts? Well....
> 
> I tried to make the change @savathoon suggested, to convert the config to a Typescript object. This failed for multiple reasons:
> 
> * `craco.config.js` is CommonJS and can't deal with Typescript
> * attempting to type the object to `AccessConfig` in `globals.d.ts` fails because it can't import that type from `src/config/accessConfig.ts`
> * trying to covert the craco config to Typescript opened up a whole bunch of other problems, starting with that `react-app-alias` doesn't publish types
> 
> Whilst I was doing that, @jonathanhle and I spoke about his need to be able to configure the Access app on both the front-end and the back-end. His use case is simple: our pre-existing Okta entities use a different pattern to the one hard-coded in the Flask and React app.
> 
> So keeping the access config file JSON gives us the flexibility to do this neatly: now you can specify a pattern and a custom error message. I also wrote an access config file loader and tests.
> 
> Please re-review.

Any chance this approach would work?  It solves a need for overrides on the frontend and backend and keeps the related overrides in the same spot.  We can maintain a slightly departed fork too for our needs, was just trying to avoid that.

somethingnew2-0:
This PR is looking good to me, I love how we can specify/validate the same config for both frontend and backend. ❤️ 

The current python segmentation fault test failure is unrelated and something I'm trying to track down with Github support. Tests pass locally for me.

Totally understand the issue of not being able to use Typescript during buildtime in the craco config.

I'm going to let @savathoon give a final pass for any opinions, but to me this is looking ready to ship.

jonathanhle (approved):
LGTM.  Works in my local testing.

savathoon (approved):
Looks good! Some open questions but nothing that's a blocking concern, will sanity check locally.

somethingnew2-0 (approved):

somethingnew2-0:
We fixed the unrelated test failures in https://github.com/discord/access/pull/229

Could you rebase or merge from main, so we can run the tests and then merge if they pass?

package.json
```diff
@@ -22,7 +22,8 @@
     "react-hook-form": "latest",
     "react-hook-form-mui": "latest",
     "react-router-dom": "latest",
-    "styled-components": "latest"
+    "styled-components": "latest",
+    "webpack": "latest"
```
somethingnew2-0, line 26: One last change request, can we move this dependency to the devDependencies list below? Then can we update `package-lock.json` by running `npm install`?

It should look like this
```
diff --git a/package.json b/package.json
index 488d302..2031c02 100644
--- a/package.json
+++ b/package.json
@@ -22,8 +22,7 @@
     "react-hook-form": "latest",
     "react-hook-form-mui": "latest",
     "react-router-dom": "latest",
-    "styled-components": "latest",
-    "webpack": "latest"
+    "styled-components": "latest"
   },
   "scripts": {
     "tsc": "./node_modules/.bin/tsc",
@@ -58,7 +57,8 @@
     "prettier": "^3.2.5",
     "react-app-alias": "^2.2.2",
     "react-scripts": "latest",
-    "typescript": "latest"
+    "typescript": "latest",
+    "webpack": "latest"
   },
   "lint-staged": {
     "*.{js,ts,jsx,tsx}": "prettier --write"
```

somethingnew2-0: Eh, I'll just do it in a quick follow-up PR

somethingnew2-0: https://github.com/discord/access/pull/231

somethingnew2-0:
Thank you so much for your contribution! We really appreciate it and thank you so much for bearing with us (especially for the delay over the winter break)! 😄

PR #219, opened 2025-01-03
Fix tag removal authorization bug in app groups

nonefaken:
There is a small authorization logic bug in `GroupResources` PUT method, which prevents removing `Tags` for app owner.

This patch fixes logic that only Access admin can remove app group `Tags`, even that app owner can add new `Tags` to owned app groups.

### DIFF
--- a/api/views/resources/group.py
+++ b/api/views/resources/group.py
@@ -118,7 +118,7 @@ class GroupResource(MethodResource):
             )
 
         json_data = request.get_json()
-        if "tags_to_remove" in json_data:
+        if "tags_to_remove" in json_data and type(group) is not AppGroup:
             if len(json_data["tags_to_remove"]) > 0 and not AuthorizationHelpers.is_access_admin():
                 abort(
                     403,


somethingnew2-0 (changes requested):
The current authorization logic for removing tags from groups is intentional. Only Access admins are allowed to remove tags from groups whereas any group owner can add tags to their group.

somethingnew2-0:
> Understood. @somethingnew2-0 can you suggest best way if i would like to implement custom authorisation logic, but still continue to use/contribute to original project code?

That's a tough question. Typically we've created [plugins specifications](https://github.com/discord/access/tree/main/api/plugins) so that folks can [add custom logic](https://github.com/discord/access/tree/main/examples/plugins) to certain parts of Access which could be done for the API authorization logic. However, that would be quite a large undertaking as there are quite a few API endpoints each with loosely shared authorization logic that would need be extracted for this refactor. Currently we don't have plans to build a plugin interface for custom authorization logic on API endpoints.

PR #266, opened 2025-05-22
Pull the membership `Chip` logic into a component

goakley:
Deeply nesting the logic for membership management inside the JSX templates made it hard to understand the meaning behind the conditionals. This change explicitly names those conditionals and reduces their duplication across the codebase.


somethingnew2-0 (approved):
Very nice, thanks for simplifying this!

PR #268, opened 2025-05-22
Adjusted dark mode

eguerrant:
Text to background contrast passes WCAG (Web Content Accessibility Guidelines) AAA for all background and text colors

Before:
Screenshot 2025-05-23 at 1 33 27 PM

After:
Screenshot 2025-10-24 at 2 05 22 PM

Light mode remains unchanged


barborico (approved):
I find the "after" picture harder to read, but approving since it passes objective accessibility guidelines--thanks for testing those. Ship it!

For future subjective style changes, maybe worth throwing a poll in #seceng-core to gauge user sentiment.

somethingnew2-0:
I agree that the text is a bit harder to read, is it possible to brighten the text a bit more so the contrast is a bit higher against the background?

PR #250, opened 2025-04-07
feat(access-requests): Added more specific audit logging for start-stop times

mhintz-clickup:
This pull request includes several enhancements to the `ReadRequest` component in the `src/pages/requests/Read.tsx` file. The most important changes involve adding timezone support, improving the display of access request statuses, and making layout adjustments.

Enhancements to timezone support:

* Added `timezone` and `utc` plugins to `dayjs` imports and extended `dayjs` with these plugins. [[1]](diffhunk://#diff-e5814227c64499f4816249470e59e36070316172d9d992dfad5e244b7ff69e3eR43-R44) [[2]](diffhunk://#diff-e5814227c64499f4816249470e59e36070316172d9d992dfad5e244b7ff69e3eR81-R85)
* Introduced a constant to get the user's local timezone.

Improvements to access request status display:

* Modified the status display to show different colors based on the status (`APPROVED`, `REJECTED`, or default).
* Added detailed information for `APPROVED` status, including access start and end times in both local and UTC formats. [[1]](diffhunk://#diff-e5814227c64499f4816249470e59e36070316172d9d992dfad5e244b7ff69e3eL456-R515) [[2]](diffhunk://#diff-e5814227c64499f4816249470e59e36070316172d9d992dfad5e244b7ff69e3eR587-R642)

Layout adjustments:

* Changed the height of the `Paper` component from a fixed value to 'auto' for better flexibility.

image

PR #274, opened 2025-06-03
Show expiration date on chip hover

eguerrant:
If a user has direct access and access via one or more roles, it can be unclear if one is granting longer access than the other since the 'Ending' column just shows the max. Added labels on hover to show expiration dates associated with each membership/ownership. Left the date format the same as on other pages (eg. the Access Request list page) to be consistent

Screenshot 2025-06-03 at 11 58 36 AM

### DIFF
--- a/src/components/MembershipChip.tsx
+++ b/src/components/MembershipChip.tsx
@@ -3,9 +3,10 @@ import {useNavigate} from 'react-router-dom';
 import {useCurrentUser} from '../authentication';
 import {canManageGroup, isGroupOwner} from '../authorization';
 import Chip from '@mui/material/Chip';
+import Tooltip from '@mui/material/Tooltip';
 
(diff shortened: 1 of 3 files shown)


savathoon:
nit - is the ending_date UTC? Could be good to append a timezone to the string to make it clear or localize based on viewer's current TZ

eguerrant:
> nit - is the ending_date UTC? Could be good to append a timezone to the string to make it clear or localize based on viewer's current TZ

All the dates shown in Access are in UTC. Fixing that everywhere would be good in the future but for now I feel like being consistent makes more sense

savathoon (approved):

somethingnew2-0:
We could probably use the code in https://github.com/discord/access/pull/250 to try and get time zone local timestamps working

PR #279, opened 2025-06-06
Diff users page

savathoon:
Screenshot 2025-06-05 at 8 04 17 PM
Screenshot 2025-06-05 at 8 09 06 PM

### DIFF
--- a/src/components/NavItems.tsx
+++ b/src/components/NavItems.tsx
@@ -164,6 +164,9 @@ export default function NavItems(props: NavItemsProps) {
           <ListItemLink to="/expiring-roles" displayText="All" displayIcon={<ExpiringAll />} sx={{pl: 4}} />
         </List>
       </Collapse>
+      <Divider />
+
+      <ListItemLink to="/users/diff" displayText="Diff Users" displayIcon={<GroupIcon />} />
     </List>
   );
 }
--- a/src/pages/users/DiffUsers.tsx
+++ b/src/pages/users/DiffUsers.tsx
@@ -4,6 +4,7 @@ import {
   Card,
   CardContent,
   Typography,
+  Link,
   List,
   ListItem,
   ListItemText,
@@ -18,6 +19,7 @@ import {
 } from '@mui/material';
 import {useGetAllUsers, useGetUserById} from '../../api/apiComponents';
 import {OktaUser, OktaUserGroupMember} from '../../api/apiSchemas';
+import {Link as RouterLink} from 'react-router-dom';
 
 const groupTypeLabels = {
   role_group: 'Role Group',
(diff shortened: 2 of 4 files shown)


src/pages/users/DiffUsers.tsx
```diff
@@ -0,0 +1,559 @@
...
+  Grid,
+} from '@mui/material';
+import {useGetAllUsers, useGetUserById} from '../../api/apiComponents';
+import {OktaUser, OktaUserGroupMember} from '../../api/apiSchemas';
+import {Link as RouterLink} from 'react-router-dom';
+
+const groupTypeLabels = {
+  role_group: 'Role Group',
+  app_group: 'App Group',
+  okta_group: 'Okta Group',
```
eguerrant, line 27: Technically they all groups (groups, app groups, roles) are  okta groups and we don't specifically call them okta groups anywhere else in the app so probably better to change this to just 'Group'. Also we made Access to be compatible with other idp providers besides Okta so users at other companies might get confused

src/pages/users/DiffUsers.tsx
```diff
@@ -0,0 +1,559 @@
...
+  const uniqueMembershipsUser1 = findUniqueMemberships(user1Details, user2Details);
+  const uniqueMembershipsUser2 = findUniqueMemberships(user2Details, user1Details);
+
+  const uniqueOwnershipsUser1 = findUniqueOwnerships(user1Details, user2Details);
+  const uniqueOwnershipsUser2 = findUniqueOwnerships(user2Details, user1Details);
+
+  return (
+    <Box>
+      <Typography variant="h4" component="h1" gutterBottom>
+        User Access Comparison
```
eguerrant, line 396: Can you make a title card to match the style of other pages? We don't have any other pages with a title directly on the background layer like that

src/pages/users/DiffUsers.tsx
```diff
@@ -0,0 +1,559 @@
...
+                            cursor: 'pointer',
+                          },
+                        }}
+                        target="_blank"
+                        rel="noopener noreferrer"
+                        component={RouterLink}>
+                        {groupName ?? 'Unnamed Group'}
+                      </Link>
+                    </Typography>
+                    <Chip
```
eguerrant, line 194: Idk why this is showing up as pink but it'd probably be better to stick to the theme colors? Also filled chips would probably be more readable

src/components/NavItems.tsx
```diff
@@ -164,6 +164,9 @@ export default function NavItems(props: NavItemsProps) {
           <ListItemLink to="/expiring-roles" displayText="All" displayIcon={<ExpiringAll />} sx={{pl: 4}} />
         </List>
       </Collapse>
+      <Divider />
+
+      <ListItemLink to="/users/diff" displayText="Diff Users" displayIcon={<GroupIcon />} />
```
eguerrant, line 169: I lean towards having this be linked from the users page as opposed to being in the navbar.

That said, all the other items in the nav bar have unique icons. If this stays here as opposed to being moved, probably change the icon to be consistent with everything else

src/pages/users/DiffUsers.tsx
```diff
@@ -0,0 +1,559 @@
...
+  const uniqueOwnershipsUser2 = findUniqueOwnerships(user2Details, user1Details);
+
+  return (
+    <Box>
+      <Typography variant="h4" component="h1" gutterBottom>
+        User Access Comparison
+      </Typography>
+
+      <Box display="flex" justifyContent="space-between" alignItems="center" mb={3}>
+        <Button variant="contained" size="large" sx={{height: '50px'}} onClick={resetComparison}>
```
eguerrant, line 400: A filled button would probably be more readable here

eguerrant (changes requested):

PR #281, opened 2025-06-09
Expiring Access do not renew

eguerrant:
Still maybe making minor changes (checking functionality for bugs atm) but more or less ready for review

This new feature makes it so that expiring access can be marked as reviewed without renewing the access. This allows the number of notifications to be reduced (we previously had no way to indicate that something had been reviewed so notifications for reviewed-but-not-renewed access would still be sent) and hopefully reduce conflicts that may come up where one group owner intends to allow access to expire while another owner of the group goes later and renews it anyways. 

The biggest change is that the bulk renewal dialogs now feature a Yes/No toggle instead of checkbox selection in order to indicate whether access should be renewed or allowed to expire. Access that has been marked as 'should expire' can still be renewed later if needed. 

The default page size for the expiring access pages is still 20 rows. I just reduced it to 10 for the sake of more compact screenshots.

Default expiring access page (doesn't show items marked as 'do not review')
Screenshot 2025-06-13 at 4 11 06 PM

View showing items marked as 'do not renew'
Screenshot 2025-06-13 at 3 28 21 PM

Bulk renewal dialog with new toggle (only showing unreviewed items)
Screenshot 2025-06-13 at 3 28 37 PM

Bulk renewal dialog showing all access (items marked 'do not renew' have 'No' in the toggle pre-selected)
Screenshot 2025-06-13 at 3 29 09 PM


api/operations/modify_role_groups.py
```diff
@@ -76,6 +78,22 @@ def __init__(
...
+            self.groups_should_expire = (
+                RoleGroupMap.query.filter(RoleGroupMap.id.in_(groups_should_expire))
+                .filter(RoleGroupMap.ended_at > db.func.now())
+                .filter(RoleGroupMap.is_owner.is_(False))
+            ).all()
+
+        self.owner_groups_should_expire = []
+        if len(owner_groups_should_expire) > 0:
+            self.owner_groups_should_expire = (
+                RoleGroupMap.query.filter(RoleGroupMap.id.in_(groups_should_expire))
```
savathoon:
```suggestion
                RoleGroupMap.query.filter(RoleGroupMap.id.in_(owner_groups_should_expire))
```

api/views/schemas/group_memberships.py
```diff
@@ -10,8 +10,10 @@ class GroupMemberSchema(Schema):
 
     members_to_add = fields.List(fields.String(validate=validate.Length(equal=20)), required=True, load_only=True)
     owners_to_add = fields.List(fields.String(validate=validate.Length(equal=20)), required=True, load_only=True)
-    members_to_remove = fields.List(fields.String(validate=validate.Length(equal=20)), required=True, load_only=True)
-    owners_to_remove = fields.List(fields.String(validate=validate.Length(equal=20)), required=True, load_only=True)
```
somethingnew2-0, line 14: Looks like `required=True` was removed from `members_to_remove` and `owners_to_remove`, was that intentional?

somethingnew2-0 (changes requested):
Add back `required=True` to `members_to_remove` and `owners_to_remove` in api/views/schemas/group_memberships.py

PR #286, opened 2025-06-13
Update README for performing DB migrations, add new indexes, update migration params

savathoon:


api/models/core_models.py
```diff
@@ -40,6 +40,16 @@ class OktaUserGroupMember(db.Model):
 
     created_reason: Mapped[str] = mapped_column(db.Unicode(1024), nullable=False, default="", server_default="")
 
+    __table_args__ = (
+        db.Index("idx_okta_user_group_member_user_id", "user_id"),
```
somethingnew2-0, line 44: Huh, I thought Foreign Keys were automatically indexed, but looking further I guess not?

https://stackoverflow.com/a/970605

> A foreign key must reference columns that either are a primary key or form a unique constraint, or are columns from a non-partial unique index. This means that the referenced columns always have an index to allow efficient lookups on whether a referencing row has a match. Since a DELETE of a row from the referenced table or an UPDATE of a referenced column will require a scan of the referencing table for rows matching the old value, it is often a good idea to index the referencing columns too. Because this is not always needed, and there are many choices available on how to index, the declaration of a foreign key constraint does not automatically create an index on the referencing columns.

https://www.postgresql.org/docs/current/ddl-constraints.html#DDL-CONSTRAINTS-FK

Good find!

PR #299, opened 2025-06-17
Notifications Update

eguerrant:
- Change the expiring access notification functions to provide more specific information (eg. list of users who will be losing access to each group instead of just a list of users and a list of groups)
  - The goal is to be able to send notifications along the lines of '<name> will be losing access to <group>' instead of the current notifications that are limited more to vague messages like 'A user will lose access next week'
  - Made this change backwards compatible for now so any messages that use the old args should still work
- Add notifications for role owners so that they know when roles they own will be losing access soon
- Updated example notifications

https://github.com/discord/access/pull/300 : changes to the Access UI so that role owners are easily able to see roles they own that will be expiring soon under 'Expiring Roles' and will allow them to make a role request from that page

### DIFF
--- a/api/syncer.py
+++ b/api/syncer.py
@@ -856,6 +856,7 @@ def expiring_access_notifications_role_owner() -> None:
         .join(RoleGroupMap.active_role_group.of_type(role_group_alias))
         .join(RoleGroupMap.active_group)
         .filter(db.and_(RoleGroupMap.ended_at >= day, RoleGroupMap.ended_at < next_day))
+        .filter(RoleGroupMap.should_expire.is_(False))
         .all()
     )
 
@@ -884,6 +885,7 @@ def expiring_access_notifications_role_owner() -> None:
         .join(RoleGroupMap.active_role_group.of_type(role_group_alias))
         .join(RoleGroupMap.active_group)
         .filter(db.and_(RoleGroupMap.ended_at >= day, RoleGroupMap.ended_at < next_day))
+        .filter(RoleGroupMap.should_expire.is_(False))
         .all()
     )
 


api/plugins/notifications.py
```diff
@@ -36,21 +37,29 @@ def access_request_completed(
 
     @hookspec
     def access_expiring_user(
-        self, groups: list[OktaGroup], user: OktaUser, expiration_datetime: datetime.datetime
+        self, groups: list[OktaUserGroupMember], user: OktaUser, expiration_datetime: datetime.datetime
```
somethingnew2-0: This and any other changes to existing hookspecs is going to be a breaking change to everyone's implementations of this notifications plugin. We should probably append new parameters/arguments instead of changing the type of existing ones.

api/plugins/notifications.py
somethingnew2-0, line 1: Can you ensure the example notification plugins are compatible with this change? https://github.com/discord/access/tree/main/examples/plugins

api/plugins/notifications.py
```diff
@@ -135,6 +156,21 @@ def access_expiring_owner(
         logger.exception("Failed to execute access expiring for owner notification callback")
 
 
+@hookimpl(wrapper=True)
+def access_expiring_role_owner(
```
somethingnew2-0, line 181: Can you add an example implementation in our example notifications plugins?

- https://github.com/discord/access/blob/main/examples/plugins/notifications_slack/notifications.py
- https://github.com/discord/access/blob/main/examples/plugins/notifications/notifications.py

api/plugins/notifications.py
```diff
@@ -122,9 +141,11 @@ def access_expiring_user(
 def access_expiring_owner(
     owner: OktaUser,
     groups: list[OktaGroup],
-    roles: list[RoleGroup],
+    roles: list[OktaGroup],
```
somethingnew2-0: Should this be changed to `OktaGroup` or should it stay as a `RoleGroup` list?

api/syncer.py
```diff
@@ -578,36 +597,43 @@ def expiring_access_notifications_owner() -> None:
...
-                owner_expiring_groups_next[owner].groups.append(group)
+                owner_expiring_groups_next[owner].groups[group] = non_owner_users
 
     for owner in owner_expiring_groups_this:
         # If the owner has members with access expiring both this week and next week, only send one message
         if owner in owner_expiring_groups_next:
             notification_hook.access_expiring_owner(
                 owner=owner,
-                groups=owner_expiring_groups_this[owner].groups + owner_expiring_groups_next[owner].groups,
+                groups=None,
```
somethingnew2-0: For backwards compatibility (here and other notification plugin calls) can we send the previous values instead of None?

api/syncer.py
```diff
@@ -578,36 +597,43 @@ def expiring_access_notifications_owner() -> None:
...
             notification_hook.access_expiring_owner(
                 owner=owner,
-                groups=owner_expiring_groups_this[owner].groups,
+                groups=None,
                 roles=None,
-                users=owner_expiring_groups_this[owner].users,
+                users=None,
                 expiration_datetime=datetime.now(),
+                group_user_associations=owner_expiring_groups_this[owner].groups,
+                role_group_associations=None,
```
somethingnew2-0, line 671: Instead of `None`, should this (and other similar notification plugin calls) be an empty dict `{}` to match the parameter type signature?

somethingnew2-0 (changes requested):

api/syncer.py
eguerrant: I set the new params as optional on the off chance someone has custom syncer logic so None should be fine

somethingnew2-0:
Voiced over, we should update the Slack example plugin as well
- https://github.com/discord/access/blob/main/examples/plugins/notifications_slack/notifications.py

We should also verify the example plugins install correctly as well as potentially verify the old version of the plugins ignore the new kwargs for backwards compatibility.

somethingnew2-0 (dismissed):

somethingnew2-0 (approved):

PR #300, opened 2025-06-19
Add role owner view for expiring roles page

eguerrant:
Prior to having role requests, it didn't make much sense to have a view under Expiring Roles that would show role owners the access expiring soon for roles they own because we could not make that knowledge actionable.

This new view shows role owners the roles they own with access expiring soon along with the ability to make a role request for that expiring access from that page (similar to how the 'My Access' tab under Expiring Groups allows users to make access requests for their expiring access).

Related to https://github.com/discord/access/pull/299 (adding role owner expiring access notifications)

Screenshot 2025-06-18 at 5 12 05 PM


src/components/NavItems.tsx
```diff
@@ -157,10 +158,16 @@ export default function NavItems(props: NavItemsProps) {
         <List disablePadding>
           <ListItemLink
             to="/expiring-roles?owner_id=@me"
-            displayText="Owned by Me"
+            displayText="Owned Groups"
```
savathoon, line 161: can you explain a little bit the distinction between /expiring-groups?owner_id=@me and /expiring-roles?owner_id=@me? I'm not clear what the difference is.

savathoon (dismissed):
One nit due to not sure what the intended behavior should be here but otherwise no issues from me

src/components/NavItems.tsx
somethingnew2-0: I believe the difference is:
- `/expiring-groups?owner_id=@me` is for group owners with individuals with expiring ownership/membership
- `/expiring-roles?owner_id=@me` is for role owners with groups with expiring ownership/membership

somethingnew2-0 (dismissed):

src/components/NavItems.tsx
savathoon: ah so something like "groups owned by a role I own"? I don't want to block on this but thanks for clarifying

eguerrant: I had a little trouble figuring out how to explain it in like less than 15ish characters (so it fit in the menu bar) but essentially : 

 /expiring-groups?owner_id=@me shows a _group owner_ when individuals' access to groups they own are expiring.

/expiring-roles?owner_id=@me shows a _group owner_ when roles' access to groups they own are expiring.

/expiring-roles?role_owner_id=@me shows a role owner_ when roles they own will be losing access to a group soon

eguerrant: Changing 'Owned by Me' under Expiring Roles to 'Owned Groups' and 'Owned Roles' was meant to differentiate the two cases since both involve ownership. I left 'Owned by Me' under Expiring Groups alone since that seemed clear enough there

PR #304, opened 2025-06-26
Introduce Access History for Group/Role Access Requests

tonydelanuez:
# Overview
This PR introduces a comprehensive access history feature for access requests by implementing user access history, role/group access history.

# Problem
Previously, reviewers had no visibility into whether a user or role had current or prior access to a group. This made it difficult to enforce least privilege and avoid redundant direct access.

# Solution
* Adds an access history section to access request review pages, showing current and past access for the requesting user or role to the target group.
* Displays both direct and role-based access, with clear summaries and empty states.
* Generalizes the access history logic into a single, context-aware component that adapts to both user-group and role-group requests.
* Standardizes UI/UX with collapsible cards, info icons, and clear, concise messaging.

# Examples
### User has no previous access
image

### Role has no current access/membership in group, but previously did
image


src/components/AccessHistory.tsx
```diff
@@ -0,0 +1,94 @@
...
+          <Typography variant="subtitle1" color="success.main" sx={{mb: 1, fontWeight: 500}}>
+            {`Current Access${currentAccess.length > 0 ? ` (${currentAccess.length} active membership${currentAccess.length !== 1 ? 's' : ''})` : ''}`}
+          </Typography>
+          <AccessHistoryTable
+            accessEntries={currentAccess as any}
+            highlightColor="success"
+            showEndedBy={false}
+            emptyMessage="No current access"
+          />
+          {/* Past Access Section */}
```
somethingnew2-0, line 79: Can we visually break apart the current and past access sections more? 

When current access is empty (like in the current screenshot), it take a bit of parsing to distinguish the two sections. 

Maybe a `<hr />` here would work?

somethingnew2-0:
```suggestion
          <hr />
          {/* Past Access Section */}
```

somethingnew2-0 (approved):

PR #313, opened 2025-07-02
Metrics - New api endpoint for metrics collection

savathoon:
Adds a new endpoint `/api/metrics` that accepts a POST request payload 

```
type: name of metrics plugin function to send data to
data: {
  name: metric name (optional)
  value: metric value (optional)
  duration: metric duration (for timing metrics, optional)
  tags: metric tags (optional)
  buckets: metric buckets (for histogram metrics, optional)
}
```
and passes through to the corresponding metrics plugin call for integrating frontend events with application metrics collection.

### DIFF
--- a/api/app.py
+++ b/api/app.py
@@ -143,7 +143,7 @@ def create_app(testing: Optional[bool] = False) -> Flask:
     @app.before_request
     def authenticate_request() -> Optional[ResponseReturnValue]:
         # Skip authentication for health check endpoint
-        if request.path.startswith("/api/healthz") or request.path.startswith("/api/metrics"):
+        if request.path.startswith("/api/healthz"):
             return None
         return AuthenticationHelpers.authenticate_user(request)
 


api/app.py
```diff
@@ -143,7 +143,7 @@ def create_app(testing: Optional[bool] = False) -> Flask:
     @app.before_request
     def authenticate_request() -> Optional[ResponseReturnValue]:
         # Skip authentication for health check endpoint
-        if request.path.startswith("/api/healthz"):
+        if request.path.startswith("/api/healthz") or request.path.startswith("/api/metrics"):
```
somethingnew2-0: If these metrics are meant to be scraped by a localhost Prometheus server we should possibly check that this request is coming via a 127.0.0.1/8 IP. Right now if someone has deployed access using the OIDC authentication mechanism on the internet, this metrics endpoint would be directly exposed.

somethingnew2-0: Ah never mind, this isn't for Prometheus metrics scraping, but for the frontend to report metrics to the backend. All frontend requests are authenticated, so this should be removed.

somethingnew2-0 (approved):

PR #318, opened 2025-07-11
Make blocked roles more clear

eguerrant:
Updated messaging around roles that cannot be added to a group due to tag constraints. In the 'Add roles' to group dialog, changed filtering out blocked roles to displaying them but making them unselectable with a message

Screenshot 2025-07-11 at 3 45 25 PM

### DIFF
--- a/src/pages/groups/AddRoles.tsx
+++ b/src/pages/groups/AddRoles.tsx
@@ -227,7 +227,22 @@ function AddRolesDialog(props: AddRolesDialogProps) {
     setRoles(roles.filter((user) => user.id != userId));
   };
 
+  // Determine if a role should be disabled
+  const isOptionDisabled = (option: RoleGroup) => {
+    // Already in the list to be added
+    if (roles.map((group) => group.id).includes(option.id)) {
+      return true;
+    }
+
+    if (disallowOwnerAdd && !isAccessAdmin(currentUser)) {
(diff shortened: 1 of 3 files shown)


src/pages/groups/AddRoles.tsx
```diff
@@ -245,9 +260,13 @@ function AddRolesDialog(props: AddRolesDialogProps) {
...
           </Typography>
           <Typography variant="subtitle1" color="text.accent">
-            {disallowOwnerAdd && !isAccessAdmin(currentUser)
-              ? 'Certain roles cannot be added to this group due to tag constraints.'
-              : null}
+            {disallowOwnerAdd && !isAccessAdmin(currentUser) ? (
+              <>
+                Certain roles cannot be added to this group due to the{' '}
+                <strong>owner can't add themselves as {ownerOrMember}</strong> tag constraint. Please get another owner
+                to add the role or, if you own the role, make a role request instead.
```
somethingnew2-0, line 267: > or create a role request if you own the role

It would be cool if we could show the Role request button instead of the text in this dialog to make the user experience more straightforward, but that might be out of scope for this PR

somethingnew2-0 (approved):

src/pages/groups/AddRoles.tsx
eguerrant: I'll add that to the backlog. Kind of out of scope since I intended this to be a quick update

PR #322, opened 2025-07-21
[metrics] Initial set of backend application metrics for Access

savathoon:


api/operations/approve_access_request.py
```diff
@@ -112,6 +114,30 @@ def execute(self) -> AccessRequest:
             )
         )
 
+        # Calculate approval time in seconds
+        creation_time = self.access_request.created_at
+        approval_time = datetime.utcnow()
+        resolution_time_seconds = (approval_time - creation_time).total_seconds()
+
+        # Record metrics for access request approval
+        group_type = "app_group" if isinstance(self.access_request.active_requested_group, AppGroup) else "role_group"
```
somethingnew2-0, line 123: There's also the third type of "vanilla" group named `OktaGroup` which is neither an app group or role group that should probably be accounted for here

Issue #324, opened 2025-08-04
Project name causes confusion


oliverba-unity:
Hi folks,

The current name makes it ambiguous and clumsy when we need to talk about the tool:

> Us: You can use access to request access to that tool
> Them: Huh? I can use access to get access? 
> Us: No, it's a separate tool called access, where you can request access to apps that are in Okta

It also makes it difficult for us to search for references to the tool, as "access" is a very common word in our documentation, Slack, emails etc.

Please consider either:

1. Renaming to a less-commonly used word, for example "Gatehouse"
2. Allow setting the application name via an environment variable

For option 2, I could only find a handful of places that would need the name substituting in. Here's a mockup:

<img width="3348" height="1726" alt="Image" src="https://github.com/user-attachments/assets/06280e1f-e3eb-4e08-ba36-cc1c6b1e7dcf" />

In any case, thanks for open-sourcing this great project!

oliverba-unity:
Awesome, thanks for responding so quickly.

I'll discuss with our team and see if we want to put together a PR for the environment variable name change.

Doing a find-and-replace in the codebase for "access" seems risky for the same reasons but let's see :D 

Thanks

PR #325, opened 2025-08-04
Move to using Vite instead of create-react-app

somethingnew2-0:
`create-react-app` is deprecated and no longer being maintained, move to a more supported build tool instead

https://react.dev/blog/2025/02/14/sunsetting-create-react-app

PR #326, opened 2025-08-13
feat: Make app name configurable

ianuriegas:
- Read app name from `REACT_APP_APP_NAME` environment variable (defaults to “Access”) via `src/config/accessConfig.ts`
- Use appName in document.title, meta description, sidebar header/title, references on `Home.tsx`
Screenshot 2025-08-13 at 11 30 16 AM

Ref: #324

### DIFF
--- a/tests/conftest.py
+++ b/tests/conftest.py
@@ -48,7 +48,9 @@ def db(app: Flask) -> Generator[SQLAlchemy, None, None]:
         _db.create_all()
 
     access_owner = OktaUserFactory.build(email=app.config["CURRENT_OKTA_USER_EMAIL"])
-    access_app = AppFactory.build(name=App.ACCESS_APP_RESERVED_NAME, description=f"The {App.ACCESS_APP_RESERVED_NAME} Portal")
+    access_app = AppFactory.build(
+        name=App.ACCESS_APP_RESERVED_NAME, description=f"The {App.ACCESS_APP_RESERVED_NAME} Portal"
+    )
     access_app_owner_group = AppGroupFactory.build(
         app_id=access_app.id,
         is_owner=True,


src/App.tsx
```diff
@@ -211,8 +213,14 @@ function Dashboard({setThemeMode}: {setThemeMode: (theme: PaletteMode) => void})
               textDecoration: 'none',
             }}>
             <Avatar src="/logo-square.png" variant="square" />
-            <Typography component="h1" variant="h5" sx={{px: 2}} color="text.accent">
-              ACCESS
+            <Typography
```
eguerrant, line 216: This will be cut off if the name selected is longer than a few characters. The original gh issue proposed Gatehouse as a title which would look like this:
Screenshot 2025-08-15 at 4 24 04 PM

somethingnew2-0:
Cosmetically this should be alright, however, there's a special reserved `Access` App and `App-Access-Owners` Group created when [setting up the database for the first time](https://github.com/discord/access/blob/main/api/manage.py#L103) that is used to determine the Access admins.
- Client https://github.com/discord/access/blob/main/src/authorization.tsx#L27
- Server https://github.com/discord/access/blob/main/api/authorization.py#L158

As is, this PR currently doesn't affect that group, but if we want the configured rename for Access to also affect that built-in group. We'd need to ensure that code using the `ACCESS_APP_RESERVED_NAME` constant in the client and the server is also updated.

api/config.py
```diff
@@ -86,3 +86,6 @@ def default_user_search() -> list[str]:
 
 # Add APP_VERSION, defaulting to 'Not Defined' if not set
 APP_VERSION = os.getenv("APP_VERSION", "Not Defined")
+
+# Specify a custom app name
+APP_NAME = os.getenv("REACT_APP_APP_NAME", "Access")
```
somethingnew2-0: Since this is a config for the Flask app (and not React), can we make it more unified? 

Additionally in #325, we removed the create-react-app dependency (which prefixed some environment variables with `REACT_APP`) with Vite. 

So maybe this environment variable should just be `APP_NAME`?

ianuriegas: Yup, that makes sense. I just replaced the existing variable to `APP_NAME`. Let me know if there is anything else that I need to change

somethingnew2-0:
```suggestion
APP_NAME = os.getenv("APP_NAME", "Access")
```

src/pages/Home.tsx
```diff
@@ -78,21 +73,17 @@ const guide: Record<string, Record<string, string>> = {
...
   },
   admin: {
-    'About Access administrators':
-      'Access administrators are the users who are members of the App-Access-Owners group. Administrators are able create or edit all groups, roles, and apps and manage the access for any group in the system. Administrators can also create tags, set tag constraints, apply tags to groups or apps, and remove tags.',
+    [`About ${appName} administrators`]: `${appName} administrators are the users who are members of the App-Access-Owners group. Administrators are able create or edit all groups, roles, and apps and manage the access for any group in the system. Administrators can also create tags, set tag constraints, apply tags to groups or apps, and remove tags.`,
```
somethingnew2-0:
```suggestion
    [`About ${appName} administrators`]: `${appName} administrators are the users who are members of the App-${appName} -Owners group. Administrators are able create or edit all groups, roles, and apps and manage the access for any group in the system. Administrators can also create tags, set tag constraints, apply tags to groups or apps, and remove tags.`,
```

somethingnew2-0:
This is looking good, there's just 
- one Python linting failure (we use ruff for Python linting with `tox -e ruff`)
- one additional spot to add an the appName in the documentation on the home page
- @eguerrant's comment above about the title bar overflowing with ellipsis for longer app names. If this isn't an issue for your planned internal app name, then we can skip it

ianuriegas:
- Python linting looks to be passing now on my end with the latest commit
- Just fixed with your suggestion
- Our app name internally is shorter than the initial proposed one and doesn't have the ellipsis shown on our end so should be good for us, but thanks for the callout

somethingnew2-0 (approved):

PR #330, opened 2025-08-25
fix: Bump pluggy to 1.5.0 for plugins and examples

ianuriegas:
Looks like pluggy needs to be bumped to [version 1.5.0](https://github.com/pytest-dev/pluggy/blob/main/docs/index.rst#warnings-on-hook-implementation) to use the [`warn_on_impl_args`](https://github.com/discord/access/blob/main/api/plugins/notifications.py#L38-L44) as intended. This was throwing some errors in our environment upgrading to Access from version 1.3.1 to 1.4.0.


somethingnew2-0:
Ah, thanks for catching this!

PR #350, opened 2025-10-24
Make app, group, and tag descriptions requirable

barborico:
# Overview of Changes

Descriptions are important for usability and auditability--let's let Access deployers make them required! This PR adds support for an environment variable `REQUIRE_DESCRIPTIONS` which, if set to `True`, will make descriptions required when creating/editing apps, groups, and tags.

## Front end environment loading

I made another change (the second commit) to how Vite loads environment so that it will pull variables set in `.env` files, but will still use the process env as the final source of truth. This was necessary to be able to test the frontend conditional form field requirements locally (or other config-based behavior like the web app name).

Before: Only read from shell/system environment
Now: Read from both .env files AND shell environment, with shell taking precedence

## Partial updates for apps, groups, and tags

To more consistently support Marshmallow's default loading without clearing properties omitted in a PUT request, I've updated the schema loading to apply partial updates.

# Testing

With `REQUIRE_DESCRIPTIONS=True`:
image

With `REQUIRE_DESCRIPTIONS=False` or omitting the variable altogether, the original behavior is preserved.

---
  - https://app.asana.com/0/0/1211884278397459


api/views/schemas/core_schemas.py
```diff
@@ -255,7 +255,7 @@ class OktaGroupSchema(SQLAlchemyAutoSchema):
             ),
         ),
     )
-    description = auto_field(load_default="", validate=validate.Length(max=1024))
```
somethingnew2-0, line 258: Just want to double-check, does this break serialization (dumping) when the database previously has empty descriptions? Do we need to add `dump_default=""`? https://marshmallow.readthedocs.io/en/stable/marshmallow.fields.html

eguerrant, line 258: Considering other companies use Access, I'm inclined to say that this should be put behind a feature flag/env variable. I imagine that other companies may not want to require descriptions and it could break their workflows to require them

barborico: The `description` columns in the DB aren't nullable, so I don't think it's an issue. But also no harm in adding that.

Claude's thoughts:
>  Even though the database columns have nullable=False and default="", adding dump_default="" ensures:
>  - Consistency: Matches the load_default="" behavior for symmetric serialization/deserialization
>  - Robustness: Handles any edge cases where the database might have NULL values (migrations, manual updates, etc.)
>  - Best Practice: Explicitly declares the serialization behavior rather than relying on implicit marshmallow defaults

barborico: Good call! Just added backwards compatibility and tested locally with `REQUIRE_DESCRIPTIONS` env var omitted (preserves existing behavior) or set to `True`.

api/config.py
```diff
@@ -89,3 +89,6 @@ def default_user_search() -> list[str]:
 
 # Specify a custom app name
 APP_NAME = os.getenv("APP_NAME", "Access")
+
+# Require descriptions for apps, groups, and tags
+REQUIRE_DESCRIPTIONS = os.getenv("REQUIRE_DESCRIPTIONS", "False") == "True"
```
eguerrant: Feel free to leave as is but since the bool is being ingested as a string, would we want to be more flexible and allow things like 'true', 'True', 'TRUE' etc?

eguerrant (dismissed):
If possible, please add tests for these changes

api/views/resources/app.py
```diff
@@ -131,7 +131,7 @@ def put(self, app: App) -> ResponseReturnValue:
                 abort(400, "Only tags can be modified for the Access application")
 
         old_app_name = app.name
-        app = schema.load(request.json, instance=app)
+        app = schema.load(request.json, instance=app, partial=True)
```
somethingnew2-0, line 134: Why was this and other examples in this PR updated to `partial=True`

barborico: To more consistently support Marshmallow's default loading without clearing properties omitted in a PUT request. When I added `load_default=""` to the App description schema for consistency, it broke a unit test that was using partial updates which emptied the description. Hence, https://github.com/discord/access/pull/350/commits/c206b4c32e87ab26b615d16a0364887033ecc9da.

somethingnew2-0 (approved):

api/config.py
somethingnew2-0: Yes, this and other boolean options in this file could be updated to 
```suggestion
REQUIRE_DESCRIPTIONS = os.getenv("REQUIRE_DESCRIPTIONS", "false").lower() == "true"
```

.github/workflows/ci.yml
```diff
@@ -43,6 +43,8 @@ jobs:
         run: tox -e test
       - name: Test with tox with postgresql
         run: tox -e test-with-postgresql
+      - name: Test with tox with require descriptions
```
somethingnew2-0: Instead of creating an entirely separate tox runner for a different edge-case we should probably use [pytest indirect parametrization](https://docs.pytest.org/en/latest/example/parametrize.html#indirect-parametrization) instead to set the environment variable on the `app()` fixture in https://github.com/discord/access/blob/main/tests/conftest.py#L34

It'd probably look something like
```
@pytest.fixture(scope="session")
def app(require_description: bool = False) -> Flask:
    load_dotenv(".testenv")
    if require_description:
       os.setenv("REQUIRE_DESCRIPTIONS", "True")
    app = create_app(testing=True)
    return app
 ```
 
 Then on some test like `test_put_app` in https://github.com/discord/access/blob/main/tests/test_app.py#L74
 ```
 @pytest.mark.parametrize("app", [False, True])
 def test_put_app(app: Flask,
    client: FlaskClient, db: SQLAlchemy, mocker: MockerFixture, access_app: App, app_group: AppGroup, tag: Tag
) -> None:
 ```

barborico: Oooh, that looks a lot cleaner. Thanks for the tip!

somethingnew2-0 (approved):

PR #376, opened 2026-02-28
Refactor user read page into categorized split-table layout

somethingnew2-0:
Screenshot 2026-05-20 at 2 29 39 PM

## Summary
- Reorganizes the user read page from two flat tables (Owner of Groups/Roles, Member of Groups/Roles) into three sections by group type: **Access Roles**, **App Groups**, and **Standard Groups**
- Each section shows side-by-side **Ownerships** and **Memberships** tables, replacing the previous separate full-width tables
- App Groups are further sub-grouped by app, with clickable app name headers linking to `/apps/{appName}`
- Sections are hidden when both ownerships and memberships are empty for that type
- Replaces the duplicated `OwnerTable`/`MemberTable` components with a single unified `GroupTable` component parameterized by an `owner` boolean
- Drops the **Type** column from tables since items are now categorized by section

### DIFF
--- a/src/pages/users/Read.tsx
+++ b/src/pages/users/Read.tsx
@@ -204,7 +204,7 @@ function GroupTable({
         pathParams: {groupId},
       });
     },
-    [putGroupUsers],
+    [putGroupUsers, owner, user.id],
   );
 
   return (
--- a/src/pages/users/Read.tsx
+++ b/src/pages/users/Read.tsx
@@ -30,14 +30,14 @@ import Typography from '@mui/material/Typography';
 import {useCurrentUser} from '../../authentication';
 import ChangeTitle from '../../tab-title';
 import {useGetUserById, usePutGroupMembersById} from '../../api/apiComponents';
-import {OktaUser, OktaUserGroupMember, PolymorphicGroup, RoleGroup} from '../../api/apiSchemas';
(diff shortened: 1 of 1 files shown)


src/pages/users/Read.tsx
barborico: The `useCallback` dep array is missing `owner` and `user.id`, both of which are captured from
props and drive the mutation body.

The old `OwnerTable`/`MemberTable` each hard-coded their API body, so the only closed-over
variable was `user.id` (already a pre-existing gap). The refactor adds `owner` as a new
closed-over prop value, making the stale-closure risk concrete: if React Query's
`putGroupUsers` object happens to be stable across a re-render where `owner` has changed,
the callback fires the wrong mutation — silently calling `owners_to_remove` when
`members_to_remove` was intended, or vice versa.

```suggestion
+  [putGroupUsers, owner, user.id],
```

barborico (commented):
Aesthetics look good to me. I hade Claude take a pass, and it found a but that seems legit.

src/pages/users/Read.tsx
somethingnew2-0: Good catch, fixed in 0d9f38a. Agreed the deps were incomplete, worth noting the stale-closure isn't reachable, today since each <GroupTable>, is rendered with a literal owner and a stable user.id from the route, but the fix is still correct and guards future refactors.

barborico (approved):

PR #406, opened 2026-04-01
Propagate app group and role prefixes

eguerrant:
Previously, reserved app group and role prefixes (by default 'App-' and 'Role-' were set in the back end but not propagated to the front end, which could lead front end group name validation to be incorrect if those values were changed from the default ones. Added these values to config so that they are set in one place and used everywhere


rashworld-max (approved):

api/models/core_models.py
```diff
@@ -602,8 +603,8 @@ class RoleGroup(OktaGroup):
 
 
 class AppGroup(OktaGroup):
-    APP_GROUP_NAME_PREFIX = "App-"
-    APP_NAME_GROUP_NAME_SEPARATOR = "-"
+    APP_GROUP_NAME_PREFIX: ClassVar[str]  # set from config at module load, see bottom of file
+    APP_NAME_GROUP_NAME_SEPARATOR: ClassVar[str]  # set from config at module load, see bottom of file
     APP_OWNERS_GROUP_NAME_SUFFIX = "Owners"
```
elaynelemos, line 608: Hey @eguerrant! I've addded issue https://github.com/discord/access/issues/382 (and PR https://github.com/discord/access/pull/388) for this customization. Is it possible to also support custom `APP_OWNERS_GROUP_NAME_SUFFIX`, `OKTA_GROUP_NAME_PREFIX`,  and configurable Okta group sync filter?

rashworld-max (approved):

rashworld-max (approved):

config/config.default.json
```diff
@@ -11,10 +11,16 @@
...
+    "ROLE_GROUP_NAME_PREFIX": "Role-"
   },
   "BACKEND": {
     "NAME_VALIDATION_PATTERN": "[A-Z][A-Za-z0-9-]*",
-    "NAME_VALIDATION_ERROR": "name must start capitalized and contain only alphanumeric characters or hyphens."
+    "NAME_VALIDATION_ERROR": "name must start capitalized and contain only alphanumeric characters or hyphens.",
+    "APP_GROUP_NAME_PREFIX": "App-",
+    "APP_NAME_GROUP_NAME_SEPARATOR": "-",
+    "ROLE_GROUP_NAME_PREFIX": "Role-"
   }
```
barborico, line 25: Why is this duplication necessary? Is it possible to have `FRONTEND` and `SHARED` config to reduce the likelihood of these being out of sync?

barborico (dismissed):

barborico:
> I _think_ this will require people to update their override files whether or not they are using non-default name prefixes, might be worth setting up a new release after this merges (since group requests have also been added recently) and noting this as a breaking change

Why is this? Do the overrides not merge with the defaults provided in `config/config.default.json`?

elaynelemos:
Hi folks. Is this feature still on the roadmap? This is relevant to ease Access adoption.

PR #411, opened 2026-04-09
Fix Okta retry timeout crashing group sync

somethingnew2-0:
## Summary
- **Fixed `_retry` timeout handling** in `okta_service.py`: The timeout handler created a 2-tuple `(None, e)` but callers expected 3 values, causing `ValueError: not enough values to unpack` when Okta rate-limited (429) and all retries timed out. Now raises a clear exception on final timeout instead of returning a mismatched tuple.
- **Added per-group error isolation** in `syncer.py`: A single group failure in `sync_group_memberships` or `sync_group_ownerships` no longer crashes the entire sync job — it logs the error, rolls back partial DB changes, and continues to the next group.
- **Added tests** for timeout-exhaustion, timeout-then-success, and group-failure resilience scenarios.

## Test plan
- [x] Existing retry tests pass (retriable/non-retriable status codes, no error)
- [x] New `test_retry_logic_all_timeouts_raises` verifies exception on exhausted timeouts
- [x] New `test_retry_logic_timeout_then_success` verifies recovery after transient timeout
- [x] New `test_membership_sync_continues_after_group_failure` verifies sync continues past failing group
- [x] Full test suite passes (253 tests)

---
  - https://app.asana.com/0/0/1213045704094187

### DIFF
--- a/api/services/okta_service.py
+++ b/api/services/okta_service.py
@@ -58,11 +58,11 @@ class OktaService:
             try:
                 result = await asyncio.wait_for(func(*args, **kwargs), timeout=REQUEST_TIMEOUT)
             except asyncio.TimeoutError as e:
-                logger.warning("Timeout on Okta request. Retrying...")
                 if attempt == REQUEST_MAX_RETRIES:
                     raise Exception(
                         f"Okta request timed out after {1 + REQUEST_MAX_RETRIES} attempts"
                     ) from e
+                logger.warning("Timeout on Okta request. Retrying...")
                 await asyncio.sleep(RETRY_BACKOFF_FACTOR * (2**attempt))
                 continue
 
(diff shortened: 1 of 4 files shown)


api/services/okta_service.py
```diff
@@ -59,7 +59,12 @@ async def _retry(func: Callable[[Any], Any], *args: Any, **kwargs: Any) -> Any:
                 result = await asyncio.wait_for(func(*args, **kwargs), timeout=REQUEST_TIMEOUT)
             except asyncio.TimeoutError as e:
                 logger.warning("Timeout on Okta request. Retrying...")
-                result = (None, e)
+                if attempt == REQUEST_MAX_RETRIES:
```
barborico, line 61: nit: This if block should go before the warning since it ends the retries.

barborico (dismissed):

api/services/okta_service.py
somethingnew2-0: Sounds good, fixed

PR #412, opened 2026-04-10
Downgrade Okta timeout errors to warnings

somethingnew2-0:
## Summary
- **Added `OktaTimeout` exception class** to distinguish timeouts from real API failures
- **Catch timeouts in async fire-and-forget methods** (`async_add/remove_user/owner_to/from_group`) — log at WARNING instead of propagating as unhandled task exceptions that Sentry captures
- **Split syncer.py catch blocks** — `OktaTimeout` logs at WARNING level, other errors remain at ERROR

Follows up on #411 which fixed the tuple-unpacking crash and added per-group error isolation.

## Test plan
- [x] `test_retry_logic_all_timeouts_raises` updated to assert `OktaTimeout`
- [x] Full test suite passes (253 tests)
- [x] ruff format/check passes

### DIFF
--- a/api/services/okta_service.py
+++ b/api/services/okta_service.py
@@ -28,6 +28,12 @@ REQUEST_TIMEOUT = 30
 logger = logging.getLogger(__name__)
 
 
+class OktaTimeout(Exception):
+    """Raised when an Okta API request times out after all retries."""
+
+    pass
+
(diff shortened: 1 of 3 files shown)


barborico (commented):
LGTM

PR #417, opened 2026-04-16
New home page with action bar

eguerrant:
Added an action bar to promp users when they have pending tasks. 

The tasks shown in the bar change depending on what is needed from the user (eg. if the user is not responsible for any pending access requests, that task will not be shown in the bar). When a section is clicked, the user will be navigated to the relevant page in the app (eg. to the "owned by me" page if groups they own have role access expiring soon). 

If there are no pending actions for the user, options to create a request are shown (which when clicked will open the request dialog on that page)

Files changed:
- Create pages: modifications to allow the request creation dialogs to be opened without the normal 'Create Request' button
- Home: all other changes to add in todo task bar

Full width action bar:
Screenshot 2026-04-20 at 1 38 21 PM

Narrow window example (number of tasks and rows change dynamically with window size):
Screenshot 2026-04-20 at 1 46 04 PM

Home page when no pending tasks:
Screenshot 2026-04-20 at 1 47 31 PM

Make a request action clicked opens dialog on the same page:
Screenshot 2026-04-20 at 1 48 03 PM

### DIFF
--- a/src/pages/Home.tsx
+++ b/src/pages/Home.tsx
@@ -605,8 +605,8 @@ export default function Home() {
       </Paper>
       <Paper>
         <Grid container spacing={2} direction="row" sx={{padding: 2}}>
-          <Grid item xs={3}>
-            <Grid container spacing={2} justifyContent="center" direction="column">
+          <Grid item xs={12} lg={3}>
+            <Grid container spacing={2}>
               <Grid item xs={12}>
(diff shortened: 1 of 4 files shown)


barborico (approved):
Feature LGTM! If you want feedback on the React, someone else or Claude's `pr-review-toolkit` plugin would probably be better.

PR #419, opened 2026-04-21
Toolbar role request fix

eguerrant:
Follow up to https://github.com/discord/access/pull/417

Previously, the option to create a role request in the toolbar was shown to anyone. Since role requests are limited to users who own at least one role, the option to create one from the toolbar should also be limited to role owners

### DIFF
--- a/src/pages/Home.tsx
+++ b/src/pages/Home.tsx
@@ -38,6 +38,7 @@ import {
   useGetRequests,
   useGetRoleRequests,
   useGetGroupRequests,
+  useGetRoles,
   useGetUserGroupAudits,
   useGetGroupRoleAudits,
 } from '../api/apiComponents';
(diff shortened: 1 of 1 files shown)


matthew-bass (approved):

src/pages/Home.tsx
```diff
@@ -382,6 +383,9 @@ export default function Home() {
   const {data: groupRequestsData} = useGetGroupRequests({
     queryParams: {assignee_user_id: '@me', status: 'PENDING', page: 0, per_page: 1},
   });
+  const {data: ownedRolesData} = useGetRoles({
```
somethingnew2-0, line 386: Shouldn't the /users/@me endpoint also show if there are any Groups (Roles) owned by the current user?

PR #421, opened 2026-04-24
[Group Requests] bug fix

eguerrant:
With group requests, an app owner could hijack ownership of any existing group in the system by setting resolved_group_name to that group's name when approving a request

Fixed group request approval so this is no longer possible and added tests

### DIFF
--- a/api/operations/approve_group_request.py
+++ b/api/operations/approve_group_request.py
@@ -1,7 +1,8 @@
 from typing import Optional
 
 from flask import current_app, has_request_context, request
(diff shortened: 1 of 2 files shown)


somethingnew2-0 (approved):

barborico (approved):
Thanks for fixing this! Two non-blockers:
1. Can the PR name be more descriptive?
2. Is there any feedback returned that explains why a request fails? It doesn't seem like it, which could create confusion.

PR #422, opened 2026-04-25
Should expire bug fix

eguerrant:
Arbitrary RoleGroupMap or OktaUserGroupMember IDs could be passed to the should_expire fields to mark unrelated access grants as "do not renew," with no check that the IDs belong to the role/group in the current request and no authorization check that the caller owns the affected group.

Added fix and tests

### DIFF
--- a/api/views/resources/role.py
+++ b/api/views/resources/role.py
@@ -135,9 +135,13 @@ class RoleMemberResource(MethodResource):
                 RoleGroupMap.id.in_(all_should_expire_ids),
                 RoleGroupMap.role_group_id == role.id,
             ).all()
-            for role_group_map in maps:
-                group = db.session.get(OktaGroup, role_group_map.group_id)
-                if group is not None and not AuthorizationHelpers.can_manage_group(group):
+            groups = (
+                db.session.query(with_polymorphic(OktaGroup, [AppGroup, RoleGroup]))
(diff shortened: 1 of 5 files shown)


api/views/resources/role.py
```diff
@@ -124,7 +124,22 @@ def put(self, role_id: str) -> ResponseReturnValue:
...
+        ):
+            all_should_expire_ids = group_changes.get("groups_should_expire", []) + group_changes.get(
+                "owner_groups_should_expire", []
+            )
+            maps = RoleGroupMap.query.filter(
+                RoleGroupMap.id.in_(all_should_expire_ids),
+                RoleGroupMap.role_group_id == role.id,
+            ).all()
+            for role_group_map in maps:
+                group = db.session.get(OktaGroup, role_group_map.group_id)
```
somethingnew2-0: This query for the OktaGroup SQLAlchemy object should be done outside the loop as it's an N+1 query that's easy to optimize. Copy the pattern below which gets alls the groups SQLAlechmey Objects to iterate over instead. 


```suggestion
           groups = (
                db.session.query(with_polymorphic(OktaGroup, [AppGroup, RoleGroup]))
                .filter(OktaGroup.deleted_at.is_(None))
                .filter(OktaGroup.id.in_([m.group_id for m in maps]))
            )
```

somethingnew2-0 (approved):

PR #14, opened 2024-03-31
Simplify docker

3UR:
I have made minor changes to simplify running this under a Docker container and have tested it.

- Created a example for `.env-production`
- Added support for `docker-compose` as it is way more simpler to get up and running locally, this does not conflict with any CI CD workflows (and only 1 command is needed to build and run the container!)
- Updated README.md to reflect the changes made

If anything else is needed before merging feel free to let me know

PR #425, opened 2026-04-27
Migrate from Flask + Marshmallow to FastAPI + Pydantic

somethingnew2-0:
## Summary

Replaces Flask + Flask-RESTful + flask_apispec + Flask-SQLAlchemy + Flask-Migrate + Marshmallow with **FastAPI + Pydantic v2 + plain Alembic**. SQLAlchemy stays sync; async is a deferred follow-up.

**Wire-compatible by design.** Same URL paths, methods, query params, request/response JSON shape, error envelope (`{"message": ...}`), and RFC822 datetime format. The React frontend and any external clients work unchanged. A new regression test pass was added for every previously-omitted field surfaced during QA.

### What's in the box

**Framework swap**
- OIDC ported from `flask_oidc` to Authlib + `SessionMiddleware`. Cloudflare Access JWT verification added (cached JWKS with thread-safe atomic refresh, `kid` rotation handling). Test/dev bypass via `app.dependency_overrides`.
- Existing models, operations, and Alembic migration files reused unchanged via a thin `db` compatibility shim (`db.Model` / `db.session` / `Model.query` / `first_or_404`). The shim is intentional scaffolding so the diff stays focused on the framework swap; removal is captured in `POST_MIGRATION_TODO.md` #1.
- New `Makefile` + `access` CLI (Click) replace `flask <cmd>` invocations. `gunicorn -k uvicorn.workers.UvicornWorker api.asgi:app` replaces `gunicorn api.wsgi:app`. K8s manifests and example plugins updated. `FLASK_ENV` → `ENV` everywhere.
- Standalone `alembic.ini` at repo root replaces the Flask-Migrate stub.

**Pydantic-first end to end**
- All 43 typeable routes declare a `response_model`; OpenAPI now publishes 80 typed components instead of advertising every payload as `{}`, unblocking the frontend codegen step (`POST_MIGRATION_TODO.md` #13).
- Routes return validated Pydantic instances directly; `paginate()` is generic on `pagination_cls: type[PaginationResponse[T]]`.
- All POST/PUT bodies validated via typed `Create*Body` / `Update*Body` / `Resolve*Body` schemas. `CreateGroupBody` / `UpdateGroupBody` / `CreateGroupRequestBody` are discriminated unions on `type`, restoring the per-type name-pattern + length + `REQUIRE_DESCRIPTIONS` validation that the original `*In` schemas did.
- All list/audit endpoints consume typed `Search*PaginationQuery` models via `Annotated[Model, Query()]`.
- Audit-log payloads modeled as explicit Pydantic refs (`_AuditUserRef`, `_AuditAppRef`, …) under `AuditLogPayload(extra="forbid")`, replacing the `__dict__` fallback that was leaking the OktaUser `profile` JSONB blob (HRIS data) into audit logs.

**Strict ORM serialization + eager-load hygiene**
- All routes run Pydantic validation in strict `from_attributes=True` mode, so unloaded relationships raise `InvalidRequestError` at the responsible route instead of being silently swallowed and emitted as `null`.
- The eager-load topology for `OktaUserGroupMemberDetail`, `RoleGroupMapDetail`, `OktaGroupTagMapDetail`, and `AppTagMapDetail`, plus the polymorphic `AppGroup`/`RoleGroup` loader, is centralized in `api/routers/_eager.py` and reused across `apps.py`, `groups.py`, `tags.py`, `users.py`, `role_requests.py`, and `access_requests.py` so the loader stays in lockstep with the response schemas.

**Auth gate**
- App-wide `Depends(require_authenticated)` runs on every registered route. Allowlist: `/api/healthz` and `/api/oidc/*` only — `/api/docs` and `/api/openapi.json` are intentionally inside the gate even though they're already DEBUG-gated.
- The SPA mount at `/` was replaced with a catch-all `GET /{spa_path:path}` route so static assets inherit the auth dependency. Path-traversal rejected; unmapped `/api/*` returns 404 instead of the SPA index.
- New `OIDCRedirectRequired` exception → 307 to `/api/oidc/login?next=…` so the SPA's `fetch()` follows the IdP login flow instead of seeing a bare 403.
- `request_id` is always uuid4 server-side; the previous `X-Request-Id` header read let a caller pin every concurrent request to one shared SQLAlchemy session-scope key.

### Tooling

- `Makefile` wraps install / migrate / init / run / sync / notify / pytest / pytest-postgres / ruff / mypy. README rewritten around `make` targets.
- `make pytest-postgres` spins up a disposable `postgres:16` container and runs the suite end-to-end against it. CI green on both backends.
- `tox -e ruff` + `tox -e mypy` clean. Strict typing on routers/schemas is intentionally deferred (`POST_MIGRATION_TODO.md` #14) to keep this diff scoped to framework swap + parity.
- `setup.py` renamed `name="api"` → `name="access"`; version bumped 0.2 → 2.0 to match the planned 2.0 release. Audit logger renamed `api.audit` → `access.audit`.

### Deferred follow-ups

See [`POST_MIGRATION_TODO.md`](./POST_MIGRATION_TODO.md) — 21 items grouped by surface area. Highlights:

- **#1** Remove the `db` compatibility shim (~150 lines, mostly mechanical).
- **#2/3** Async SQLAlchemy + `async_scoped_session` (gates #11, async Okta HTTP).
- **#5–8** Wire-shape changes that need frontend coordination (ISO 8601 datetimes, `exclude_none`, RFC 9457 errors, `fastapi-pagination`).
- **#13** Frontend OpenAPI codegen (now unblocked by the typed `response_model` work).
- **#20/21** Nonce-based CSP + trust `X-Forwarded-*` only from an allowlist (lifted from audit-report MED-4 / LOW-3).

## Test plan

- [x] `pytest tests/` — 281/281 passing (sqlite); 268/268 passing on Postgres via `make pytest-postgres`
- [x] `tox -e ruff` (check + format) clean; `tox -e mypy` clean (137 source files)
- [x] `alembic upgrade head` against scratch sqlite — schema matches `main`
- [x] `make run-backend` + manual smoke through the React UI: groups list, group detail (Okta/Role/App), access-request flow, group-request create/read/resolve, role-request resolve, audit pages (users/groups/expiring), `/apps/<name>` owner+non-owner groups, `/users/@me` group memberships, `/tags/<name>`
- [x] `/api/docs` Swagger UI loads under DEBUG CSP
- [x] OpenAPI now exposes 80 typed components on 43/47 endpoints
- [x] CF Access end-to-end smoke in a deployed dev env (real JWT + service token)
- [x] `npm run dev` regression sweep through groups/roles/apps/requests/audit
- [ ] Frontend OpenAPI codegen run (POST_MIGRATION_TODO #13) once this lands


POST_MIGRATION_TODO.md
```diff
@@ -0,0 +1,185 @@
...
+- `db.Column` → `mapped_column`
+- `db.relationship` → `relationship`
+- `db.func` / `db.or_` / `db.not_` / `db.and_` → direct `sqlalchemy` imports
+- `Model.query.filter(...)` → `db.session.query(Model).filter(...)` or
+  `session.execute(select(Model).filter(...))`
+
+**Net change:** removes ~150 lines of shim, makes the model layer indistinguishable
+from any other SQLAlchemy 2.0 project. Mostly mechanical.
+
+### 2. Switch to async SQLAlchemy
```
willk-discord, line 28: TBH if I could go back I would not use async sqlalchemy. At the time we moved to it to get the "async everything" benefits from FastAPI.... but in practice we have spent an _extremely high amount of time_ dealing with the _inane_ async contract SQLAlchemy provides. For an example:
```python
class TicketModel(SiriusBaseModel):
    __tablename__ = "tickets"

    vulnerabilities: Mapped[List["VulnerabilityModel"]] = relationship(
        "VulnerabilityModel",
        secondary=ticket_vulnerabilities,
        lazy="immediate",  # Don't lazy-load these to make the translation between `schemas` and `models` easier
        back_populates="tickets",
    )
    # .... snipped ....

    select: Select[TicketModel] = (
        Select(TicketModel)
        .limit(1)

   ticket: TicketModel = await db.scalar(select)

   print(ticket.tickets[0]). # WE BLOW UP THE STACK HERE. REALLY.
```

Any relationship is a lazy loaded attribute. Async sqlalchemy just ignores our directives for immediate loading. Any lazy loaded attribute must be explicitly re-loaded before access just in time using the async await pattern. Failure to do so results in a `greenlet not awaited` error. This has caused Sirius to go to the pattern of _immediately_ coercing from SQLAlchemy model -> Pydantic schema to avoid this cursed lazy loading. Other DB accesses can expire the lazy loaded attributes, especially when running in a green threaded context, which means that the safest option is to just either A) never use relationships B) always coerce to pydantic immediately  (our choice) or C) not do async sqlalchemy.

My overwhelming experience with it is that async sqlalchemy is not production ready. It actively lies about its API contracts, and does not uphold them. This has caused us to invest _heavily_ in integration tests because we cannot sanely reason about the different failure patterns.

willk-discord (dismissed):
This is a huge change set, so I'm not reviewing everything by hand. I looked at a sample of each major change set looking for patterns (expecting that LLMs are good at doing the same thing everywhere). This looks like a reasonable conversion to FastAPI! My main points of feedback are around not using Async SQLAlchemy if you can avoid it (not in this PR, noted as future work). Otherwise this seems idiomatic to me!

tophersmith:
IMO a big thing here is that you can use pydantic for POST/PUT body shapes, which didn't come through. The rest mostly looks fine so long as tests run/manual review looks ok. I did a big LLM-based review and it pulled out these 5 things, which look right:

---
1. POST /api/requests returns 500 instead of 400 for unmanaged groups (confidence: 100)
                                                                                                                                                                                                        
Bug site — ar is dereferenced without a None guard:
https://github.com/discord/access/blob/803e8e906bc3c31594a686b034fd948f27dea9d3/api/routers/access_requests.py#L78-L87                                                                                
                                                                                                                                                                                                        
Operation that returns None — when requested_group.is_managed == False:                                                                                                                               
https://github.com/discord/access/blob/803e8e906bc3c31594a686b034fd948f27dea9d3/api/operations/create_access_request.py#L62-L67                                                                       
                                                                                                                                                                                                        
Old behavior on main — upfront is_managed check with abort(400, ...):                                                                                                                                 
https://github.com/discord/access/blob/5a8e85f1ecfd76908d1637f910b51aa77b8231ed/api/views/resources/access_request.py#L373-L378                                                                       
                                                                                                                                                                                                        
Old belt-and-suspenders — None guard after .execute() returned:                                                                                                                                       
https://github.com/discord/access/blob/5a8e85f1ecfd76908d1637f910b51aa77b8231ed/api/views/resources/access_request.py#L403-L405                                                                       
                                                                                                                                                                                                        
---                                               
2. GET /api/requests dropped 5 query filters (confidence: 100)                                                                                                                                        
                                                                                                                                                                                                        
Bug site — only q is honored:       
https://github.com/discord/access/blob/803e8e906bc3c31594a686b034fd948f27dea9d3/api/routers/access_requests.py#L40-L52                                                                                
                                                                                                                                                                                                        
Old filter logic on main — status, requester_user_id, requested_group_id, assignee_user_id, resolver_user_id:                                                                                         
https://github.com/discord/access/blob/5a8e85f1ecfd76908d1637f910b51aa77b8231ed/api/views/resources/access_request.py#L215-L295                                                                       
                                                                                                                                                                                                        
Frontend actively reads filters from URL into state:                                                                                                                                                  
https://github.com/discord/access/blob/803e8e906bc3c31594a686b034fd948f27dea9d3/src/pages/requests/List.tsx#L54-L59                                                                                   
                                                                                                                                                                                                        
Frontend actively sends filters as query params:                                                                                                                                                      
https://github.com/discord/access/blob/803e8e906bc3c31594a686b034fd948f27dea9d3/src/pages/requests/List.tsx#L70-L75                                                                                   
                                                                                                                                                                                                        
---                                               
3. POST /api/requests no longer auto-cancels duplicate PENDING requests (confidence: 85)                                                                                                              
                                                                                                                                                                                                        
Bug site — new POST handler goes straight from group lookup to CreateAccessRequest, no duplicate-cancel block:
https://github.com/discord/access/blob/803e8e906bc3c31594a686b034fd948f27dea9d3/api/routers/access_requests.py#L63-L87                                                                                
                                                                                                                                                                                                        
Old cancel-duplicates logic on main — query existing PENDING and reject each:                                                                                                                         
https://github.com/discord/access/blob/5a8e85f1ecfd76908d1637f910b51aa77b8231ed/api/views/resources/access_request.py#L380-L394                                                                       
                                                                                                                                                                                                        
Operation also lacks the cancel logic — confirms it wasn't moved into CreateAccessRequest.execute():                                                                                                  
https://github.com/discord/access/blob/803e8e906bc3c31594a686b034fd948f27dea9d3/api/operations/create_access_request.py#L63-L82                                                                       
                                                                                                                                                                                                        
---                                               
4. Pagination next URL points to current page on the last page (confidence: 85)                                                                                                                       
                                                                                                                                                                                                        
Bug site — next_page = min(page + 1, pages - 1) resolves to page itself when page == pages - 1: https://github.com/discord/access/blob/803e8e906bc3c31594a686b034fd948f27dea9d3/api/pagination.py#L84-L92                                                                                             
                                                                                                                                                                                                        
No has_next / has_prev in the response envelope (so clients can't otherwise detect end of list):                                                                                                      https://github.com/discord/access/blob/803e8e906bc3c31594a686b034fd948f27dea9d3/api/pagination.py#L88-L94                                                                                             
                                                                                                                                                                                                        
Old behavior on main — used page_obj.has_next / has_prev to compute the URL conditionally:                                                                                                            https://github.com/discord/access/blob/5a8e85f1ecfd76908d1637f910b51aa77b8231ed/api/pagination.py#L29-L42                                                                                             
                                                                                                                                                                                                        
---                                               
5. JWKS TTLCache is not thread-safe (confidence: 75)                                                                                                                                                  
                                                                                                                                                                                                        
Bug site — @cached(cache=TTLCache(...)) with no lock=:
https://github.com/discord/access/blob/803e8e906bc3c31594a686b034fd948f27dea9d3/api/auth/cloudflare.py#L26-L27                                                                                        
                                                                                                                                                                                                        
Race trigger — _refresh_keys does cache_clear() immediately followed by _signing_keys(); not atomic:                                                                                                  
https://github.com/discord/access/blob/803e8e906bc3c31594a686b034fd948f27dea9d3/api/auth/cloudflare.py#L39-L41                                                                                        
                                                                                                                                                                                                        
Where it's invoked under load — verify_cloudflare_token calls _refresh_keys on kid miss (every key rotation):                                                                                         
https://github.com/discord/access/blob/803e8e906bc3c31594a686b034fd948f27dea9d3/api/auth/cloudflare.py#L67-L72

eguerrant (dismissed):

eguerrant (dismissed):

eguerrant (approved):

PR #426, opened 2026-04-27
Create app bug fix and tests

eguerrant:
If an attacker can pre-create a group named App-<target>-Owners, it to be silently adopted as the owner group when the target app is later created, granting any existing group owners ownership of that app.

Added fix and tests. 
- Groups with the app group owner name pattern will only be used as the owner group during app creation if the group has no owners
- Banned the use of the reserved 'app group prefix'-name-'app owner group suffix' name pattern in group creation and group requests
- Banned the reserved app group prefix from being used with vanilla groups (both new groups/group requests and changing group type)

### DIFF
--- a/api/views/resources/app.py
+++ b/api/views/resources/app.py
@@ -340,16 +340,16 @@ class AppList(MethodResource):
         poly_group = with_polymorphic(OktaGroup, [AppGroup, RoleGroup])
         existing_owner_group = (
             db.session.query(poly_group)
-            .options(joinedload(poly_group.active_user_memberships_and_ownerships))
+            .options(joinedload(poly_group.active_user_ownerships))
             .filter(func.lower(OktaGroup.name) == func.lower(owner_group_name))
             .filter(OktaGroup.deleted_at.is_(None))
             .first()
         )
-        if existing_owner_group is not None and len(existing_owner_group.active_user_memberships_and_ownerships) > 0:
+        if existing_owner_group is not None and len(existing_owner_group.active_user_ownerships) > 0:
             abort(
                 409,
-                f"An owner group with members and/or owners already exists for this app name. Select a different app"
-                f" name, change the name of {owner_group_name}, or remove the existing members and/or owners of"
+                f"An owner group with existing owners already exists for this app name. Select a different app"
+                f" name, change the name of {owner_group_name}, or remove the existing owners of"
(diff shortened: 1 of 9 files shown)


api/views/resources/app.py
```diff
@@ -337,6 +337,22 @@ def post(self) -> ResponseReturnValue:
...
 
+        poly_group = with_polymorphic(OktaGroup, [AppGroup, RoleGroup])
+        existing_owner_group = (
+            db.session.query(poly_group)
+            .options(joinedload(poly_group.active_user_ownerships))
+            .filter(func.lower(OktaGroup.name) == func.lower(owner_group_name))
+            .filter(OktaGroup.deleted_at.is_(None))
+            .first()
+        )
+        if existing_owner_group is not None and len(existing_owner_group.active_user_ownerships) > 0:
```
somethingnew2-0, line 348: I'm not sure if this will break the development bootstrapping flow where we import existing groups from our staging Okta instance and then import/create the built-in "App-Access-Owners" group for the built-in "Access" app. I guess since Okta doesn't have group owners, an "App-Access-Owners" group imported from Okta wouldn't have any initial owners, so it shouldn't matter?

api/views/resources/group.py
```diff
@@ -488,6 +491,17 @@ def post(self) -> ResponseReturnValue:
         if existing_group is not None:
             abort(400, "Group already exists with the same name")
 
+        if type(group) is not AppGroup and group.name.startswith(AppGroup.APP_GROUP_NAME_PREFIX):
+            abort(400, "The App- prefix cannot be used for non-app groups. Please choose a different group name.")
```
somethingnew2-0, line 512: Should we also do the same thing for `Role-` groups?

api/views/resources/app.py
eguerrant: Double checked the syncer and bootstrapping flow and I think we're good here since only groups with existing owners (which only exist in Access and not in Okta) are being blocked. Groups with existing members are fine and should be imported without any problems.

somethingnew2-0 (dismissed):

api/views/resources/group.py
eguerrant: Added

somethingnew2-0 (approved):

PR #429, opened 2026-05-05
[roles_add_groups] updated visibility for "Add Groups"

Jeff-Rowell:
## Summary

Previously the **Add Groups** and **Add Owner Groups** buttons on a role page were gated by `isAccessAdmin`, hiding them from role owners who weren't also Access admins.
  - Visibility is now `isAccessAdmin || (role owner who owns ≥1 non-role group)`.
  - For non-admins, the dialog's autocomplete is restricted to groups the user owns, and an info note explains that unowned groups must be added via a role request from the group's page.
  - This brings the role page in parity with the group page, where `AddRoles` already permits group owners (not just admins).

## Testing
  - [x] Locally stubbed [`isAccessAdmin`](https://github.com/discord/access/blob/main/src/authorization.tsx#L24) to return `false` to simulate a non-admin role owner.

image

  - [x] As role owner of `Role-JeffTestRole` who owns multiple groups, both buttons appear, the dialog shows the info note, and the autocomplete only lists owned groups.

Screenshot 2026-05-05 at 10 11 55 AM

Screenshot 2026-05-05 at 10 12 03 AM

### DIFF
--- a/src/pages/roles/AddGroups.tsx
+++ b/src/pages/roles/AddGroups.tsx
@@ -108,6 +108,12 @@ function AddGroupsDialog(props: AddGroupsDialogProps) {
     props.group.active_user_memberships?.map((membership) => membership.active_user!.id).includes(currentUser.id) ??
     false;
 
+  const userOwnedNonRoleGroupIds = !isAccessAdmin(currentUser)
+    ? (currentUser.active_group_ownerships ?? [])
+        .filter((ownership) => ownership.active_group != null && ownership.active_group.type !== 'role_group')
+        .map((ownership) => ownership.active_group!.id!)
+    : null;
+
   const [until, setUntil] = React.useState(accessConfig.DEFAULT_ACCESS_TIME);
   const [groupSearchInput, setGroupSearchInput] = React.useState('');
   const [groups, setGroups] = React.useState<Array<OktaGroup | AppGroup>>([]);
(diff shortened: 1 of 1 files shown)


src/pages/roles/AddGroups.tsx
```diff
@@ -387,7 +400,12 @@ interface AddGroupsProps {
 export default function AddGroups(props: AddGroupsProps) {
   const [open, setOpen] = React.useState(false);
 
-  if (props.group.deleted_at != null || !isAccessAdmin(props.currentUser)) {
+  const ownsAtLeastOneNonRoleGroup = (props.currentUser.active_group_ownerships ?? []).some(
+    (ownership) => ownership.active_group != null && ownership.active_group.type !== 'role_group',
+  );
+  const isRoleOwnerWithGroups = isGroupOwner(props.currentUser, props.group.id ?? '') && ownsAtLeastOneNonRoleGroup;
+
+  if (props.group.deleted_at != null || !(isAccessAdmin(props.currentUser) || isRoleOwnerWithGroups)) {
```
eguerrant:
```suggestion
  if (props.group.deleted_at != null || !isAccessAdmin(props.currentUser) || !isRoleOwnerWithGroups) {
```
nit: can you change this to distribute the `not`? It makes this easier to read so the reader doesn't have to parse parentheses

eguerrant (dismissed):
Just left one tiny style comment. Lgtm otherwise

somethingnew2-0 (approved):

PR #443, opened 2026-05-12
Access mcp server

eguerrant:
Add embedded MCP server (off by default)

Adds an optional MCP server at /mcp, gated behind ENABLE_MCP=False. Lets any MCP-compatible LLM client (Claude Code, Cursor, Zed, OpenAI clients, MCP Inspector, self-hosted models) read Access data and submit access/role/group requests on behalf of the authenticated user.

### What's added
- ENABLE_MCP config flag. Off by default; operators not running LLM tooling pay nothing.
- FastMCP server (mcp[cli]==1.27.0) mounted at /mcp via a Starlette Route (works without a trailing slash).
- Two built-in MCP auth providers, mutually exclusive for /mcp:
  - Cloudflare Access (verifies the CF JWT; activates when CLOUDFLARE_TEAM_DOMAIN is set).
  - OIDC bearer-token (verifies a JWT against your IdP's JWKS; activates when OIDC_SERVER_METADATA_URL is set; requires OIDC_MCP_AUDIENCE so the aud claim is validated, fails closed at startup if missing).
  - Plus a dev provider for local development (gates on ENV=development/test).
- source field on AuditLogSchema and RequestContext. Auto-injected from context so existing operations tag correctly with zero per-operation code changes.
- MCP_FALLBACK_SCOPES config (default read_all,create_requests). Tokens carrying an explicit scope claim override it.
- MCP_ALLOWED_HOSTS config. Empty by default (DNS-rebinding protection off; structural defenses cover the threat). Set to your public host for defense-in-depth, or to localhost:* for dev work with a browser open.
- @requires_scope decorator on every tool; scope errors are returned as a standard {"error": "..."} JSON envelope.
- 30 MCP tests; full suite is 451 green. Ruff and mypy clean.

### Tool surface (21 tools, 1 prompt)
- Reads (read_all scope): list/get for groups, roles, apps, users, tags, access requests, role requests, group requests; plus list_audit_entries and list_group_memberships.
- Writes (create_requests scope): create_access_request, create_role_request, create_group_request. Each runs the same authorization predicate as the matching REST endpoint (e.g. create_role_request gates on can_manage_group(role)).
- Prompt: request_access guides users toward role-based requests, time-bounded grants, and surfacing tag constraints up front.
- Approvals, rejections, and direct membership edits are intentionally NOT exposed.

### Three-layer authorization
1. @requires_scope(...) decorator: the token attenuates which tools can be called.
2. Bare predicates from api/auth/permissions.py: gate what the resolved user can do.
3. Operation constraints: tag constraints, is_managed, reason requirements.

Scopes never grant permissions. MCP tools can do strictly less than what the same user can do via REST.

### Open-source posture
- LLM-agnostic: standard MCP Streamable HTTP, any compliant client connects by URL.
- IdP-agnostic: CF Access and OIDC ship as the two built-in providers. Access was originally Okta-only and the Okta prefix on some model names (OktaGroup, OktaUser, etc.) is legacy; the MCP server's instructions tell the LLM to treat them as a vestigial label, not a deployment requirement.
- Off by default: no overhead for operators not running LLM tooling.

### Operator notes
- ENABLE_MCP=true to activate.
- For Cloudflare deployments: set CLOUDFLARE_TEAM_DOMAIN + CLOUDFLARE_APPLICATION_AUDIENCE.
- For OIDC deployments: set OIDC_SERVER_METADATA_URL + OIDC_MCP_AUDIENCE. The MCP path is a bearer-token resource server, not the REST OIDC session-cookie integration.
- Default fallback scope set is read_all,create_requests. Set MCP_FALLBACK_SCOPES=read_all for read-only sessions, or MCP_FALLBACK_SCOPES=\"\" to fail closed for tokens that don't carry a scope claim.

## TODO
- [x] Review code
- [x] Add README + deployment notes
- [ ] Disallow client-side auto-approval for write tools (elicitations on create_access_request / create_role_request / create_group_request; follow-up)
- [x] Test locally
- [ ] Test in staging
- [ ] Add rate limiting for /mcp (follow-up PR)


api/plugins/mcp_auth.py
somethingnew2-0, line 1: Can we provide an example implementation for this in `examples/plugins` for an example Okta OIDC authentication setup? Or should we provide OIDC as another built-in auth method like we do today in `api/auth/dependencies.py` instead of just a Cloudflare?

api/config.py
```diff
@@ -108,6 +108,30 @@ def TESTING(self) -> bool:
...
+    # runs the same authorization predicate (Layer 2) and operation
+    # constraints (Layer 3) the matching REST endpoint applies, so this
+    # grants the user no capability they don't already have via REST; it
+    # just permits the tool to be CALLED. Operators who want a stricter
+    # posture can set this to `"read_all"` (read-only sessions) or `""`
+    # (fail-closed; only tokens with an explicit scope claim work).
+    # When CF (or your provider) starts emitting scope claims, this
+    # fallback never fires and the token controls scope per session;
+    # the value here becomes inert.
+    MCP_FALLBACK_SCOPES: str = "read_all,create_requests"
```
somethingnew2-0, line 149: Should this default to read_all instead of both scopes?

requirements.txt
```diff
@@ -35,6 +35,12 @@ okta==2.9.8
 # Plugins
 pluggy==1.5.0
 
+# MCP — embedded Model Context Protocol server (gated behind ENABLE_MCP).
+# Pinned to the same minor as Sirius's lockfile so we're tracking a build
```
somethingnew2-0: Remove the reference to Sirius

api/mcp/server.py
```diff
@@ -0,0 +1,256 @@
+"""FastMCP server construction, ASGI auth middleware, and lifespan.
+
+Wiring overview (matches Sirius's reporting-service for shape, adapted to
```
somethingnew2-0: Remove the reference to Sirius

api/mcp/server.py
```diff
@@ -0,0 +1,256 @@
...
+
+
+def get_mcp_route() -> Route:
+    """Return a Starlette Route bound at ``/mcp`` (no trailing slash).
+
+    ``StreamableHTTPASGIApp`` is the raw ASGI handler the session manager
+    exposes; ``Route`` registers it as a normal Starlette route so the
+    request actually traverses the outer middleware stack (RequestId,
+    CacheControl, etc.) on the way in and out. A ``Mount`` would require
+    ``/mcp/`` and 405 on bare-``/mcp`` POSTs (per Sirius's experience),
```
somethingnew2-0: Remove the reference to Sirius

requirements.txt
```diff
@@ -35,6 +35,12 @@ okta==2.9.8
...
+# that's already running in prod. Bumping the major (2.x) will break
+# imports — FastMCP's API has historically reshaped between majors.
+mcp[cli]==1.27.0
```
somethingnew2-0, line 40: Does this include the FastMCP package as a dependency?

api/mcp/auth/cloudflare.py
```diff
@@ -0,0 +1,192 @@
...
+            .filter(OktaUser.deleted_at.is_(None))
+            .first()
+        )
+        if user is None:
+            logger.warning(
+                f"CF Access JWT verified for email={payload['email']!r} but no matching active OktaUser exists"
+            )
+            return None
+        user_id = user.id
+    elif "common_name" in payload:
```
somethingnew2-0: Not sure we need to support this since Cloudflare Service Auth tokens likely won't be able to authenticate to the MCP endpoint via Cloudflare Managed OAuth

api/mcp/prompts.py
```diff
@@ -0,0 +1,146 @@
...
+- **Least privilege**: never recommend a permission tier broader than
+  the task requires.
+
+## Available tools (v1)
+- Read/list: ``list_groups`` / ``get_group``, ``list_roles`` /
+  ``get_role``, ``list_apps`` / ``get_app``, ``list_users`` /
+  ``get_user``, ``list_tags`` / ``get_tag``, ``list_access_requests`` /
+  ``get_access_request``, ``list_role_requests`` /
+  ``get_role_request``, ``list_group_requests`` /
+  ``get_group_request``, ``list_audit_entries``,
```
somethingnew2-0: We should also highlight the list_audit_entries tool as a way for an AI agent to debug what access was lost recently for a user, group, or role

api/mcp/server.py
```diff
@@ -0,0 +1,256 @@
...
+     swaps the active ``RequestContext`` for one tagged
+     ``source="mcp"`` so the audit log can distinguish MCP-driven
+     entries. Returns 401 when every provider returns None.
+  4. ``mcp_lifespan`` is an async context manager that the FastAPI app
+     enters during its lifespan so the FastMCP session manager's task
+     group is alive while requests are served.
+
+The FastAPI app-wide ``Depends(require_authenticated)`` does NOT
+propagate into a Starlette ``Route`` mounted this way, so the MCP path
+runs in its own auth/authz lane. That's correct: AuthN is the
```
somethingnew2-0:
```suggestion
runs in its own auth/authz lane. AuthN is the
```

api/mcp/server.py
```diff
@@ -0,0 +1,256 @@
...
+    from api.mcp.tools import register_tools
+
+    mcp = FastMCP(
+        "Access",
+        instructions=ACCESS_MCP_INSTRUCTIONS,
+        stateless_http=True,
+        # The deployment edge handles DNS rebinding — Cloudflare for
+        # Discord, the operator's proxy elsewhere. Leaving it on would
+        # block legitimate same-origin proxying setups.
+        transport_security=TransportSecuritySettings(enable_dns_rebinding_protection=False),
```
somethingnew2-0: Do we only want to disable DNS rebinding protection in dev (on localhost)? Should we enable it and provide a list of `allowed_hosts` instead for staging and production?

api/mcp/server.py
```diff
@@ -0,0 +1,256 @@
...
+        # normal FastAPI auth gate.
+        if path != MCP_PATH and not path.startswith(MCP_PATH + "/"):
+            await self.app(scope, receive, send)
+            return
+
+        # Run the pluggy hook chain. ``firstresult=True`` means we get
+        # the first non-None identity back, or None if every registered
+        # provider deferred.
+        try:
+            identity: Optional[MCPIdentity] = get_mcp_auth_hook().mcp_resolve_identity(scope=scope)
```
somethingnew2-0: I do wonder if this plugin interface will be sufficient to implement an OIDC flow which requires running an authorization server with callback urls etc.

api/mcp/tools.py
```diff
@@ -0,0 +1,1505 @@
+"""MCP tool registrations.
+
+Each tool mirrors the corresponding REST endpoint:
+
+  - Reuses the same eager-load shape so ``lazy="raise_on_sql"`` doesn't
+    bite at runtime (every relationship the response schema reads must
+    be loaded before serialization).
+  - Reuses the same Pydantic response model so the wire shape matches
```
somethingnew2-0, line 8: Probably fine for now, but we should consider different MCP responses that may be easier for the AI agent to parse instead of requiring potentially unintuitive understanding of the naming of fields in our API responses like `active_group_memberships` and `active_role_group_mapping`

api/mcp/tools.py
```diff
@@ -0,0 +1,1505 @@
...
+            "and description. Set 'managed=false' to include externally-managed "
+            "groups. Requires the 'read_all' scope."
+        ),
+        annotations=_READ_ANNOTATIONS,
+    )
+    def list_groups(
+        q: str = "", managed: Optional[bool] = None, page: int = 0, size: int = MCP_DEFAULT_PAGE_SIZE
+    ) -> str:
+        try:
+            require_scope(MCP_SCOPE_READ_ALL)
```
somethingnew2-0: Can this be a decorator on the method instead of inline?

api/mcp/tools.py
```diff
@@ -0,0 +1,1505 @@
...
+        "size": size,
+        "has_next": page < pages - 1,
+        "has_prev": page > 0,
+        "results": results,
+    }
+
+
+def _serialize_model(model: Any) -> Any:
+    """Dump a Pydantic model to a plain JSON-serializable dict."""
+    return json.loads(model.model_dump_json())
```
somethingnew2-0: This feels weird to do especially as the follow-up call is to `json.dumps` later. Maybe this can be optimized

api/mcp/tools.py
```diff
@@ -0,0 +1,1505 @@
...
+    ) -> str:
+        try:
+            require_scope(MCP_SCOPE_READ_ALL)
+        except MCPScopeError as e:
+            return _error(str(e))
+        page, size, err = _clamp_pagination(page, size)
+        if err:
+            return _error(err)
+        db = _db_shim.session
+        query = (
```
somethingnew2-0, line 385: Since most of this code is copied from the `api/routers` I wonder if we can abstract it where both the router and MCP have a shared method to call maybe defined in `api/views` or something?

api/mcp/server.py
somethingnew2-0: Actually, this is mostly a protection when running the MCP over localhost. So it's probably not necessary for non-dev environments as CORS would apply then.

requirements.txt
eguerrant: Yes it does

api/mcp/server.py
eguerrant: Made it config driven (off by default to avoid 421s in the case that an allowlist isn't provided). I agree with the second comment that this probably isn't crucial for stg/prd

api/config.py
eguerrant: Could go either way on this but given that tool use is still gated on whether or not the user has the permissions to do an action (eg. creating a role request requires them to own a role) I was leaning towards leaving both scopes as the default

api/mcp/auth/cloudflare.py
eguerrant: Removed this code

api/mcp/prompts.py
eguerrant: Done

api/mcp/server.py
eguerrant: Oh I don't think it is. I think the idea was to be able to support different OIDC proxies that Access just needs to verify. Open to expanding this but I figured it was out of scope for a v1 of the mcp

api/mcp/tools.py
eguerrant: Added description of fields in tool descriptions, can follow up later with either serialization aliases or new LLM-specific models

api/plugins/mcp_auth.py
eguerrant: Removed the plugin system and added in OIDC

api/mcp/tools.py
eguerrant: I think it'd be possible to create a top level layer for query construction that would remove some duplicate code and reduce the possibility of drift but since that would touch the routers as well I'm inclined to say that should be left as a follow up. This PR is already massive

api/mcp/auth/oidc.py
```diff
@@ -0,0 +1,160 @@
+"""Built-in OIDC MCP auth provider.
+
+Activates when ``settings.OIDC_SERVER_METADATA_URL`` is set. The MCP path
+uses a **resource-server** OIDC integration, not the browser
+session-cookie flow ``api/auth/oidc.py`` runs for REST: MCP clients
+aren't browsers and the MCP OAuth spec uses Bearer tokens in
+``Authorization: Bearer <jwt>``. The OAuth dance (authorization code,
```
somethingnew2-0, line 7: We'll likely need to support steps 1 through 5 in the [MCP Authorization Guide](https://modelcontextprotocol.io/docs/tutorials/security/authorization#the-authorization-flow-step-by-step) for OIDC. [Cloudflare Managed OAuth for Cloudflare Access already handles those for us](https://blog.cloudflare.com/managed-oauth-for-access/), but for the MCP OIDC flow in Discord Access we'll likely need to implement them ourselves.

eguerrant: I implemented steps 1 + 2 (newly added) and 6 (already there for CF). As far as I understand, this makes Access a compliant MCP resource server, which is what we'd need to have this be compatible with Keycloak, Okta (custom authorization server), Auth0... The implementation validates tokens with JWKS signature checks so the IdP would need to issue a JWT access token but that should be handled by a lot of Idps. 

If we want the Okta org default server to also work I think an intorspection-based variant will need to be added but I'm inclined to leave that as future work

somethingnew2-0 (dismissed):

api/mcp/auth/oidc.py
somethingnew2-0: Thanks for adding this!

somethingnew2-0 (approved):

matanbaruch:
@somethingnew2-0 @eguerrant is this released?

somethingnew2-0:
> @somethingnew2-0 @eguerrant is this released?

@matanbaruch The MCP server is merged and ready to use. We're about to release Access 2.0 with a large list of changes including the MCP server very soon.

PR #446, opened 2026-05-14
Wire Claude Code Desktop Preview for the dev stack

somethingnew2-0:
## Summary

- Adds `.claude/launch.json` so Claude Code Desktop's **Preview** dropdown can auto-start uvicorn and Vite, open the embedded browser, and run auto-verify after edits.
- Both servers use `autoPort: true` so multiple worktree sessions can run side-by-side without colliding on the dev ports.
- Pushes compound startup logic (venv activation, `alembic upgrade head`, port handoff) into the Makefile so it's reusable from the terminal and discoverable via `make help`.
- Carries the local SQLite DB across new worktrees via `.worktreeinclude`.

## What changed

**`.claude/launch.json`** — two minimal configs that shell out to `make run-backend` / `make run-frontend`, with `autoPort: true` and preferred ports `6060` / `3000` (the historical defaults).

**`Makefile`** —
- New `BACKEND_PORT` / `FRONTEND_PORT` variables that honor the `PORT` env var (which Claude injects when autoPort picks a free port) and fall back to `6060` / `3000`.
- `run-backend` now depends on `db-migrate`, writes the bound port to `.claude/.api-port`, and uses `$(BACKEND_PORT)` for uvicorn's `--port`. Same shape for `run`.
- `run-frontend` switched from `npm start` to `npx vite --port $(FRONTEND_PORT)` so the flag wins over `vite.config.ts`.
- Help text interpolates the variables.

**`vite.config.ts`** — keeps `server.port: 3000` and the static `target: 'http://localhost:6060'` proxy fallback, and reads `.claude/.api-port` at Vite startup to pick up the autoPort-assigned backend port (matches the keysmith pattern; restart Vite if the backend port changes).

**`.gitignore`** — adds `.claude/.api-port`.

**`.worktreeinclude`** — adds `instance/access.db` so new worktrees inherit the local dev SQLite DB instead of starting empty. (Already gitignored via `instance/*`, which `.worktreeinclude` requires.)

## Why the file-based port handoff

`launch.json`'s `env` is static (same value across worktrees), so we can't bake the backend port into the frontend config. The backend writes its assigned port to `.claude/.api-port` when it starts, and Vite reads that file once at startup to configure the `/api` proxy target — so the backend must be started before the frontend, and Vite needs to be restarted if the backend port changes. Each worktree has its own `.claude/` directory, so worktrees stay isolated.

The browser only talks to Vite, which proxies `/api/*` with `changeOrigin: true`, so the backend sees server-to-server requests from Vite — `CLIENT_ORIGIN_URL`/CORS need no changes.

## No port changes for terminal users

Dev port defaults stay at `6060` (backend) and `3000` (frontend), so `make run` from a terminal behaves exactly as before. autoPort only changes ports when the defaults are taken (e.g. a second worktree session running in parallel).

## Test plan

- [x] Single worktree: open Desktop → **Preview** dropdown shows `backend` and `frontend`. Start both. Backend binds 6060, frontend binds 3000, `.claude/.api-port` contains `6060`, embedded browser loads Access, a sample `/api/*` request returns expected JSON.
- [ ] Multi-worktree: open a session in a second worktree and start Preview there. Backend autoports to e.g. 6061, frontend to e.g. 3001, that worktree's `.claude/.api-port` contains `6061`. Confirm in DevTools that `/api/*` from the second worktree's frontend hits the second worktree's backend.
- [ ] Backend-port change requires Vite restart: change the backend port and confirm `/api/*` fails until Vite is restarted (re-reads `.claude/.api-port` at startup).
- [ ] Terminal flow: `make run` in a fresh checkout → backend on 6060, frontend on 3000, app works as before.
- [ ] Fallback: delete `.claude/.api-port` and start Vite. Confirm `/api/*` proxies to `localhost:6060` and the app still works as long as a backend is listening there.
- [ ] Worktree DB carryover: create a new worktree from a checkout with a populated `instance/access.db`; confirm the DB is copied into the new worktree and `make run` doesn't require re-running `make db-init`.
- [ ] Auto-verify: ask Claude to make a frontend edit; confirm Desktop screenshots the preview and verifies the change.

### DIFF
--- a/.claude/launch.json
+++ b/.claude/launch.json
@@ -5,14 +5,14 @@
       "name": "backend",
       "runtimeExecutable": "make",
       "runtimeArgs": ["run-backend"],
-      "port": 8000,
+      "port": 6060,
       "autoPort": true
     },
     {
       "name": "frontend",
       "runtimeExecutable": "make",
       "runtimeArgs": ["run-frontend"],
-      "port": 5173,
+      "port": 3000,
       "autoPort": true
     }
   ]
--- a/.claude/launch.json
+++ b/.claude/launch.json
@@ -1,0 +1,19 @@
+{
+  "version": "0.0.1",
+  "configurations": [
+    {
+      "name": "backend",
+      "runtimeExecutable": "make",
+      "runtimeArgs": ["run-backend"],
+      "port": 8000,
+      "autoPort": true
+    },
+    {
+      "name": "frontend",
+      "runtimeExecutable": "make",
+      "runtimeArgs": ["run-frontend"],
+      "port": 5173,
+      "autoPort": true
+    }
+  ]
+}
--- a/.gitignore
+++ b/.gitignore
@@ -8,6 +8,8 @@
 instance/*
 !instance/.gitignore
 .env
+# Port handoff for Claude Code Desktop Preview (see .claude/launch.json)
+.claude/.api-port
 
 ### Python ###
 # Byte-compiled / optimized / DLL files
--- a/.worktreeinclude
+++ b/.worktreeinclude
@@ -10,3 +10,7 @@
 config/config.override.json
 config/config.development.json
 config/config.test.json
+
+# Local SQLite DB (LOCAL_DB_URI in the Makefile). Carrying it across
+# worktrees skips re-running `make db-init` for each new worktree.
+instance/access.db
--- a/Makefile
+++ b/Makefile
@@ -7,8 +7,8 @@ LOCAL_DB_URI ?= sqlite:///instance/access.db
 
 # Ports — overridable via the PORT env var so Claude Code Desktop Preview's
 # autoPort can pick a free port per worktree. See .claude/launch.json.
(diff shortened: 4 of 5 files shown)


matthew-bass (approved):
> vite.config.ts — keeps server.port: 3000 and the static target: 'http://localhost:6060' proxy fallback, and adds a router callback that reads .claude/.api-port per request to follow autoPort-assigned backend ports.

I think this part of the description might be inaccurate, it seems like it reads the port file on startup, not per request, same as the keysmith changes.

Approving, but just flagging in case you wanted that different behavior

somethingnew2-0:
> > vite.config.ts — keeps server.port: 3000 and the static target: 'http://localhost:6060' proxy fallback, and adds a router callback that reads .claude/.api-port per request to follow autoPort-assigned backend ports.
> 
> I think this part of the description might be inaccurate, it seems like it reads the port file on startup, not per request, same as the keysmith changes.
> 
> Approving, but just flagging in case you wanted that different behavior

Oh yeah, I think that was from a previous commit where Vite proxied more dynamically on each request

PR #444, opened 2026-05-12
Make config's APP_CREATOR_ID accept a comma-separated list

tonydelanuez:
Parsed as comma-separated so multiple identifiers can be permitted to create apps. 
Existing single-value configs parse to a one-element list and are unchanged.

PR #447, opened 2026-05-14
Allow APP_GROUP_DELETER_ID users to delete managed app groups

tonydelanuez:
Adds a dedicated `APP_GROUP_DELETER_ID` config, parsed comma-separated like `APP_CREATOR_ID` in #444. Identifiers in the list may DELETE managed AppGroups via `/api/groups/{id}`. The grant is scoped narrowly: plain OktaGroups and unmanaged AppGroups still 403, and the existing group-owner / app-owner / access-admin paths are unchanged.

### DIFF
--- a/api/auth/permissions.py
+++ b/api/auth/permissions.py
@@ -113,6 +113,12 @@ def can_manage_group(db: Session, current_user_id: str, group: OktaGroup) -> boo
     return False
 
 
+def can_delete_group(db: Session, current_user_id: str, group: OktaGroup) -> bool:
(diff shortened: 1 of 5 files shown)


somethingnew2-0 (changes requested):
Probably should make a new config `APP_GROUP_DELETER_ID` or similar. Also, we'd only want it to be able to delete app groups and as long as their "managed" by Access (group.is_managed = True)

somethingnew2-0 (approved):

PR #454, opened 2026-05-27
Rename resolution_reason → reason on ResolveGroupRequestBody

somethingnew2-0:
## Summary
One of four PRs splitting the original combined wire-shape change into focused, independently reviewable pieces (this PR *was* the original combined #454, now scoped down to just item 22). Targets `main`.

Aligns the group-request resolve body with the access-request and role-request resolve endpoints, which both already accept `reason`. `PUT /api/group-requests/{id}` previously accepted `resolution_reason`.

- `api/schemas/requests_schemas.py` — `ResolveGroupRequestBody.resolution_reason` → `reason`.
- `api/routers/group_requests.py` — handler reads `body.reason`.
- The `resolution_reason` DB column and all **response** fields keep their name — only the request-body key changes.
- Frontend: `ResolveGroupRequest` type + `group_requests/Read.tsx` resolve form field.

This is a **breaking** wire-shape change for clients posting to `PUT /api/group-requests/{id}`.

Closes POST_MIGRATION_TODO.md item #22.

## Test plan
- [x] `mypy .` — no new errors
- [x] `pytest tests/test_group_request.py` — 45 passed
- [x] CI green

## ⚠️ Cross-dependency with the sibling split PRs
This is one of four independent PRs splitting the original combined wire-shape change, all branched off `main` (not stacked, so they can merge in any order — but they **overlap on shared files** and will conflict on merge):

- #463 — RFC 9457 error format (item 7)
- #464 — ISO 8601 datetimes (item 5)
- #465 — exclude_none + fastapi-pagination (items 6 & 8)

Files this PR shares with siblings (expect a rebase for whichever merges second):
- `POST_MIGRATION_TODO.md` — every split PR deletes its own item section (conflicts with all three).
- `api/schemas/requests_schemas.py` — also edited by #464 (datetime types) and #465 (docstring).
- `api/routers/group_requests.py` — also edited by #465 (route_class + pagination).
- `src/api/apiSchemas.ts` — also edited by #465 (pagination types).
- `src/pages/group_requests/Read.tsx` — also edited by #464 and #465.

After any sibling merges, rebase this branch and re-resolve those files.

### DIFF
--- a/api/schemas/requests_schemas.py
+++ b/api/schemas/requests_schemas.py
@@ -346,7 +346,7 @@ CreateGroupRequestBody = Annotated[
 class ResolveGroupRequestBody(BaseModel):
     model_config = ConfigDict(extra="ignore")
     approved: StrictBool
-    resolution_reason: Optional[str] = ""
+    reason: Optional[str] = ""
     resolved_group_name: Optional[str] = None
     resolved_group_description: Optional[str] = None
     resolved_group_type: Optional[str] = None
--- a/POST_MIGRATION_TODO.md
+++ b/POST_MIGRATION_TODO.md
@@ -76,24 +76,6 @@ Drop the hand-rolled `{total, pages, next, prev, results}` envelope in
 and consistent shape across endpoints for free. Wire shape changes; coordinate
 with frontend.
 
-### 22. Rename `resolution_reason` → `reason` on `ResolveGroupRequestBody`
-
-`PUT /api/group-requests/{id}` accepts `resolution_reason` while
-`PUT /api/access-requests/{id}` and `PUT /api/role-requests/{id}` both
-accept `reason`. Align on `reason` for the group-request resolve body
-to make the three resolve endpoints consistent.
-
-**Touches:**
-- `api/schemas/requests_schemas.py` — rename the field on
-  `ResolveGroupRequestBody`.
-- `api/routers/group_requests.py` — update the handler to read
-  `body.reason`.
-- Frontend client(s) posting to `PUT /api/group-requests/{id}` — switch
-  the body key.
-- Tests in `tests/test_group_request.py` that exercise the resolve PUT.
-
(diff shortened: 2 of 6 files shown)


api/schemas/requests_schemas.py
```diff
@@ -346,7 +346,7 @@ class _AppGroupRequestBody(_GroupRequestBodyBase):
 class ResolveGroupRequestBody(BaseModel):
     model_config = ConfigDict(extra="ignore")
     approved: StrictBool
-    resolution_reason: Optional[str] = ""
+    reason: Optional[str] = ""
```
Jeff-Rowell, line 349: Because this model uses `extra="ignore"`, any client still posting the old `resolution_reason` key after this ships will get a `200` with an empty reason rather than an error. The resolution succeeds with no recorded reason, which tends to surface weeks later as blank resolution reasons rather than a clear failure.

If any client outside this repo posts to `PUT /api/group-requests/{id}`, consider accepting both keys during a deprecation window via `AliasChoices("reason", "resolution_reason")` and dropping the alias later. If the only client is this repo's frontend (updated in the same PR), the silent-drop risk is negligible and this can stay as-is.

Jeff-Rowell (approved):
One small comment, otherwise LGTM!

api/schemas/requests_schemas.py
somethingnew2-0: Thanks! Agreed the silent drop is worth flagging in general. In this case the only client that posts to `PUT /api/group-requests/{id}` is this repo's React frontend, and this PR updated it (`group_requests/Read.tsx`) to send `reason` in the same change — so there's no window where a known caller keeps sending the old key. Per your last point, leaving it as-is.

If an external integration against this endpoint ever shows up, the clean migration path is `AliasChoices("reason", "resolution_reason")` for a deprecation window, then drop the alias — easy to add later if needed.

🤖 Posted by Claude Code

PR #457, opened 2026-06-01
Hidden app owner groups bug

eguerrant:
POST /api/groups accepts submitting a value for is_owner on app-group create bodies and writes it verbatim. The router only checks the name doesn't end in -Owners. is_app_owner_group_owner considers any AppGroup.is_owner=True for an app as an owner-group. This allows for a group to be set as and behave as an owner group without it being clearly marked as one, potentially allowing for hidden persistent access

Since app groups are only set as owner groups at app creation time, removed is_owner from request bodies and defaulted to setting is_owner to False when creating a new app group

### DIFF
--- a/api/routers/groups.py
+++ b/api/routers/groups.py
@@ -149,7 +149,7 @@ def post_group(
             name=body.name,
             description=description,
             app_id=body.app_id,
-            is_owner=body.is_owner,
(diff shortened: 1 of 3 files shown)


Jeff-Rowell (approved):
LGTM!

PR #459, opened 2026-06-02
Observability metrics fix

Jeff-Rowell:
# Summary

Before the migration to FastAPI, Flask ran directly under gunicorn, and gunicorn's built-in statsd integration emitted `access.gunicorn.requests` and `access.gunicorn.request.duration.*` because gunicorn handled every HTTP request. After the migration to `gunicorn -k uvicorn.workers.UvicornWorker`, uvicorn handles HTTP and gunicorn only spawns workers. The notification metrics were a separate gap of the same shape

As a result, nothing inside gunicorn ever calls into statsd per request so emission stopped entirely.

## What this PR does
- New ASGI middleware emits `access.requests` and `access.request.duration` per request via the existing `metrics_reporter` plugin hook.
- Each notification wrapper emits `access.notifications.{event}.sent` on successful delivery (with `kind:` tag on the three expiring-access variants).
- OIDC query-string redaction moved from the dead `RedactingGunicornLogger` to a `logging.Filter` on `uvicorn.access`.

### DIFF
--- a/api/app.py
+++ b/api/app.py
@@ -21,7 +21,7 @@ from api import exception_handlers, middleware
 from api.config import settings
 from api.database import build_engine
 from api.extensions import db
-from api.log_filters import TokenSanitizingFilter
+from api.log_filters import RedactingUvicornLogger, TokenSanitizingFilter
 from api.services import okta
 
 logger = logging.getLogger(__name__)
@@ -35,6 +35,7 @@ def _configure_logging() -> None:
     token_filter = TokenSanitizingFilter()
     logging.getLogger("authlib").addFilter(token_filter)
     logging.getLogger("uvicorn.access").addFilter(token_filter)
+    logging.getLogger("uvicorn.access").addFilter(RedactingUvicornLogger())
     logging.root.addFilter(token_filter)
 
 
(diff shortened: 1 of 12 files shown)


api/plugins/metrics_reporter.py
eguerrant: Update examples to add tags parameter

api/app.py
eguerrant: Comment out of date

api/plugins/metrics_reporter.py
```diff
@@ -18,18 +18,24 @@ class MetricsReporterPluginSpec:
     def record_counter(
         self,
         metric_name: str,
-        value: float = 1.0,
-        tags: Optional[Dict[str, str]] = None,
+        value: float,
+        tags: Optional[Dict[str, str]],
```
eguerrant, line 29: Pretty sure pluggy forwards an argument to an impl only if neither the spec nor the impl gives that argument a default so any parameter with a default silently ignores the caller's value

This could be a problem if an org using Access updates the repo code but not their plugin implementations, might be worth seeing if you can make this behavior noisy downstream since I think as is it'll be silent

api/plugins/metrics_reporter.py
eguerrant: Might be worth removing tags = None default here and for record_summary to be consistent?

eguerrant (approved):

PR #464, opened 2026-06-09
Normalize datetime wire format to ISO 8601

somethingnew2-0:
## Summary
One of four PRs splitting the original combined wire-shape change into focused, independently reviewable pieces. Targets `main`.

Drops the Marshmallow-era RFC 822 datetime serializer; the wire now carries Pydantic's default ISO 8601 (e.g. `2026-04-26T13:45:00`).

- `api/schemas/rfc822.py` — removes the RFC 822 `PlainSerializer`, keeps the permissive `parse_datetime_value` input parser, and exports `FlexibleDatetime` / `FlexibleDatetimeOpt`.
- `api/schemas/{requests_schemas,core_schemas,audit_rows}.py` — swap all datetime field types.
- Input stays permissive: `parse_datetime_value` still accepts RFC 822 / RFC 2822 **and** ISO 8601, so clients still emitting the old format on input keep working until they migrate. Only the **output** format changes.
- Frontend: 11 `.tsx` files drop the `RFC822_FORMAT` constant and send `dayjs(...).toISOString()`.

This is a **breaking** wire-shape change for clients that parse response datetimes.

Closes POST_MIGRATION_TODO.md item #5.

## Test plan
- [x] `mypy .` — no new errors (22 pre-existing `api/mcp/tools.py` errors are on `main`)
- [x] `pytest` for access/role/group requests + audit + user — 115 passed
- [x] CI green

## ⚠️ Cross-dependency with the sibling split PRs
This is one of four independent PRs splitting the original combined wire-shape change, all branched off `main` (not stacked, so they can merge in any order — but they **overlap on shared files** and will conflict on merge):

- #454 — rename `resolution_reason` → `reason` (item 22)
- #463 — RFC 9457 error format (item 7)
- #465 — exclude_none + fastapi-pagination (items 6 & 8)

Files this PR shares with siblings (expect a rebase for whichever merges second):
- `POST_MIGRATION_TODO.md` — every split PR deletes its own item section (conflicts with all three).
- `api/schemas/requests_schemas.py` — also edited by #454 (field rename) and #465 (docstring); this PR does the `RFC822*` → `Flexible*` type swaps.
- `src/pages/group_requests/Read.tsx` — also edited by #454 and #465.

After any sibling merges, rebase this branch and re-resolve those files.


api/schemas/datetimes.py
somethingnew2-0, line 1: Can we rename this file to datetime or something more appropriate?

api/schemas/requests_schemas.py
```diff
@@ -254,7 +254,7 @@ class CreateRoleRequestBody(BaseModel):
     group_id: str
     group_owner: bool = False
     reason: Optional[str] = ""
-    ending_at: Optional[RFC822DatetimeOpt] = None
+    ending_at: Optional[FlexibleDatetimeOpt] = None
```
somethingnew2-0: If FlexibleDatetimeOpt is already an Optional, does it need to be wrapped in one as well?

api/schemas/datetimes.py
somethingnew2-0: Done — renamed to `api/schemas/datetimes.py` (it's the ISO 8601 serializer + permissive parser now, no longer RFC 822). Updated the three importers in `requests_schemas.py`, `core_schemas.py`, and `audit_rows.py`. (21a0671)

🤖 Posted by Claude Code

api/schemas/requests_schemas.py
somethingnew2-0: Good catch — it's redundant. `FlexibleDatetimeOpt` is already `Annotated[Optional[datetime], BeforeValidator(...)]`, so the outer `Optional[...]` added nothing. Dropped it on all six fields that had it (across `requests_schemas.py` and `core_schemas.py`); they're now just `FlexibleDatetimeOpt = None`. (21a0671)

🤖 Posted by Claude Code

somethingnew2-0: Follow-up: went one step further and removed `FlexibleDatetimeOpt` entirely — nullable fields now use `Optional[FlexibleDatetime]` and required ones use `FlexibleDatetime`, so there's a single Annotated type. (3166168)

One behavioral note worth flagging: `Optional[FlexibleDatetime]` rejects an empty-string (`""`) datetime input with a 400, whereas `FlexibleDatetimeOpt` coerced `""` → `None`. The difference is that the `BeforeValidator` inside `FlexibleDatetimeOpt` wrapped `Optional[datetime]` (so `""`→`None` validated fine), while in `Optional[FlexibleDatetime]` the validator's `None` result hits the inner non-optional `datetime` and is rejected. I confirmed nothing in the repo (tests or frontend) sends `""` for a datetime — the frontend omits the key — so this only tightens validation for malformed input; callers send `null` or omit the field for "no datetime".

🤖 Posted by Claude Code

api/schemas/datetimes.py
```diff
@@ -66,21 +69,4 @@ def _to_naive_utc(value: datetime) -> datetime:
...
-    datetime,
-    BeforeValidator(parse_datetime_value),
-    PlainSerializer(_rfc822, when_used="json"),
-]
-RFC822DatetimeOpt = Annotated[
-    Optional[datetime],
-    BeforeValidator(parse_datetime_value),
-    PlainSerializer(_rfc822, when_used="json"),
-]
+FlexibleDatetime = Annotated[datetime, BeforeValidator(parse_datetime_value)]
```
Jeff-Rowell: **Blocking (correctness): naive-UTC output shifts all displayed times by the viewer's timezone offset.**

`parse_datetime_value` normalizes every value to *naive* UTC (`_to_naive_utc` strips tzinfo), so Pydantic's default serialization emits no timezone marker: `2026-04-26T13:45:00` (no `Z`, no offset) instead of the old `... -0000`.

The frontend has no UTC handling (dayjs `utc` plugin is never loaded); response timestamps go straight into `dayjs(...)` / `new Date(...)`:
- `src/components/Started.tsx:35`, `src/components/Ending.tsx:31` (`.fromNow()`)
- `src/pages/requests/Read.tsx:176`, `src/pages/role_requests/Read.tsx:193`
- `src/pages/roles/Expiring.tsx:327-331`, `src/pages/roles/Audit.tsx:252` (expiry `.isBefore()/.isAfter()`)
- `src/components/AccessHistory.tsx:33` (`new Date(...) > new Date()`)

An ISO string with `-0000` parses as UTC; one with no offset parses as browser-local time. So every displayed timestamp and every expiry comparison shifts by the viewer's UTC offset (e.g. a US/Pacific user sees expirations ~7-8h off). For an access-control tool, wrong expiry display/highlighting is a real regression. The outbound `.toISOString()` edits are correct (always `Z`); the inbound render path is what breaks.

Fix (pick one):
1. Serialize as UTC with a `Z` (add a `PlainSerializer` that appends `Z`, or store/serialize tz-aware UTC). No frontend change needed; smallest, safest diff.
2. Update all the read sites above to parse as UTC (dayjs `utc` plugin / `dayjs.utc(...)`).

Note the docstring above claims output is `"2026-04-26T13:45:00Z"` (with `Z`), which contradicts the actual naive serialization — option 1 would make it true.

Jeff-Rowell (changes requested):

api/schemas/datetimes.py
somethingnew2-0: Good catch — fixed in f20a88e. You're right that `_to_naive_utc` strips tzinfo and Pydantic's default then emits an offset-less `2026-04-26T13:45:00`, which `dayjs(...)`/`new Date(...)` read as browser-local.

Went with option 1: added a `PlainSerializer` to `FlexibleDatetime` that emits an explicit `Z`:

```
def _to_iso_utc(value: Optional[datetime]) -> Optional[str]:
    if value is None:
        return None
    if value.tzinfo is None:
        value = value.replace(tzinfo=timezone.utc)
    else:
        value = value.astimezone(timezone.utc)
    return value.isoformat().replace("+00:00", "Z")
```

Now `{"created_at": "2026-04-26T13:45:00Z", "deleted_at": null}` — `Z` on every datetime, `null` for absent optionals, and microseconds preserved (`...123456Z`). No frontend change, and the docstring's `Z` example is now accurate. Storage stays naive-UTC (the `DateTime()` columns are tz-naive), so only the wire serialization changed.

🤖 Posted by Claude Code

Jeff-Rowell (approved):

PR #465, opened 2026-06-09
Replace paginate() with fastapi-pagination

somethingnew2-0:
## Summary
One of the PRs splitting the original combined wire-shape change into focused, independently reviewable pieces. Targets `main`.

> **Scope note:** this PR originally also carried item 6 (`exclude_none`), but that work was dropped **and item 6 has been removed from `POST_MIGRATION_TODO.md` as a decided non-goal** (not deferred). Including `null` fields is the FastAPI/Pydantic default and the more standard choice for a schema-first API: responses stay aligned with the declared `response_model` / OpenAPI schema that the `apiSchemas.ts` codegen consumes, and clients keep a stable key set instead of having to distinguish "absent" from "null". This PR is now **item 8 only**.

### Replace `paginate()` with `fastapi-pagination`
- Drops the hand-rolled `paginate()` + `PaginationResponse[T]` (10 named subclasses) for `fastapi-pagination`'s `Page[T]` + a custom `PageParams` (1-indexed, `size` capped at `MAX_SIZE=1000`).
- Wire envelope changes from `{total, pages, has_next, has_prev, next, prev, results}` → `{items, total, page, size, pages}`.
- Audit endpoints use `transformer=` to keep the ORM-row → wire-row factory.
- All other list endpoints use `transformer=validated(<Model>)` to **preserve the eager-load safety net** (see below).
- `add_pagination(app)` in `api/app.py`; `fastapi-pagination` added to `requirements.txt`.

#### Eager-load safety net (`validated()` transformer)
The hand-rolled `paginate()` validated each row in-handler via `TypeAdapter(model).validate_python(row, from_attributes=True)`, which deliberately surfaced un-eager-loaded `lazy="raise_on_sql"` relationships as `InvalidRequestError` at test time. `fastapi-pagination`'s default path hands raw ORM rows to FastAPI's `response_model` serialization, where a missing relationship would instead lazily load (the session is usually still open during serialization) — silently turning a test-time failure into a production one. To keep that net, `api/pagination.py` adds `validated(model)`, a `transformer=` factory that re-runs the same in-handler validation; the 8 plain list routers pass `transformer=validated(<Model>)`. The full suite stays green, confirming the per-route eager-load options are complete.

#### Page indexing (frontend)
The wire `page` is now 1-indexed (`page: int = Query(1, ge=1)`). The fix differs by surface:
- **Primary server-paged tables** (every List/Audit/Expiring page) drive a **zero-based** MUI `TablePagination` `page` state and send it to the API. These now send `page + 1`; the local state stays zero-based and the displayed page control reads it directly (the server's returned `page` is never read back, so no `page - 1` map-back is needed). Sending `page: 0` would otherwise 422.
- **Secondary fixed-first-page** autocomplete/lookup queries that previously hard-coded `page: 0` now send `page: 1`.
- **Client-side grids** (`RoleMembers.tsx`) use MUI `DataGrid`'s own zero-based `paginationModel.page`, which is **not** a server page number — left at `page: 0`.
- `results`→`items` across the consuming `.tsx` files; pagination types in `apiSchemas.ts`; page/size params in `apiComponents.ts`. These TS types are hand-maintained — `api/swagger.json` (removed on purpose in #425) is **not** reintroduced here; the checked-in OpenAPI snapshot will come back only when the client codegen is set up (POST_MIGRATION_TODO #13).

This is a **breaking** wire-shape change.

#### Behavior change: invalid `size` rejected with 422 (no clamping)
The old helper clamped `size > 1000` and treated the legacy `per_page=-1` ("all rows") sentinel as "give me everything, capped at the max". With `size: int = Query(..., ge=1, le=MAX_SIZE)` the API now returns **HTTP 422** for both rather than silently clamping — the DoS cap is preserved and out-of-range input is rejected explicitly.

The only first-party caller of the `-1` sentinel was the frontend's **"All" rows-per-page** option (MUI emits `rowsPerPage = -1`). That option has been **removed**: the shared `perPage` is now a concrete `[5, 10, 20, 50, 200, 1000]` (the former "All" becomes a literal `1000` — the API cap — and a `200` is added). With the dropdown no longer able to emit an out-of-range value, the client sends `size` straight through; there is deliberately **no** client-side clamp duplicating the server's `ge=1, le=MAX_SIZE` bound. The three Audit tables, which hardcoded `rowsPerPageOptions={[5, 10, 20]}`, now also use the shared `perPage`. No monorepo client sends `-1` or `size > 1000` (verified separately); any external tooling that relied on the old clamp would need to send a concrete in-range `size`.

#### Empty-row padding (frontend)
The List/Audit/Expiring tables padded each page to `rowsPerPage` height with blank filler rows, to avoid a layout jump when paging to a shorter last page. With the new large page sizes that padding became unwieldy — e.g. 34 rows at `size=200` rendered a ~6000px spacer that pushed the footer far down the page. The computation is now centralized in `emptyTableRows(page, rowsPerPage, rowCount)` in `helpers.tsx` (replacing the duplicated expression across the 13 tables) and gated on `page > 0 && rowsPerPage <= 50`: padding applies only past the first page (a first/only page has no earlier height to match) and only for the smaller sizes, so the `200`/`1000` sizes never pad.

Closes POST_MIGRATION_TODO.md item #8, and removes item #6 as a decided non-goal (see scope note above).

Note: the MCP tool surface (`api/mcp/tools.py`) keeps its own pagination envelope; unifying it is out of scope here.

## Test plan
- [x] `mypy .` — no new errors (22 pre-existing `api/mcp/tools.py` errors are on `main`)
- [x] `pytest tests/` — 425 passed (the `ModuleNotFoundError: No module named 'mcp'` errors are a pre-existing local-venv/test-isolation issue, unrelated to this change)
- [x] `tsc --noEmit` / `prettier --check` — clean
- [x] CI green

## ⚠️ Cross-dependency note
Rebased onto current `main` after the sibling split PRs (#454 reason rename, #463 RFC 9457, #464 ISO 8601) merged. No remaining cross-PR conflicts.


api/pagination.py
```diff
@@ -1,119 +1,45 @@
...
 
-    # Cap `per_page=-1` (sentinel "all") and any positive value above
-    # MAX_PER_PAGE at MAX_PER_PAGE — a small DoS shield against requests
-    # like `?per_page=-1` that would otherwise stream multi-GB of ORM
-    # objects into memory per request.
-    if per_page == -1 or per_page > MAX_PER_PAGE:
-        per_page = MAX_PER_PAGE
-    elif per_page <= 0:
-        per_page = DEFAULT_PER_PAGE
+    page: int = Query(1, ge=1, description="Page number")
```
Jeff-Rowell, line 30: This makes the API 1-indexed and rejects `page < 1`. The frontend's main table queries were not migrated to match: they still feed MUI's zero-based `page` state straight through (see `src/pages/users/List.tsx:43` and the equivalent line in every other List/Audit/Expiring page). `URLSearchParams` serializes `page: 0` as `?page=0` (zeros are not dropped), so `ge=1` returns **HTTP 422** and the table errors out. Even past page 1 it's off-by-one (MUI page 1 -> API page 1 = the first page again).

The `page: 0` -> `page: 1` bumps in this PR only touched the secondary autocomplete queries, not the primary list queries. This needs a consistent fix at the call boundary (e.g. send `page + 1`, map results back with `page - 1`) across all List/Audit/Expiring pages.

api/pagination.py
```diff
@@ -1,119 +1,45 @@
...
+    size: int = Query(DEFAULT_SIZE, ge=1, le=MAX_SIZE, description="Items per page")
```
Jeff-Rowell, line 31: Minor behavior change worth calling out: `size > 1000` and the legacy `per_page=-1` ("all rows") sentinel now return **422** instead of being clamped to `MAX_SIZE` as the old helper did. The DoS cap itself is preserved (good), but any external API consumer or tooling relying on `-1`/oversized values to mean "give me everything, capped" will now get a validation error rather than capped results.

src/pages/users/List.tsx
```diff
@@ -36,18 +36,18 @@ export default function ListUsers() {
       setSearchInput(searchParams.get('q') ?? '');
     }
     setPage(parseInt(searchParams.get('page') ?? '0', 10));
-    setRowsPerPage(parseInt(searchParams.get('per_page') ?? '20', 10));
+    setRowsPerPage(parseInt(searchParams.get('size') ?? '20', 10));
   }, [searchParams]);
 
   const {data, error, isLoading} = useGetUsers({
-    queryParams: Object.assign({page: page, per_page: rowsPerPage}, searchQuery == null ? null : {q: searchQuery}),
+    queryParams: Object.assign({page: page, size: rowsPerPage}, searchQuery == null ? null : {q: searchQuery}),
```
Jeff-Rowell: Looks like `page` here is MUI `TablePagination`'s zero-based index (`useState(0)`, defaulted from `searchParams.get('page') ?? '0'`) and is sent verbatim. With the new backend `page: int = Query(1, ge=1)` this produces `?page=0` -> **HTTP 422** on initial load, breaking the table. This same pattern is unconverted in every List page (groups/roles/apps/tags/requests/role_requests/group_requests), both Audit pages, and both Expiring pages. Convert at the call boundary (`page: page + 1`) and map paging back when wiring `TablePagination`, applied consistently.

src/pages/role_requests/RoleMembers.tsx
```diff
@@ -54,7 +54,7 @@ interface RoleMembersDialogProps {
 function RoleMembersDialog(props: RoleMembersDialogProps) {
   const [paginationModel, setPaginationModel] = React.useState({
     pageSize: 10,
-    page: 0,
+    page: 1,
```
Jeff-Rowell: This grid is **client-side** (rows come from `props.rows`, no API call keyed on page), and MUI DataGrid's `paginationModel.page` is also zero-based. The blanket `page: 0` -> `page: 1` find/replace makes this dialog open on the *second* page. Revert this one to `page: 0`.

api/routers/users.py
```diff
@@ -96,7 +95,7 @@ def list_users(
                 )
             )
 
-    return paginate(request, query, UserPagination, extract=lambda: (q_args.page, q_args.per_page))
+    return paginate(db, query)
```
Jeff-Rowell: The old `paginate()` validated each row via `TypeAdapter(...).validate_python(row, from_attributes=True)`, which deliberately surfaced un-eager-loaded relationships as `InvalidRequestError` (its docstring called this "eager-load coverage stays enforced"). The new path hands raw ORM rows to FastAPI's `response_model` serialization, where a missing relationship will instead lazily load (the session is typically still open during serialization) - silently turning a test-time failure into a production failure. Worth confirming the eager-load options on each list query are still complete, since that safety net is gone.

Jeff-Rowell (changes requested):

api/pagination.py
somethingnew2-0: Fixed in ddbc34f. All primary List/Audit/Expiring pages now send `page + 1` at the call boundary. The local MUI `TablePagination` `page` state stays zero-based and the displayed control reads it directly (`page={page}`); the server's returned `page` is never read back into the UI, so a `page - 1` map-back isn't needed — only the request side needs the `+1`. That resolves both the `?page=0` → 422 on first load and the off-by-one past page 1.

🤖 Posted by Claude Code

api/pagination.py
somethingnew2-0: Good call — I've kept the strict `ge=1, le=MAX_SIZE` (so `size > 1000` and the legacy `per_page=-1` sentinel now 422 rather than clamp) and documented it explicitly under "Behavior change: invalid `size` now 422s instead of clamping" in the PR description. No first-party caller relies on `-1`/oversized-means-capped — the frontend always sends a concrete in-range `size` — so rejecting an out-of-range value is preferable to silently clamping. Any external tooling that depended on the old clamp would need to send a concrete `size`.

🤖 Posted by Claude Code

src/pages/users/List.tsx
somethingnew2-0: Fixed in ddbc34f — converted at the call boundary (`page: page + 1`) consistently across every affected surface: all List pages (users/groups/roles/apps/tags/requests/role_requests/group_requests), both Audit pages, and both Expiring pages. The zero-based local state is unchanged.

🤖 Posted by Claude Code

src/pages/role_requests/RoleMembers.tsx
somethingnew2-0: Reverted to `page: 0` in ddbc34f. Agreed — this is a client-side MUI `DataGrid` `paginationModel.page` (zero-based, rows from `props.rows`, no page-keyed API call), not a server page number, so it must not be bumped.

🤖 Posted by Claude Code

api/routers/users.py
somethingnew2-0: Restored the safety net in ddbc34f rather than relying on the eager-load options alone. `api/pagination.py` now exposes `validated(model)`, a `fastapi-pagination` `transformer=` factory that re-runs the same in-handler `TypeAdapter(model).validate_python(row, from_attributes=True)` the old `paginate()` did — so an un-eager-loaded `lazy="raise_on_sql"` relationship surfaces as `InvalidRequestError` at the pagination boundary (test time) instead of lazily loading during response serialization. Applied to all 8 plain list routers (`transformer=validated(<Model>)`); the audit endpoints already had their own `transformer=`. The full suite stays green, which also confirms the per-route eager-load options are currently complete.

🤖 Posted by Claude Code

Jeff-Rowell (dismissed):

api/pagination.py
somethingnew2-0: Correction to my reply above — I was wrong that "the frontend always sends a concrete in-range `size`." There **is** a first-party caller of the `-1` sentinel: the MUI **"All" rows-per-page** option (`perPage` in `helpers.tsx` emits `{label: 'All', value: -1}`), which the List/Audit/Expiring tables fed straight through as `size=-1`. Under the new `ge=1` that 422s and the table renders "0–0 of 0" — a real regression this PR introduced.

Fixed in 8095ae2 by moving the clamp to the client call boundary rather than special-casing `-1` on the backend: `pageSizeParam()` maps `-1` (and any value above the cap) to `MAX_PAGE_SIZE = 1000`, mirroring the backend `MAX_SIZE` and reproducing the old `paginate()` behavior where `per_page=-1` clamped to `MAX_PER_PAGE`. "All" now sends `size=1000`. Verified live: the request flipped from `GET /api/users?size=-1 → 400` to `GET /api/users?size=1000 → 200` and the table repopulated. The API stays strict (`ge=1, le=MAX_SIZE`). No external/monorepo client sends `-1` or `size > 1000` (verified separately).

🤖 Posted by Claude Code

somethingnew2-0: Final approach (992e0da), simpler than the client-clamp in my previous reply: the `-1` "All" option is **gone** rather than clamped. The shared `perPage` is now a concrete `[5, 10, 20, 50, 200, 1000]` — the former "All" becomes a literal `1000` (the API cap) and a `200` step is added. Since the dropdown can no longer emit an out-of-range value, the client sends `size` straight through and the `pageSizeParam()` clamp was removed — no point duplicating the server's `ge=1, le=MAX_SIZE` bound on the client. The three Audit tables that hardcoded `[5, 10, 20]` now use the shared `perPage` too. Verified live: Users and Audit dropdowns render `5/10/20/50/200/1000` and lists load. The API stays the single source of truth for `size` validation.

🤖 Posted by Claude Code

Jeff-Rowell (approved):

PR #468, opened 2026-06-10
Approve/reject operations silently no-op on stale state; no row-level       locking

eguerrant:
Original issue
- Approve/reject operations silently no-op'd on stale state (already-resolved request, deleted requester, deleted or unmanaged target group), returning the request unchanged with no error and no status flip
- No row-level locking on the request load (no SELECT FOR UPDATE)
- Two concurrent approvers could both pass the pending check and double-grant creating duplicate OktaUserGroupMember rows
- Stale approvals looked like successes to the caller masking the problem

Changes
- Converted silent guards to typed HTTPExceptions across approve and reject operations:
  - already-resolved, now 409
  - deleted requester (access, role), now 410
  - deleted requested group (access, role), now 410
  - unmanaged target group (access, role), now 400
- Added row-level locking on each operation's request load, with_for_update(of=...) on approve (keeps FOR UPDATE off the joinedload's nullable outer-join side, Postgres-safe), plain with_for_update() on reject, no-op on SQLite, so concurrent resolvers serialize on the row and the loser hits the resolved guard
- Updated the not-pending pre-check in all 3 resolve routers from 400 to 409, so the HTTP contract matches the operations
- Left unchanged on purpose (authorization and validation, mirror router checks): self-approval, reason, app-owner/admin authz, name-prefix/collision
- Group-request deleted-requester left silent, its active_requester is an inner join filtering deleted_at, so a deleted requester fails to load and hits the existing not-found guard before reaching that check

### DIFF
--- a/api/operations/approve_access_request.py
+++ b/api/operations/approve_access_request.py
@@ -4,6 +4,7 @@ from typing import Optional
 import logging
 
 from api.context import get_request_context
+from fastapi import HTTPException
 from sqlalchemy.orm import joinedload, selectin_polymorphic
 
 from api.extensions import db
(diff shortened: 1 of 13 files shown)


api/operations/approve_access_request.py
```diff
@@ -51,9 +58,11 @@ def __init__(
         self.notification_hook = get_notification_hook()
 
     def execute(self) -> AccessRequest:
-        # Don't allow approving a request that is already resolved
+        # Don't allow approving a request that is already resolved. Raise
+        # rather than silently no-op so a stale/concurrent approval surfaces
+        # as a conflict instead of looking like a success.
         if self.access_request.status != AccessRequestStatus.PENDING or self.access_request.resolved_at is not None:
-            return self.access_request
+            raise HTTPException(409, "Access request is no longer pending")
```
somethingnew2-0: Feels weird to raise a fastapi HTTPException from an operation which are largely decoupled from the FastAPI stack. Maybe this should be a custom AccessException that can be handled further up the stack and rethrown as an HTTPException?

somethingnew2-0 (approved):

PR #474, opened 2026-06-12
Enforce App- name prefix for app groups server-side

somethingnew2-0:
## What

Server-side enforcement of the `App-{app name}-` naming rule for app groups, replacing a model-level validator that never ran.

- **`CreateGroup`** now validates the prefix alongside the existing app-existence check — covers `POST /api/groups`, group request approval, and app creation.
- **`ModifyGroupDetails`** validates renames of app groups. Unchanged names are tolerated, so description-only edits of legacy non-conforming rows keep working. A `validate_app_group_prefix` flag lets `put_group` skip the check when the rename is part of a type conversion — converting away from an app group legitimately drops the prefix, and `ModifyGroupType` already requires it gone before converting (blind enforcement would deadlock conversions).
- **`put_group`** validates conversions **to** `app_group` against the target app — whether or not the request also renames — before anything commits. (The conversion rename runs while the group is still its old type, so the operation-layer check can't see it.)
- The dead `@validates("name")` hook on `AppGroup` is removed, with a NOTE comment explaining why a model-level validator can't work there.

## Why

`AppGroup`'s `@validates("name")` was provably dead code: with joined-table inheritance, `name` is mapped on the `OktaGroup` base mapper, so a subclass validator never gets wired to the attribute's `set` event — the AppGroup mapper records it in `mapper.validators` but never installs a listener (verified empirically: construction and reassignment both skip it).

That left the prefix rule unenforced on every path, while other code depends on it: app renames rewrite group names by prefix substitution (`api/routers/apps.py`), and group request approval reasons about reserved prefixes. `POST /api/groups` and `PUT /api/groups/{id}` only checked the inverse rules (non-app groups must not use `App-`).

## Also fixed along the way

- `POST /api/groups` with a bogus `app_id` returned a **500** (the operation's `ValueError` was never caught); now a 400 with `"App for AppGroup does not exist"`. Group request approval had the same unwrapped path.
- Converting a group to `app_group` without an `app_id` committed the rename and then died on an IntegrityError (500, with the rename left applied). Now a clean 400 before any mutation.

## Reviewer notes

- **Behavioral change for direct API clients**: non-conforming app group names were previously silently accepted; they now get a 400 with a copy-pasteable example prefix (same message the old validator intended: `App Group name "{name}" should be prefixed with App name. For example: "{prefix}"`). The frontend is unaffected — all three surfaces (group create/update, request creation, approval) compose the prefix client-side and users only type the suffix. Deployments can check for pre-existing non-conforming rows with:
  ```sql
  SELECT og.name FROM app_group ag
  JOIN okta_group og ON og.id = ag.id
  JOIN app ON app.id = ag.app_id
  WHERE og.name NOT LIKE 'App-' || app.name || '-%' AND og.deleted_at IS NULL;
  ```
  Such rows remain fully usable; only an actual rename forces conformance.
- `ModifyGroupType` is deliberately left unchanged: `CreateApp` adopts pre-existing prefix-matched groups case-insensitively (`ilike`), so a strict check inside `ModifyGroupType` would break app creation for case-mismatched adoptees. Conversions via the API are enforced in the router instead.
- Tests cover: POST with non-conforming name / wrong app's prefix / bogus app_id, PUT rename to non-conforming names, legacy-row leniency, conversion to `app_group` (with and without rename, plus state-untouched assertions after each 400), and approval with a non-conforming resolved name (request stays pending, then approves cleanly with a conforming one).

`pytest` (520 passed), `ruff check`, `ruff format`, and `mypy` are all clean.

### DIFF
--- a/api/operations/modify_group_details.py
+++ b/api/operations/modify_group_details.py
@@ -5,7 +5,7 @@ from sqlalchemy import func
 from sqlalchemy.orm import with_polymorphic
 
 from api.extensions import db
-from api.models import AppGroup, OktaGroup, OktaUser, RoleGroup
+from api.models import App, AppGroup, OktaGroup, OktaUser, RoleGroup
 from api.plugins.app_group_lifecycle import get_app_group_lifecycle_hook, get_app_group_lifecycle_plugin_to_invoke
 from api.services import okta
 from api.schemas import AuditLogSchema, EventType
@@ -21,11 +21,17 @@ class ModifyGroupDetails:
         name: str | None = None,
         description: str | None = None,
         current_user_id: str | None = None,
+        validate_app_group_prefix: bool = True,
     ):
         self.group = group
         self.name = name
         self.description = description
         self.current_user_id = current_user_id
+        # Renaming an app group must keep the "App-{app name}-" prefix, except
+        # when the rename is part of a group type conversion: converting away
+        # from an app group legitimately drops the prefix (ModifyGroupType
+        # requires it gone before it will convert).
+        self.validate_app_group_prefix = validate_app_group_prefix
 
     def execute(self) -> OktaGroup:
         old_name = self.group.name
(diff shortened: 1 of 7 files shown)


api/operations/modify_group_details.py
```diff
@@ -42,6 +48,28 @@ def execute(self) -> OktaGroup:
             if existing_group is not None:
                 raise ValueError("Group already exists with the same name")
 
+        # Enforce the "App-{app name}-" prefix on app group renames. Unchanged
+        # names are tolerated so description-only edits of legacy non-conforming
+        # groups keep working.
```
eguerrant, line 53: I'd challenge this maybe. Inclined to say that legacy naming should be allowed as long as the group remains _unedited_ but if a user goes to edit the description they should also be forced to update the name. We have other instances of this, eg. if the required descriptions flag is turned on groups without descriptions are allowed to remain as-is until someone goes to edit any part of the group details then a description is required

eguerrant (approved):
Just the one comment but I'd consider it not blocking

api/operations/modify_group_details.py
somethingnew2-0: Claude wrote this behavior, but I think I like it over a more strict approach. If I'm going to update a group description (or because the group description is now required), I'd prefer to not be blocked on also updating the group name which could fundamentally alter its usage. I'm going to merge in the current conservative behavior, but we can always add in the strict behavior latter especially since it only affect legacy non-conforming group names.

PR #477, opened 2026-06-12
[async 3/7] refactor: eliminate lazy loading and commit-expiration from the ORM layer

somethingnew2-0:
Part of the async-SQLAlchemy migration stack; now rebased directly on `main` (#475, #476 merged).

An `AsyncSession` raises `MissingGreenlet` on any implicit IO from attribute access, so both remaining sources of implicit IO go away before the flip:

1. **`expire_on_commit=False`** on the sessionmaker. Operations and the syncer keep using ORM objects across mid-flow commits; expired-attribute reloads would be implicit IO under async. Post-operation reload sites in routers/MCP now call `db.expire_all()` before re-querying so responses still reflect exactly what the operation committed (the identity map would otherwise serve pre-operation state).
2. **The five remaining `lazy="select"` relationships flip to `raise_on_sql`** (`OktaGroup.active_group_tags`, `RoleGroupMap.active_group`, `AppGroup.app`, `OktaGroupTagMap.active_tag`/`active_app_tag_mapping`). The audit-route eager topology already covers the nested role-associated-mapping paths the old "can't load this" comment claimed were unsupported; the one operation relying on lazy loading (the `group_members_added` hook path in `ModifyRoleGroups`) now uses its already-eager-loaded groups.

Tests gain explicit `db.session.expire_all()` at staleness points: tests share one session with the app under test (production requests each get a fresh scoped session), so the identity map keeps pre-operation state that `expire_on_commit=True` used to discard implicitly.

Full suite green (540 tests), ruff + mypy clean.

### Polymorphic loader regression battery

Also adds `tests/test_polymorphic_loader_coverage.py`: a regression battery for the production `InvalidRequestError ... lazy='raise_on_sql'` errors that originally motivated demoting these five relationships. Those errors never reproduced in tests because `raise_on_sql` raises identically under test — the missing ingredient was data shape (polymorphic subtypes in nested positions: role groups associated to app groups, tagged groups inside request refs). The battery builds that shape and walks every read surface, asserting the nested payloads are populated; all 17 scenarios pass. The suite-conversion commit in #480 carries its async form.

The battery caught a real loader gap that flaked in CI: the access-request **summary** loader (`_summary_load_options`, the `GET /api/requests` list shape) placed `selectin_polymorphic` at the top level of `select(AccessRequest)` — where it targets the wrong entity and never runs — and left `requested_group`/`active_requested_group` as bare `selectinload`s, so `AppGroup.app` was never eager-loaded and serializing an app-group requested_group raised. It's now nested under each group relationship, matching `_detail_load_options` and the MCP loaders. The battery had masked it because the TestClient shares the test session (a prior request warmed `AppGroup.app` in the identity map); `_get_ok` now `expunge_all()`s before every request so any missing eager-load fails deterministically here instead of flaking by hash-seed/GC order in CI.

## PR stack

1. ~~#475 — async dependency prep (SQLAlchemy[asyncio] 2.0.50, asyncpg, aiosqlite)~~ ✅ merged
2. ~~#476 — legacy Query API -> 2.0-style statements (~350 sites)~~ ✅ merged
3. #477 — eliminate lazy loading and commit-expiration from the ORM layer ⬅️ **this PR** (now on `main`)
4. #478 — operation DB resolution out of `__init__` into `_resolve()`
5. #479 — keep `db.session` access out of spawned notification tasks

### DIFF
--- a/api/operations/modify_role_groups.py
+++ b/api/operations/modify_role_groups.py
@@ -106,7 +106,12 @@ class ModifyRoleGroups:
         self.groups_to_remove = []
         if len(groups_to_remove) > 0:
             self.groups_to_remove = db.session.scalars(
-                select(OktaGroup).where(OktaGroup.id.in_(groups_to_remove)).where(OktaGroup.deleted_at.is_(None))
+                select(OktaGroup)
+                # `app` is eager-loaded so the app-group-lifecycle hook path below
+                # can read `group.app` without tripping `lazy="raise_on_sql"`.
+                .options(selectin_polymorphic(OktaGroup, [AppGroup]), joinedload(AppGroup.app))
+                .where(OktaGroup.id.in_(groups_to_remove))
+                .where(OktaGroup.deleted_at.is_(None))
             ).all()
 
         self.owner_groups_to_remove = []
@@ -249,6 +254,7 @@ class ModifyRoleGroups:
                 )
             ).all()
 
+            groups_to_remove_by_id = {group.id: group for group in self.groups_to_remove}
             for group_id in groups_to_remove_ids:
                 removed_members_with_other_access_ids = [
                     m.user_id
@@ -261,7 +267,9 @@ class ModifyRoleGroups:
 
                 # Invoke app group lifecycle plugin hooks for removed members
                 if len(okta_members_to_remove_ids) > 0:
-                    group = db.session.get(OktaGroup, group_id)
+                    # Use the eager-loaded group (with `app`) rather than a bare
+                    # db.session.get, so the hook path can read `group.app`.
+                    group = groups_to_remove_by_id[group_id]
                     plugin_id = get_app_group_lifecycle_plugin_to_invoke(group)
                     if plugin_id is not None:
                         members_losing_access = db.session.scalars(
--- a/api/operations/modify_role_groups.py
+++ b/api/operations/modify_role_groups.py
@@ -6,12 +6,13 @@ import logging
 
 from sqlalchemy import func, or_, select, update
 from api.context import get_request_context
-from sqlalchemy.orm import selectinload
+from sqlalchemy.orm import joinedload, selectin_polymorphic, selectinload
 
 from api.extensions import db
 from api.models import (
     AccessRequest,
     AccessRequestStatus,
+    AppGroup,
     OktaGroup,
     OktaGroupTagMap,
     OktaUser,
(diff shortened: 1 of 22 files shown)


api/operations/modify_role_groups.py
eguerrant: This might be missing load options since group.app loading was changed and  `get_app_group_lifecycle_plugin_to_invoke()` checks `group.app.app_group_lifecycle_plugin`

eguerrant (commented):
I think there's two uses of `get_app_group_lifecycle_plugin_to_invoke()` in modify_group_users.py that will also fail in a similar way to how modify_role_groups.py does in the other comment. Idk how to comment on a file not included in the PR...

api/operations/modify_role_groups.py
somethingnew2-0: Fixed in 134fc0d — good catch. `db.session.get(OktaGroup, group_id)` doesn't eager-load `app`, so with `AppGroup.app` now `lazy="raise_on_sql"` this raises for app groups in the lifecycle-hook path. `self.groups_to_remove` now loads `selectin_polymorphic` + `joinedload(AppGroup.app)`, and the loop uses a by-id lookup (mirroring the existing `groups_added_by_id` add path) instead of the bare `.get()`.

Tests missed it because `AppGroup.app` is many-to-one and resolves from the warm identity map without SQL; only a cold production session emits SQL and trips `raise_on_sql`. Added a cold-session regression test that reproduces the exact `InvalidRequestError`.

🤖 Posted by Claude Code

somethingnew2-0:
Thanks @eguerrant — both confirmed and fixed in 134fc0d. The two `modify_group_users.py` cascades take `associated_group = role_associated_group_map.active_group`, and the mapping queries loaded `active_group` with a bare `joinedload` (no subtype/app), so `associated_group.app` tripped `raise_on_sql`. Both the add and remove cascades now use `selectinload(active_group)` + `selectin_polymorphic` + `joinedload(AppGroup.app)`, matching `role_group_map_options`.

Good instinct flagging a file outside the diff, too — the `AppGroup.app` flip has blast radius beyond the changed files, which is exactly why it wasn't inline-commentable. I audited all 9 `get_app_group_lifecycle_plugin_to_invoke` call sites: the two here plus the `modify_role_groups` removal path were the real gaps; the other six load `self.group` (or the router's group) with `joinedload(AppGroup.app)` and are safe.

Added `test_role_associated_lifecycle_hooks_survive_cold_session`, which runs `expunge_all()` before each membership change to force the cold load a production request sees — it covers all three cascades and fails with the exact `'AppGroup.app' is not available due to lazy='raise_on_sql'` without the fix.

🤖 Posted by Claude Code

eguerrant (approved):

PR #478, opened 2026-06-12
[async 4/7] refactor: resolve operation DB lookups as locals in execute()

somethingnew2-0:
Part of the async-SQLAlchemy migration stack; rebased directly on `main` (#475–#477 merged).

`__init__` can't be `async`, so the DB work it currently does must move behind `execute()` before the flip makes those queries awaitable. This PR relocates it — without the intermediate `_resolve()`/on-`self` scaffolding the first revision used (see review history below):

- **Lookups are local variables in the execute method.** Each `execute()` (or `_execute()` for the four `asyncio.run` ops, and each `execute_for_group()`/`execute_for_role()` for the constraint checks) performs its queries, row locks, and id lookups as locals. Nothing DB-loaded is stashed on `self`. Values genuinely read across methods stay on `self` (e.g. `ModifyGroupUsers.current_user_id`, `ModifyRoleGroups.role`/`current_user_id`, read by the notify/approve helpers).
- **`__init__` stores entity ids, not ORM model references.** Every `Model | str` constructor arg is reduced to its id in `__init__` (`self.x_id = x if isinstance(x, str) else x.id`, None-guarded for optional args); the execute method looks the entity up by that id. Holding an ORM reference across the `__init__` → `execute()` gap is fragile (it can expire/detach), and these args were only ever used to re-query by id. The few ops that previously trusted a passed-in object (the `create_*_request` requester/role/group args, the `approver_user` args) now always re-query by the stored id — equivalent, since those entities are always persisted before the op runs.

Every call site already chains `.execute()` immediately and uses the return value, so no caller reads operation state afterward — behavior is unchanged. Side benefit: the `with_for_update` row locks in the approve/reject ops now begin at execute time rather than construction time, narrowing the lock window to the actual mutation.

Full suite green (541 tests), ruff (format + check) + mypy clean.

### Review history
The first revision used a private `_resolve()` that stashed results on `self`. Per review (@somethingnew2-0, @Jeff-Rowell): dropped `_resolve()` in favor of locals-in-execute (commit `1d3bdd8`), then switched `__init__` to store ids instead of model references (commit `11237e2`). That also removes the shared-`_resolve()`-from-both-constraint-entry-points re-query concern.

## PR stack

Merge bottom-up; each PR retargets to `main` as predecessors merge.

1. ~~#475 — async dependency prep~~ ✅ merged
2. ~~#476 — legacy Query API → 2.0-style statements~~ ✅ merged
4. #478 — resolve operation DB lookups as locals in execute() ⬅️ **this PR** (on `main`)
5. #479 — keep `db.session` access out of spawned notification tasks (on `main`)

### DIFF
--- a/api/operations/constraints/check_for_reason.py
+++ b/api/operations/constraints/check_for_reason.py
@@ -19,6 +19,15 @@ class CheckForReason:
         members_to_add: list[str] = [],
         owners_to_add: list[str] = [],
     ):
+        self._group_arg = group
+
+        self.reason = reason
+
+        self.members_to_add = members_to_add
+        self.owners_to_add = owners_to_add
+
+    def _resolve(self) -> None:
+        group = self._group_arg
         self.group = db.session.scalars(
             select(OktaGroup)
             .options(
@@ -37,16 +46,12 @@ class CheckForReason:
             .where(OktaGroup.id == (group if isinstance(group, str) else group.id))
         ).first()
 
-        self.reason = reason
-
-        self.members_to_add = members_to_add
-        self.owners_to_add = owners_to_add
-
     @staticmethod
     def invalid_reason(reason: Optional[str]) -> bool:
         return reason is None or reason.strip() == ""
 
     def execute_for_group(self) -> Tuple[bool, str]:
+        self._resolve()
         if self.invalid_reason(self.reason):
             tags = [tag_map.active_tag for tag_map in self.group.active_group_tags]
             if len(self.owners_to_add) > 0:
@@ -92,6 +97,7 @@ class CheckForReason:
         return True, ""
 
     def execute_for_role(self) -> Tuple[bool, str]:
+        self._resolve()
         if type(self.group) is not RoleGroup:
             return True, ""
 
--- a/api/operations/constraints/check_for_reason.py
+++ b/api/operations/constraints/check_for_reason.py
@@ -26,9 +26,13 @@ class CheckForReason:
         self.members_to_add = members_to_add
         self.owners_to_add = owners_to_add
 
-    def _resolve(self) -> None:
-        group = self._group_arg
-        self.group = db.session.scalars(
+    @staticmethod
+    def invalid_reason(reason: Optional[str]) -> bool:
+        return reason is None or reason.strip() == ""
+
+    def execute_for_group(self) -> Tuple[bool, str]:
+        group_arg = self._group_arg
+        group = db.session.scalars(
             select(OktaGroup)
             .options(
                 selectin_polymorphic(OktaGroup, [AppGroup, RoleGroup]),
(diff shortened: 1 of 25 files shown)


api/operations/constraints/check_for_reason.py
```diff
@@ -19,6 +19,15 @@ def __init__(
...
         owners_to_add: list[str] = [],
     ):
+        self._group_arg = group
+
+        self.reason = reason
+
+        self.members_to_add = members_to_add
+        self.owners_to_add = owners_to_add
+
+    def _resolve(self) -> None:
```
Jeff-Rowell: `_resolve()` runs at the top of both `execute_for_group()` and `execute_for_role()`, so resolution is no longer memoized the way it was in `__init__`. Every call site today calls exactly one of the two, so there's no double-query in practice, but if a future caller invokes both on one instance, these queries re-run.

Worth a guard (`if getattr(self, "_resolved", False): return; self._resolved = True`) so repeated calls are no-ops ahead of the async flip.

api/operations/constraints/check_for_self_add.py
```diff
@@ -20,6 +20,16 @@ def __init__(
         members_to_add: list[str] = [],
         owners_to_add: list[str] = [],
     ):
+        self._group_arg = group
+        self._current_user_arg = current_user
+
+        self.members_to_add = members_to_add
+        self.owners_to_add = owners_to_add
+
+    def _resolve(self) -> None:
```
Jeff-Rowell: Same as the note on `check_for_reason._resolve()`: invoked at the top of both `execute_for_group()` and `execute_for_role()`, so it re-queries per call rather than memoizing once in `__init__`.

api/operations/constraints/check_for_reason.py
somethingnew2-0: I'm not sure I like the `_resolve()` pattern anyways, maybe the db loading should all just be included in `execute()`?

somethingnew2-0: Agreed — done in 1d3bdd8. Dropped `_resolve()` across all 25 operation/constraint classes; each `execute()` (or `_execute()` for the four `asyncio.run` ops, and each `execute_for_group`/`execute_for_role` here) now does its DB lookups as local variables. `__init__` keeps only the raw constructor args. Values genuinely read across methods stay on `self`.

🤖 Posted by Claude Code

somethingnew2-0: Fixed in 1d3bdd8 — `_resolve()` is gone. `execute_for_group()` and `execute_for_role()` each now load `group` as a local, so there's no shared resolution invoked from both entry points and nothing to double-query.

🤖 Posted by Claude Code

api/operations/constraints/check_for_self_add.py
somethingnew2-0: Same fix as the `check_for_reason` thread — addressed in 1d3bdd8. Each entry point now resolves its own locals instead of calling a shared `_resolve()`.

🤖 Posted by Claude Code

Jeff-Rowell (dismissed):

api/operations/constraints/check_for_reason.py
```diff
@@ -19,7 +19,20 @@ def __init__(
         members_to_add: list[str] = [],
         owners_to_add: list[str] = [],
     ):
-        self.group = db.session.scalars(
+        self._group_arg = group
```
somethingnew2-0: Since a group be an OktaGroup model object or a group id (str), should we store `self.group_id = (group if isinstance(group, str) else group.id)` during init instead of doing so later when the OktaGroup model might be invalidated? Can this be fixed for all of the other operations as well, so that ids are stored during init instead of SQLAlchemy model references?

somethingnew2-0: Good call — done in 347d3ec. Every operation/constraint that takes a `Model | str` arg now extracts the id in `__init__` (`self._<x>_id = <x> if isinstance(<x>, str) else <x>.id`, None-guarded for optional args) and looks the entity up by that id in `execute()`, so we never hold an ORM reference across the `__init__` → `execute()` gap.

A few ops previously trusted a passed-in object (the `create_*_request` requester/role/group args, the `approver_user` args); those now always re-query by the stored id. The entities are always persisted before the op runs, so it's equivalent and removes the stale-reference risk. List-of-id and scalar args (`members_to_add`, `tags`, `ending_at`, …) are left as-is since they're already ids, not model refs.

🤖 Posted by Claude Code

Jeff-Rowell (dismissed):

somethingnew2-0:
@Jeff-Rowell sorry for the review thrash, I thought Claude renamed all the variables how I wanted, but it needed to take a second pass. PR is fully ready for review now.

Jeff-Rowell (approved):

PR #479, opened 2026-06-12
[async 5/7] refactor: keep db.session access out of spawned notification tasks

somethingnew2-0:
Part of the async-SQLAlchemy migration stack; rebased directly on `main` (#475–#478 merged).

`ModifyGroupUsers` and `ModifyRoleGroups` previously created asyncio tasks whose coroutines queried `db.session` (requester lookup + `get_all_possible_request_approvers`) before invoking the notification hook. Under async SQLAlchemy, ContextVar-scoped sessions are inherited by child tasks, and concurrent operations on one `AsyncSession` raise `InvalidRequestError` — so spawned tasks may only perform network I/O.

The approve helpers now just mutate and return the request; after the final commit, the main coroutine resolves requester/approvers/group sequentially and spawns hook-only coroutines that take those pre-resolved values as arguments.

Also fixes a latent bug in `ModifyRoleGroups`: its pending-request queries never eager-loaded `AccessRequest.requested_group` / `RoleRequest.requester_role` / `RoleRequest.requested_group` (all `raise_on_sql`), so the notify task raised — and `asyncio.wait` silently swallowed it — whenever an approved request had been created in a different session than the one approving it. The notifications now always have their relationships loaded. `resolved_at` (expired at flush after the `func.now()` assignment) is explicitly refreshed so hooks see a concrete timestamp.

Full suite green (541 tests), ruff (format + check) + mypy clean.

## PR stack

Merge bottom-up; each PR retargets to `main` as predecessors merge.

1. ~~#475 — async dependency prep~~ ✅ merged
2. ~~#476 — legacy Query API → 2.0-style statements~~ ✅ merged
4. ~~#478 — resolve operation DB lookups as locals in execute()~~ ✅ merged
5. #479 — keep `db.session` access out of spawned notification tasks ⬅️ **this PR** (on `main`)


api/operations/modify_role_groups.py
```diff
@@ -586,6 +592,32 @@ async def _execute(self) -> RoleGroup:
         # Commit all changes
         db.session.commit()
 
+        # Resolve everything the notification hooks need on the main coroutine
+        # before spawning tasks: spawned tasks must only perform network I/O,
+        # never db.session access (a session cannot be used concurrently, and
+        # under async SQLAlchemy that raises rather than interleaving).
+        for access_request in approved_access_requests:
```
matthew-bass: nit: Do we also want to check `self.notify` for approved_access_requests similar to the `approved_role_requests` below this? Or, is it by design that individual access requests will always get DM'd?

matthew-bass (dismissed):
LGTM! had a small finding around notifying all `approved_access_requests` that seemed introduced in a previous pr, but nothing blocking

api/operations/modify_role_groups.py
somethingnew2-0: No, it's not by design — it's an oversight, and `ModifyGroupUsers` is the one that has it right: it gates its access-request completion DMs on `self.notify`, while `ModifyRoleGroups` only gated the role-request loop. So `notify=False` still leaked the individual access-request DMs from the auto-approval cascade.

Fixed in d15e3d4 — both loops now sit under a single `if self.notify:`, with a test asserting `notify=False` suppresses both completion notifications while the requests still resolve to APPROVED.

🤖 Posted by Claude Code

somethingnew2-0: Also added some additional tests, so we don't accidentally ignore the `self.notify` gate in the future for other operations.

matthew-bass (approved):

PR #480, opened 2026-06-12
[async 6/7] Switch to async SQLAlchemy end-to-end

somethingnew2-0:
The atomic async-SQLAlchemy flip — completes POST_MIGRATION_TODO items 2, 3, 11, and 15 (stacked on #479; #475–#478 are merged and #479 is the base, so this diff is mostly `async`/`await` insertion).

## Production

- **Session layer**: `async_sessionmaker` + `async_scoped_session` keyed on the existing `_session_scope` ContextVar (item 3). The `db.session` global proxy now yields an `AsyncSession`. ContextVar scoping (not `asyncio.current_task`) is load-bearing: `BaseHTTPMiddleware.call_next` runs handlers in a child task, and context copies propagate where task identity would not.
- **Engine**: `create_async_engine` with a driver-rewrite map so existing deployment URLs (`postgresql://`, legacy `postgresql+pg8000://`, `sqlite://`) keep booting on asyncpg/aiosqlite; Cloud SQL goes through `connector.connect_async` + asyncpg IAM auth.
- **OktaService** (item 11): every method is natively async under its primary name; the `asyncio.run`-per-call sync wrappers and duplicated `async_*` variants are gone. One event loop end-to-end. The SDK stays at `okta==2.9.8` — the v3 upgrade is tracked as new TODO item 11a behind the facade.
- **Routers/operations/auth/syncer/MCP**: `async def` + awaits throughout; pagination via `apaginate`; the four fan-out operations keep their `create_task` Okta parallelism on the request loop. Spawned tasks perform network I/O only — never `db.session`.
- **Sync plugin hooks** (their async conversion is deferred to item 18) stay synchronous and are invoked from the async operations. `ModifyGroupDetails`'s post-commit `group_updated` hook runs through the `AsyncSession.run_sync` greenlet bridge; the other session-bound `app_group_lifecycle` hooks (`group_created`/`group_deleted`/`group_members_added`/`group_members_removed`) are called directly with their relationships eager-loaded (so they read without emitting lazy SQL). Notification and conditional-access hooks are invoked inline on the request loop — moving their network I/O off the loop is left to item 18.
- **CLI**: Click commands stay sync; `_with_app_context` drives the async body with one `asyncio.run` per invocation. Plugin-contributed CLI commands that touch `db.session` synchronously will need the same treatment (documented in TODO item 18).
- **Alembic**: async-template `env.py` (`connection.run_sync`) over the same engine builder, so Cloud SQL migrations keep working; `pg8000` dropped.

Rebased onto the current #479: the App-prefix enforcement (#474) and nested `active_tag` eager-load (#492) paths that merged to `main` after the original flip are async-converted here too.

## Tests (item 15)

pytest-asyncio 1.4.0 in `asyncio_mode=auto` with function-scoped loops; the `db` fixture builds an aiosqlite StaticPool in-memory engine; the client is an `httpx.AsyncClient` over `ASGITransport` (required, not stylistic: requests must run on the test's own loop because the engine is bound to it). Most tests are `async def`; `coroutine was never awaited` is escalated to an error as a missed-await tripwire.

**asyncpg datetime strictness**: the schema stores naive-UTC timestamps, but request bodies and the expiring-access/time-limit paths bind timezone-aware UTC datetimes; pg8000/sqlite coerced those silently while asyncpg raises DataError. DateTime columns now go through a `NaiveUTCDateTime` TypeDecorator that converts aware values to UTC and strips tzinfo at bind time; naive values and `date` binds pass through unchanged.

## Verification

- Full suite: **542 passed**; ruff (format + check) + mypy clean across 166 files
- The async conversion introduces no wire-format changes (OpenAPI surface unchanged)
- `alembic upgrade head` / `check` / `downgrade -1` green through the async engine
- Real uvicorn boot: health check, paginated lists with eager-loaded polymorphic data, and a persisted mutation
- `access notify` CLI end-to-end through the new `asyncio.run` boundary

## PR stack

1. ~~#475 — async dependency prep (SQLAlchemy[asyncio] 2.0.50, asyncpg, aiosqlite)~~ ✅ merged
2. ~~#476 — legacy Query API → 2.0-style statements (~350 sites)~~ ✅ merged
4. ~~#478 — operation DB lookups resolved as locals in `execute()`~~ ✅ merged
5. #479 — keep `db.session` access out of spawned notification tasks ⬅️ **base of this PR**
6. #480 — **the async flip** (production + tests + docs) ⬅️ **this PR**
7. #481 — surface failed Okta/notification fan-out tasks (stacks on this PR)


api/database.py
```diff
@@ -1,61 +1,78 @@
...
             db_user=settings.DATABASE_USER,
             db_name=settings.DATABASE_NAME,
             uses_public_ip=settings.DATABASE_USES_PUBLIC_IP,
         )
         # CloudSQL connector creator handles connection details
-        url = "postgresql+pg8000://"
+        url = make_url("postgresql+asyncpg://")
     else:
-        url = settings.SQLALCHEMY_DATABASE_URI or "sqlite:///instance/access.db"
+        url = to_async_url(settings.SQLALCHEMY_DATABASE_URI or "sqlite:///instance/access.db")
```
Jeff-Rowell, line 101: `to_async_url` forwards query params unchanged, and `build_async_engine` sets no `ssl`/`connect_args` on this non-CloudSQL path. SQLAlchemy's asyncpg dialect passes query params straight into `asyncpg.connect()`, which accepts `ssl=`, not `sslmode=`.

The documented non-CloudSQL deployment path in `.env.production.example` and `examples/kubernetes/deployment.yaml` will fail to boot with `TypeError: connect() got an unexpected keyword argument 'sslmode'`. This is fail-closed (not a silent TLS downgrade).

Suggested fix: translate `sslmode` into asyncpg's `ssl=` connect arg here. asyncpg accepts `disable`/`allow`/`prefer`/`require` as plain strings; the cert-verifying modes (`verify-ca`/`verify-full`) can't pass their cert paths through asyncpg's connect kwargs, so we build an `SSLContext` explicitly and raise if the CA bundle (or a half-specified client cert/key pair) is missing.

```suggestion
    else:
        url = to_async_url(settings.SQLALCHEMY_DATABASE_URI or "sqlite:///instance/access.db")
        # asyncpg takes `ssl=` (an SSLContext or a libpq-style mode string), not the
        # `sslmode=` query param that pg8000/psycopg2 URLs carry. Translate it so
        # existing DATABASE_URIs keep booting on asyncpg.
        if url.drivername == "postgresql+asyncpg" and "sslmode" in url.query:
            query = dict(url.query)
            sslmode = query.pop("sslmode")
            sslrootcert = query.pop("sslrootcert", None)
            sslcert = query.pop("sslcert", None)
            sslkey = query.pop("sslkey", None)
            if sslmode in ("verify-ca", "verify-full"):
                # Cert-verifying modes need a CA bundle (and, for mutual TLS, a
                # client cert/key). asyncpg can't take these as connect kwargs, so
                # build the SSLContext here and fail loudly if inputs are missing.
                import ssl

                if not sslrootcert:
                    raise RuntimeError(
                        f"DATABASE_URI sslmode={sslmode} requires sslrootcert to verify the server certificate"
                    )
                if bool(sslcert) != bool(sslkey):
                    raise RuntimeError(
                        "DATABASE_URI sslcert and sslkey must be set together for client certificate auth"
                    )
                ssl_context = ssl.create_default_context(cafile=sslrootcert)
                ssl_context.check_hostname = sslmode == "verify-full"
                if sslcert and sslkey:
                    ssl_context.load_cert_chain(certfile=sslcert, keyfile=sslkey)
                kwargs["connect_args"] = {"ssl": ssl_context}
            else:
                # disable / allow / prefer / require: asyncpg accepts these as strings.
                kwargs["connect_args"] = {"ssl": sslmode}
            url = url.set(query=query)
```

Note: `import ssl` is kept local to the branch so the suggestion applies cleanly; feel free to hoist it to the module's stdlib imports. `create_default_context` verifies the chain against `sslrootcert` and, for `verify-full`, also checks the hostname; `verify-ca` disables the hostname check.

README.md
```diff
@@ -252,7 +252,7 @@ The `.env.production` file is where you configure the application.
 
 - `OKTA_DOMAIN`: Specifies the [Okta](https://okta.com) domain to use.
 - `OKTA_API_TOKEN`: Specifies the [Okta](https://okta.com) [API Token](https://developer.okta.com/docs/api/openapi/okta-management/management/tag/ApiToken/) to use.
-- `DATABASE_URI`: Specifies the Database connection URI. **Example:** `postgresql+pg8000://<POSTGRES_USER>:<POSTGRES_PASSWORD>@postgres:5432/<DB_NAME>`.
+- `DATABASE_URI`: Specifies the Database connection URI. **Example:** `postgresql://<POSTGRES_USER>:<POSTGRES_PASSWORD>@postgres:5432/<DB_NAME>`. The app runs on async SQLAlchemy: `postgresql://` (and legacy `postgresql+pg8000://`) URIs are normalized to the `asyncpg` driver, `sqlite://` to `aiosqlite`. Note that asyncpg does not accept `sslmode=` URL parameters (use `ssl=`), and deployments behind transaction-mode PgBouncer should disable asyncpg's prepared-statement cache (`statement_cache_size=0`).
```
Jeff-Rowell: If the `sslmode`->`ssl` translation suggested on `api/database.py` is accepted, this note becomes inaccurate: `sslmode=` would then be supported (translated to asyncpg's `ssl=`). Suggested revision (pairs with that change):

```suggestion
- `DATABASE_URI`: Specifies the Database connection URI. **Example:** `postgresql://<POSTGRES_USER>:<POSTGRES_PASSWORD>@postgres:5432/<DB_NAME>`. The app runs on async SQLAlchemy: `postgresql://` (and legacy `postgresql+pg8000://`) URIs are normalized to the `asyncpg` driver, `sqlite://` to `aiosqlite`. libpq-style `sslmode=` parameters are translated to asyncpg's `ssl=` connect arg; `verify-ca`/`verify-full` require `sslrootcert` (and, for mutual TLS, `sslcert`/`sslkey`). Deployments behind transaction-mode PgBouncer should disable asyncpg's prepared-statement cache (`statement_cache_size=0`).
```

If the code suggestion is not taken, leave this note as-is.

Jeff-Rowell (changes requested):

api/operations/create_group.py
```diff
@@ -109,31 +117,33 @@ def execute(self, *, _group: Optional[T] = None) -> T:
...
                 )
-            db.session.commit()
+            await db.session.commit()
 
         # Invoke app group lifecycle plugin hook, if configured
         plugin_id = get_app_group_lifecycle_plugin_to_invoke(self.group)
         if plugin_id is not None:
             try:
                 hook = get_app_group_lifecycle_hook()
                 hook.group_created(session=db.session, group=self.group, plugin_id=plugin_id)
```
eguerrant: The app group lifecycle code expects a `Session` not an `AsyncSession` (eg. https://github.com/discord/access/blob/e3d82ac0233ccd979fc4986fa22f8939d65a4aad/api/plugins/app_group_lifecycle.py#L163)

Should change all calls to the plugin to invoke the hook through await db.session.run_sync(...), passing a callable that receives the bridged sync session and forwards it plus the same group/members/plugin_id arguments to the hook

api/syncer.py
```diff
@@ -509,30 +532,41 @@ def expiring_access_notifications_user() -> None:
...
         if user in grouped_next_week:
-            notification_hook.access_expiring_user(
-                # TODO eventually clean this up, leaving for now for backwards compatibility
-                groups=grouped_tomorrow_old[user] + grouped_next_week_old[user],
-                user=user,
-                expiration_datetime=None,
-                okta_user_group_members=grouped_tomorrow[user] + grouped_next_week[user],
+            # Notification hooks are sync plugin code that may perform blocking
+            # network I/O, so run them on a worker thread.
+            await anyio.to_thread.run_sync(
```
eguerrant: Notification hooks are offloaded to a worker thread via anyio.to_thread.run_sync while holding AsyncSession-bound ORM objects.
Only active_user/active_group/active_role_group are eager-loaded on the passed rows.

Need to fix here and in other expiring_access_notification functions

api/database.py
somethingnew2-0: Good catch — fixed in 820296f. `build_async_engine` now translates libpq `sslmode=` into asyncpg's `ssl=` connect arg on the non-CloudSQL asyncpg path:

- `disable`/`allow`/`prefer`/`require` → forwarded as the `ssl=` mode string
- `verify-ca`/`verify-full` → an `ssl.SSLContext` (verify-full also sets `check_hostname`), which requires `sslrootcert` and, for mutual TLS, both `sslcert`/`sslkey` — it raises rather than silently downgrading if those are missing or only partly specified.

`sslmode` and the consumed cert params are stripped from the URL so asyncpg never sees them. Added `tests/test_database_ssl.py` covering require / absent / non-asyncpg passthrough plus the two fail-loud cases.

🤖 Posted by Claude Code

README.md
somethingnew2-0: Done in 820296f — updated the note to reflect the translation. It now states that libpq `sslmode=` params are translated to asyncpg's `ssl=` connect arg, that `verify-ca`/`verify-full` require `sslrootcert` (plus `sslcert`/`sslkey` for mutual TLS), and flags `statement_cache_size=0` for deployments behind transaction-mode PgBouncer.

🤖 Posted by Claude Code

api/operations/create_group.py
somethingnew2-0: Fixed in 6a9c689. Every app-group lifecycle hook invocation now goes through `db.session.run_sync(lambda s: hook.<hook>(session=s, ...))` — the greenlet bridge hands the plugin a real sync `Session`, matching the hookspec — instead of passing the `AsyncSession`. This covers `group_created`/`group_deleted` (create_group, delete_group, modify_group_type) and `group_members_added`/`group_members_removed` (modify_group_users, modify_role_groups); `group_updated` already used this pattern.

The lifecycle-plugin tests' dummy plugin now reads through that sync `Session` (and reads `group.app`) before recording each call, so if any site regresses to passing the raw `AsyncSession` — or an operation stops eager-loading `AppGroup.app` — the hook raises and the recorded-call assertions fail.

🤖 Posted by Claude Code

api/syncer.py
somethingnew2-0: Fixed in 53c2ecd. I went a different route than eager-load-more-and-expunge once I looked closer: every relationship on these rows is `lazy="raise_on_sql"`, so a hook can't lazily load past what's eager-loaded regardless of thread or greenlet — and the `active_*` joinedloads already seed the identity map, so the plain `.group`/`.user`/`.role_group` many-to-ones resolve from it via `use_get` without emitting SQL. The real hazard was the cross-thread handoff itself: reading those session-bound objects (including the identity-map `use_get` lookups) from an `anyio` worker thread.

So rather than `anyio.to_thread.run_sync`, the syncer now invokes these hooks via `db.session.run_sync` — on the session's own greenlet, so the objects are never touched across a thread boundary. These three functions only run from the CLI `asyncio.run` boundary (`access sync` / `notify`), never a request handler, so briefly blocking that batch loop while a hook does network I/O is fine.

Added tests asserting the hook runs on the loop thread (fails if reverted to `to_thread`) and that it can read the non-active `.group`/`.user`/`.role_group` off the association rows; if a future change narrows the eager-loads, those reads raise and the tests fail. Also corrected the TODO note (item 18) to describe the bridging the flip actually ships.

🤖 Posted by Claude Code

api/operations/create_group.py
```diff
@@ -109,31 +117,37 @@ def execute(self, *, _group: Optional[T] = None) -> T:
...
+                await db.session.run_sync(
+                    lambda s: hook.group_created(session=s, group=self.group, plugin_id=plugin_id)
+                )
+                await db.session.commit()
             except Exception:
                 logging.getLogger("api").exception(
                     f"Failed to invoke group_created hook for group {self.group.id} with plugin '{plugin_id}'"
                 )
-                db.session.rollback()
+                await db.session.rollback()
```
Jeff-Rowell: nit: This try/get-hook/`run_sync`/commit/except-log/rollback block is duplicated near-verbatim across 10 call sites in 6 files (`delete_group.py`, `modify_group_details.py`, `modify_group_users.py` x2, `modify_group_type.py` x2, `modify_role_groups.py` x2, plus this one), right down to the same inline comment. Since this PR touched every one of them to add the `run_sync` bridge, it seems like a natural point to extract a shared helper, e.g.:

```python
async def invoke_app_group_lifecycle_hook(hook_method: str, **kwargs) -> None:
    ...
```

so future changes to this bridging logic (a timeout around `run_sync`, a metric on failures, etc.) only need to happen in one place instead of 10.

somethingnew2-0: Good call — done in cbf3b04. Extracted `invoke_app_group_lifecycle_hook(hook_method, *, group, **kwargs)` into `app_group_lifecycle.py` and collapsed all 9 call sites to a single `await`, e.g.:

```python
await invoke_app_group_lifecycle_hook("group_created", group=self.group)
```

The helper owns the whole `plugin-gate → run_sync bridge → commit → except-log-rollback` sequence, so a future change (a `run_sync` timeout, a failure metric) only lands in one place. Membership sites pass `members=`; `group_updated` passes `old_name`/`old_description`. Net −99 lines.

One nuance I kept: the two `modify_role_groups` sites only fetch the affected `OktaUser` rows when a lifecycle plugin is actually configured, so I left that `get_app_group_lifecycle_plugin_to_invoke(group) is not None` gate around the query and delegate just the invocation to the helper — no unconditional extra query when no plugin is present.

🤖 Posted by Claude Code

api/extensions.py
```diff
@@ -135,48 +150,35 @@ def drop_all(self) -> None:
 Db = _DB
 
 
-# Process-wide singleton Connector, shared by every engine/creator for the
```
Jeff-Rowell, line 138: Is this intentional to revert the singleton that was added in #501 ?

Jeff-Rowell (commented):
I think this is looking ready once the singleton comment is addressed. The only other finding I had is regarding the sync plugin hooks which looks like its [deferred as a TODO](https://github.com/discord/access/blob/pcollins/async-2-flip/POST_MIGRATION_TODO.md?plain=1#L126).

api/extensions.py
somethingnew2-0: Intentional — it carries #501's fix into the async world rather than reverting it. The reuse that mattered (one `Connector` for the life of the worker, so the background cert-refresh + refresh-ahead cache aren't torn down mid-flight → the `ServerDisconnectedError` #501 chased) is preserved: `get_cloudsql_async_conn` builds one `Connector` lazily on first connect and reuses it for the engine's lifetime — one engine per worker, so effectively process-wide, same as before.

What changed is that the single instance is a **creator-scoped** lazy singleton instead of a module global. The async `Connector` binds to the running loop (`Connector(loop=asyncio.get_running_loop())`), and a module global would pin to whichever loop touched it first — fine for the server's long-lived loop, but it breaks the CLI (`access sync`/`notify`) and the tests, which each run on a fresh loop per `asyncio.run`. Scoping to the creator gives each engine a Connector bound to its own loop while still reusing it.

#501's other two properties carry over: lazy build (Connector created in the worker on first DB use, post-fork) and no per-connection teardown. The `threading.Lock` is gone because the async creator runs on the single event-loop thread — the `if connector is None` check and assign have no `await` between them, so they can't interleave.

Net for the server: one Connector per worker reused for its lifetime, identical to #501. Happy to switch to a loop-keyed module-level cache if you'd want one instance shared across multiple engines in a process, but there's a single engine per worker today so it'd be the same object either way.

🤖 Posted by Claude Code

Jeff-Rowell (approved):

PR #481, opened 2026-06-12
[async 7/7] fix: surface failed Okta/notification fan-out tasks instead of dropping them

somethingnew2-0:
The async flip (#480) has merged; this is the final piece of the async-SQLAlchemy stack, now based on `main`.

The four fan-out operations (`ModifyGroupUsers`, `ModifyRoleGroups`, `DeleteUser`, `DeleteGroup`) drained their `asyncio.create_task` lists with `asyncio.wait`, which does not propagate task exceptions — a failed Okta call or notification hook vanished silently, so partial Okta failures during membership changes were invisible. (This predates the async migration; it was preserved through the flip for behavior compatibility, and flagged there as a follow-up.)

They now drain through a shared `drain_fan_out_tasks` helper (`asyncio.gather` with `return_exceptions=True`) that logs every failure at ERROR with the operation context and traceback. Failures still do not fail the request: the local DB state is committed before the tasks are awaited, so raising would turn an Okta straggler into a misleading request failure — and the syncer reconciles Okta drift on its next run. Per the concurrency rule in `api/extensions.py`, these tasks only perform network I/O (Okta calls, notification hooks), never `db.session` access.

The guaranteed behavior is covered end to end: a failing `okta.add_user_to_group` during a member-add still returns 200 with the membership committed, a failing `okta.delete_group` still soft-deletes locally, and in both cases the failure is logged with context, never propagates, and never cancels sibling tasks.

## PR stack

The async-SQLAlchemy migration landed as #475–#480 (all merged); this is the final follow-up, now targeting `main`:

1. #475 — async dependency prep (`SQLAlchemy[asyncio]` 2.0.50, asyncpg, aiosqlite) — merged
2. #476 — legacy Query API → 2.0-style statements — merged
3. #477 — eliminate lazy loading and commit-expiration from the ORM layer — merged
4. #478 — operation DB resolution out of `__init__` into `_resolve()` — merged
5. #479 — keep `db.session` access out of spawned notification tasks — merged
6. #480 — **the async flip** (production + tests + docs) — merged
7. #481 — surface failed Okta/notification fan-out tasks ⬅️ **this PR**

### DIFF
--- a/api/operations/_fan_out.py
+++ b/api/operations/_fan_out.py
@@ -1,0 +1,55 @@
+"""Draining helper for the Okta/notification task fan-out in operations.
+
+The fan-out operations (`ModifyGroupUsers`, `ModifyRoleGroups`, `DeleteUser`,
+`DeleteGroup`) spawn Okta API calls and notification dispatch with
+`asyncio.create_task` and drain them here after the local DB state has
+committed. A failed task therefore must not fail the request — the
+authoritative state change already happened — but it must not vanish either:
+the previous idiom (`asyncio.wait(tasks)` whose result was discarded) dropped
+task exceptions on the floor, which made partial Okta failures during
+membership changes invisible.
+
+Per the concurrency rule in `api/extensions.py`, these tasks only perform
+network I/O (Okta calls, notification hooks) — never `db.session` access.
+"""
+
+from __future__ import annotations
+
+import asyncio
+import logging
+from typing import Any
+
+logger = logging.getLogger("api")
+
+
+async def drain_fan_out_tasks(tasks: list[asyncio.Task[Any]], context: str) -> None:
+    """Await every fan-out task and log each failure with its context.
+
+    Failures are logged at WARNING (with the traceback) and swallowed: the
+    DB commit these tasks trail behind has already happened, so raising
+    would turn an Okta/notification straggler into a misleading request
+    failure. These are best-effort, on-demand Okta calls and the syncer
+    cronjob reconciles any resulting drift on its next run, so they're a
+    WARNING rather than an ERROR — an ERROR here is just Sentry noise for a
+    condition the system already recovers from eventually.
+
+    Drains with ``asyncio.wait`` (not ``asyncio.gather``) so that a
+    cancellation of the awaiting request — client disconnect, request timeout,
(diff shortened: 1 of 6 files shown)


api/operations/_fan_out.py
```diff
@@ -0,0 +1,45 @@
...
+    failure. These are best-effort, on-demand Okta calls and the syncer
+    cronjob reconciles any resulting drift on its next run, so they're a
+    WARNING rather than an ERROR — an ERROR here is just Sentry noise for a
+    condition the system already recovers from eventually.
+    """
+    if not tasks:
+        return
+    results = await asyncio.gather(*tasks, return_exceptions=True)
+    for result in results:
+        if isinstance(result, BaseException):
```
Jeff-Rowell: Previously this used `asyncio.wait()` which waits for coroutines to complete. Moving that call to `asyncio.gather()` will cancel in-flight coroutines if/when there's a client disconnect, request timeout, SIGTERM during rolling deploy, etc etc.

Is this intended for some reason? If not we should probably still use `asyncio.wait()`

```suggestion
    done, _ = await asyncio.wait(tasks)
    for task in done:
        result = task.exception()
        if result is not None:
```

Jeff-Rowell (changes requested):

api/operations/_fan_out.py
somethingnew2-0: Good catch — not intended. Fixed in e869673e6cb6c475f8b0d566525b39a34434f3aa: back to `asyncio.wait(tasks)`, reading the outcome off the finished tasks via `task.exception()`, so a cancellation of the awaiting request doesn't tear down the in-flight Okta calls (`gather` propagates the cancel to its children; `wait` leaves them running) — which is what we want, since the DB change they trail is already committed. Failures are still surfaced.

```python
done, _ = await asyncio.wait(tasks)
for task in done:
    exc = task.exception()
    if exc is not None:
        logger.warning(...)
```

Also added a docstring note on why `wait` over `gather` (so it doesn't get "simplified" back), plus a regression test that cancels the drain mid-flight and asserts the in-flight task still runs to completion and isn't cancelled — it fails if swapped to `gather`.

🤖 Posted by Claude Code

Jeff-Rowell (approved):

PR #482, opened 2026-06-12
[async 8/8] test: reproduce the Flask-era lazy-load Sentry issues with polymorphic data

somethingnew2-0:
Regression battery for the four Flask-era Sentry issues (`ACCESS-FLASK-4N`, `ACCESS-FLASK-44`, `ACCESS-FLASK-4E`) — the `InvalidRequestError: ... lazy='raise_on_sql'` production failures that motivated demoting five relationships to `lazy="select"`, which #477 flips back to `raise_on_sql`.

**Why they never reproduced locally:** `raise_on_sql` raises identically under test, so the only production-only ingredient was *data shape*. The failing serializations required polymorphic subtypes in nested positions — a role group whose associated groups are **app groups**, tagged groups inside request refs, app groups inside membership rows — which the suite never built. The error chain in the reports maps exactly onto the schema nesting: `RoleGroup.active_role_associated_group_member_mappings` (4N) → each mapping's `active_group` (44) → that group's `app` when it's an AppGroup (4E), plus `active_group_tags` (4E) on every rich group ref.

This battery builds that shape once (tagged role group with member+owner associations to a tagged app group and a tagged plain group, via-role memberships, pending access/role requests targeting both group types) and walks every read surface: groups list + detail per type, roles, users, access-request detail with role-group and app-group targets, role requests, tags, apps, and `/api/audit/users` + `/api/audit/groups` both unfiltered and with the `user_id`/`group_id` filters the old model comments singled out. Assertions verify the nested payloads are **populated**, not just 200 — so both failure modes are covered: a missing eager load fails loudly instead of paging on-call, and a silently-empty collection fails instead of shipping wrong data.

**Result: all 17 scenarios pass against the current loaders** — the eager-load topology from the FastAPI migration plus #477's fixes closes every gap behind the four Sentry issues. Full suite: 536 passed.

## PR stack

1. #475 — async dependency prep (SQLAlchemy[asyncio] 2.0.50, asyncpg, aiosqlite)
2. #476 — legacy Query API -> 2.0-style statements (~350 sites)
3. #477 — eliminate lazy loading and commit-expiration from the ORM layer
4. #478 — operation DB resolution out of `__init__` into `_resolve()`
5. #479 — keep `db.session` access out of spawned notification tasks
8. #482 — regression battery for the Flask-era lazy-load Sentry issues ⬅️ **this PR**

### DIFF
--- a/.github/workflows/db-migrations.yml
+++ b/.github/workflows/db-migrations.yml
@@ -20,7 +20,7 @@ jobs:
         include:
           - python-version: "3.13"
             database: postgresql
-            database-uri: postgresql+pg8000://postgres:postgres@localhost:5432/postgres
+            database-uri: postgresql://postgres:postgres@localhost:5432/postgres
           - python-version: "3.13"
             database: sqlite
             database-uri: sqlite:///migrations_ci.db
--- a/POST_MIGRATION_TODO.md
+++ b/POST_MIGRATION_TODO.md
@@ -8,23 +8,6 @@ Items grouped roughly by surface area; ordering within each group is rough prior
 
 ## Database / SQLAlchemy
 
-### 2. Switch to async SQLAlchemy
-
-The migration kept SQLAlchemy synchronous. To go async:
-- Engine: `create_engine` → `create_async_engine`
-- Sessionmaker: `sessionmaker` → `async_sessionmaker`
-- `get_db`: `Generator[Session, …]` → `AsyncGenerator[AsyncSession, …]`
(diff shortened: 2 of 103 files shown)


somethingnew2-0:
Folded into #477 per review feedback: the battery now lands as a sync test in the same PR that returns the five relationships to raise_on_sql (and the suite-conversion commit in #480 carries its async form), with the internal incident identifiers removed.

🤖 Posted by Claude Code

PR #491, opened 2026-06-17
feat: added a generic way to directly link to IdP for Access users and groups

Jeff-Rowell:
Adds an "Open in IdP" button to the group and user detail pages that deep-links out to the configured identity provider's console for that resource, using the underlying Okta ID. The link target is driven by configurable URL templates (IDP_NAME, IDP_USER_URL_TEMPLATE, IDP_GROUP_URL_TEMPLATE) so it works for any IdP backend, and the button is hidden entirely when no template is configured. No backend changes; config defaults to empty so existing deployments are unaffected until they opt in.

Users verification:

Screenshot 2026-06-17 at 2 46 22 PM

Groups verification:

Screenshot 2026-06-17 at 2 43 44 PM

### DIFF
--- a/src/pages/users/Read.tsx
+++ b/src/pages/users/Read.tsx
@@ -55,7 +55,7 @@ import Loading from '../../components/Loading';
 import RemoveGroupsDialog, {RemoveGroupsDialogParameters} from '../roles/RemoveGroups';
 import RemoveOwnDirectAccessDialog, {RemoveOwnDirectAccessDialogParameters} from '../groups/RemoveOwnDirectAccess';
 import {groupBy, displayUserName} from '../../helpers';
-import {canManageGroup, isGroupOwner} from '../../authorization';
+import {canManageGroup, isGroupOwner, isAccessAdmin} from '../../authorization';
 import {EmptyListEntry} from '../../components/EmptyListEntry';
 import MembershipChip from '../../components/MembershipChip';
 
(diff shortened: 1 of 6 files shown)


somethingnew2-0:
Since the Okta links would only be accessible to Okta admin, we should probably limit displaying them to App-Access-Owners exclusively instead using the `isAccessAdmin()` helper in https://github.com/discord/access/blob/main/src/authorization.tsx#L24 ?

src/pages/users/Read.tsx
```diff
@@ -642,6 +644,7 @@ export default function ReadUser() {
                       <AuditIcon />
                     </IconButton>
                   </Tooltip>
+                  <IdpLinkButton url={idpUserUrl(user.id)} />
```
matthew-bass, line 647: While I'd like this change, iiuc anyone who isn't an IDP admin would have this button 401

I wonder if we can gate showing this button based on if someone's an [AccessAdmin](https://github.com/discord/access/blob/2ce2b4bbc23ab028e050d1ce65a05cdc7faafba9/src/authorization.tsx#L24-L33), since i'm guessing that roughly matches the set of people i'd expect to be IDP admins?

Jeff-Rowell:
> Since the Okta links would only be accessible to Okta admin, we should probably limit displaying them to App-Access-Owners exclusively instead using the `isAccessAdmin()` helper in https://github.com/discord/access/blob/main/src/authorization.tsx#L24 ?

Good call, updated to check `isAccessAdmin()` with a short-circuit evaluation before showing the button.

src/pages/users/Read.tsx
matthew-bass: oh, ignore me, seems like i was slow on the draw

matthew-bass (approved):

PR #493, opened 2026-06-18
feat: serve IdP deep-link config from /api/config at runtime

Jeff-Rowell:
## Summary
Adds a runtime `GET /api/config` endpoint and switches the "Open in IdP" button (#491) to read its `IDP_*` config from it, instead of the build-time-baked frontend `ACCESS_CONFIG`.

## Motivation

#491 reads `IDP_NAME` / `IDP_USER_URL_TEMPLATE` / `IDP_GROUP_URL_TEMPLATE` from the frontend `AccessConfig`, which Vite inlines into the JS bundle at build time. Because the values are baked into the bundle when the image is built, a single published image cannot carry different IdP URLs for different environments. Deployments that share one built image can therefore only ever point at one IdP console. Moving the config to runtime removes that constraint.

To make this more loosely coupled, this PR introduces a backend config endpoint that the frontend calls to retrieve runtime config values.

## Validation

Confirmed new API endpoint serves as expected:

image

Also confirmed frontend calls the new API endpoint as expected and IdP deep-links properly redirect.

### DIFF
--- a/api/app.py
+++ b/api/app.py
@@ -322,6 +322,7 @@ def create_app(testing: Optional[bool] = False) -> FastAPI:
         apps,
         audit,
         bugs,
+        config,
         group_requests,
         groups,
         health,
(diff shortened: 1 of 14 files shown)


somethingnew2-0 (changes requested):
We have a few ways to share config between the server and the client. A dynamic API could work, but I would prefer a static client-side config instead of requiring a web request on page load if the config itself is static per deployment

This feels similar to @eguerrant's PR https://github.com/discord/access/pull/406 so we should probably unify on a solution

Jeff-Rowell:
> We have a few ways to share config between the server and the client. A dynamic API could work, but I would prefer a static client-side config instead of requiring a web request on page load if the config itself is static per deployment
> 
> This feels similar to @eguerrant's PR #406 so we should probably unify on a solution

Agreed that a per-request fetch is overkill for config that's static within a deployment. The reason I went with a runtime source rather than baking it at build time is two-fold:
1. the IdP URLs differ per environment but we ship a single image
2. static client-side config would mean committing real IdP admin URLs (it's all placeholders today), which I'd rather avoid

The current endpoint approach sidesteps both by sourcing from the backend on a per-request basis.

PR #498, opened 2026-06-24
Add Google Group Management app-group-lifecycle plugin

barborico:
# Summary
Adds a `google_group_manager` app-group-lifecycle plugin that automatically creates, updates, and deletes Google groups for Access groups and links them to the Access-managed Okta group via Okta group push, plus the core-Access enablers it needs. Group CRUD uses the [Cloud Identity Groups API](https://docs.cloud.google.com/identity/docs/groups).

**Urgency**: 1 week?
**Expected review effort**: HIGH (~1h), since there's quite a lot of code but it's well-organized and -tested, if I do say so myself :)

# Motivation
Access apps need Google groups that mirror their Access groups: when an Access group is created, reconfigured, or deleted, the matching Google group should be created/updated/deleted and linked via Okta group push, which then keeps membership in sync. This plugin delivers that as an opt-in example plugin which can replace manual toil to keep the two group management systems in sync.

# Plugin Capabilities
- **Enabled at app level**: The plugin is configured (and can be separately enabled/disabled) at the scope of individual applications.
- **Lifecycle automation**: creating, updating (display name / description), or deleting an Access group with the plugin enabled creates/updates/deletes the corresponding Google group.
- **Okta group push linking**: creates the push mapping between the Access-managed Okta group and the Google group; Okta then owns membership sync.
- **Idempotent reconciliation**: a single `_reconcile` drives create/update and a periodic `sync_all_groups` sweep; it is safe to re-run and converges to the desired state.
- **Access as source of truth, with graceful backfill from Google**: When Google group config is present in Access, it's pushed to Google. Google groups already linked out-of-band are adopted (missing Access-side config is inferred from live Google/Okta state). If there's no Access-side config or existing link, the plugin does nothing.
- **App-level email-prefix pattern**: an optional regex (e.g. `^gcp-`) is enforced on create/update and surfaced as both client-side and server-side validation.
- **Status visibility**: each group surfaces `sync_status` (synced / pending / error), `sync_error`, `last_synced_at`, `google_group_id`, and `push_mapping_id`.

## Limitations & behavior notes
- **Membership is Okta's responsibility**: the plugin implements no membership hooks; it only manages the Google group and the push mapping. Membership flows entirely through Okta group push.
- **Email is immutable**: the Google group email (Cloud Identity `groupKey`) cannot change after creation, so the `email` config field is enforced immutable on update (rendered read-only in the UI when editing). Display name and description stay mutable.
- **Linking is best-effort and eventually consistent** — if Okta has not yet imported the new Google group, the push mapping cannot be created; the group is marked `pending` and linked on a later reconcile/sweep.
- **The join key is the email**: link discovery relies on Okta's `googleGroupEmail` profile attribute on the imported target app group (Cloud Identity IDs differ from the Directory IDs Okta stores--two different Google APIs for group management).
- **Errors are recorded, not fatal**: blocking Google/Okta errors during reconcile are written to `sync_status=error` (committed so they survive the host's post-error rollback) rather than aborting the operation; a legacy email that predates a later-added pattern is grandfathered (never re-enforced, since the email can't change), and editing a group's other properties is not blocked by an unchanged, non-conforming legacy email.
- **Opt-in**: the published Access image ships the plugin source (`/app/plugins`) but does not install it; a downstream/overlay image opts in by pip-installing it. The README documents this, the Cloud Identity auth model, and the required `GOOGLE_WORKSPACE_*` env vars.

<details>

**Core Access enablers** (`api/`, `src/`)
- `OktaService` gains `create` / `delete` / `list_group_push_mappings` (Okta Group Push Mapping API).
- App-config-aware group-config hooks, plus `set_config_value` / `is_plugin_config_changed` framework helpers. The bulk-sync hook `sync_all_group_membership` is renamed to `sync_all_groups` (with a backward-compatible deprecated alias, and renamed CLI command / make target).
- New `ModifyGroupPluginData` operation: merges a partial `plugin_data` patch (preserving other plugins' entries and plugin-managed status) and fires `group_updated` **only when the plugin's `configuration` changes** — never on status writes, which would otherwise re-trigger reconciliation in a loop. `put_group` routes through it; `post_group` validates group config at creation.
- Immutable group-config fields: a schema flag, generic host-side enforcement on update, and read-only rendering in the config form.
- The `group-config-props` endpoint accepts `?app_id=` so the schema can reflect app-level constraints; the config form validates patterns client-side.
- The Dockerfile copies the plugin source into the image (`/app/plugins`) without installing it, so downstream images can opt in while the source stays single-homed in this repo.
- Misc dev QoL improvements: configurable root `LOG_LEVEL`; a `sync-app-groups` make target that loads `.env`.

**Plugin** (`examples/plugins/app_group_lifecycle_google/`)
- Built up Cloud-Identity-first: package scaffold + docs → config/status schema, validation, email helpers → Cloud Identity Groups API wrappers → Okta push-mapping + link discovery → the idempotent reconcile core → lifecycle hooks (`group_created`/`group_updated`/`group_deleted`) and the `sync_all_groups` sweep.
</details>

<details>
<summary><h1>Web UI Screenshots</h1></summary>

**App Config**
image
image

**Group Config & Status**
Creating a group (with validation):
image
Pending Okta re-pulling groups from Google:
image
Success:
image
With sync errors:
image
Notice grandfathering of bad-pattern email in backfill:
image
Failure to takeover an existing group:
image
</details>

<details>

- The app-group-lifecycle framework, Okta service, and group/app router suites pass, as does the full standalone plugin suite (60 tests). `ruff check` and `ruff format --check` are clean; the frontend `tsc` typecheck and the plugin config form's vitest pass.
- Coverage includes: the Okta push-mapping methods; `ModifyGroupPluginData` (config-change vs status-only, and via `put_group`); create/update validation including immutable-field rejection; and full plugin behavior (schema/validation, Cloud Identity wrappers, push-mapping/discovery, reconcile create/adopt/enforce/grandfather/pending/error branches, and the lifecycle hooks including delete ordering and sweep isolation).
- **Manual smoke testing** the Okta and Google integrations and E2E flow
  - [x] Link a "legacy" group out-of-band with unallowed email -> backfill to Access
  - [x] Create a new group -> group created in Google and pending -> refresh Okta app groups -> reconcile -> group linked in Okta
  - [x] Try to create a group with a bad email -> blocked
  - [x] Modify group name and description with preexisting unallowed email -> change propagated to Google
  - [x] Delete a group -> group and link deleted
</details>

# Guidance for Reviewers
History is clean and deliberately logically grouped — **review commit-by-commit**. It is two halves: the core-Access enablers (`api/`, `src/`) precede the plugin (`examples/`); the tail is two formatting-only commits and a final Dockerfile commit that ships (but doesn't install) the plugin source.

I've reviewed and exercised the code both manually and with the Claude code-review skill. Consider overall functionality--does the behavior match what you'd expect/want this plugin to do?

### DIFF
--- a/api/plugins/app_group_lifecycle.py
+++ b/api/plugins/app_group_lifecycle.py
@@ -467,9 +467,7 @@ def set_config_value(app_or_group: App | AppGroup, config_property_name: str, va
     app_or_group.plugin_data[plugin_id] = asdict(data)
 
 
-def is_plugin_config_changed(
-    old_plugin_data: dict[str, Any], new_plugin_data: dict[str, Any], plugin_id: str
-) -> bool:
+def is_plugin_config_changed(old_plugin_data: dict[str, Any], new_plugin_data: dict[str, Any], plugin_id: str) -> bool:
     """
     Determine whether a particular app group lifecycle plugin's configuration differs
     between two plugin_data payloads. Only the plugin's configuration is compared; status
@@ -566,9 +564,7 @@ def get_app_group_lifecycle_plugin_group_config_properties(
     """
     app_configuration = _get_data_for_plugin(app_plugin_data or {}, plugin_id).configuration
     hook = get_app_group_lifecycle_hook()
-    return _get_hook_call_response(
-        hook.get_plugin_group_config_properties, plugin_id, app_config=app_configuration
-    )
+    return _get_hook_call_response(hook.get_plugin_group_config_properties, plugin_id, app_config=app_configuration)
 
 
 def validate_app_group_lifecycle_plugin_app_config(plugin_data: dict[str, Any], plugin_id: str) -> dict[str, str]:
--- a/api/plugins/app_group_lifecycle.py
+++ b/api/plugins/app_group_lifecycle.py
@@ -51,6 +51,10 @@ class AppGroupLifecyclePluginConfigProperty:
     default_value: Any = None
     required: bool = False
     validation: dict[str, Any] | None = None
+    # When True, the host rejects edits to this field on update (group config only);
+    # the value may be set freely at create time. Enforced in
+    # validate_app_group_lifecycle_plugin_group_config.
+    immutable: bool = False
 
 
 @dataclass
@@ -587,6 +591,7 @@ def validate_app_group_lifecycle_plugin_group_config(
     plugin_data: dict[str, Any],
     plugin_id: str,
     app_plugin_data: dict[str, Any] | None = None,
+    old_plugin_data: dict[str, Any] | None = None,
 ) -> dict[str, str]:
     """
     Validate the group-level configuration data for a particular app group lifecycle plugin.
@@ -596,6 +601,8 @@ def validate_app_group_lifecycle_plugin_group_config(
         plugin_id: The ID of the plugin which should respond.
         app_plugin_data: The owning app's plugin data, so the plugin can validate the group
                          config against app-level settings. Defaults to empty (no app context).
+        old_plugin_data: The group's existing plugin data, supplied on update so the host can
+                         reject edits to immutable config properties. Omitted on create.
 
     Returns:
         A dictionary mapping any invalid fields to error messages.
@@ -603,9 +610,31 @@ def validate_app_group_lifecycle_plugin_group_config(
     configuration = _get_data_for_plugin(plugin_data, plugin_id).configuration
     app_configuration = _get_data_for_plugin(app_plugin_data or {}, plugin_id).configuration
     hook = get_app_group_lifecycle_hook()
-    return _get_hook_call_response(
-        hook.validate_plugin_group_config, plugin_id, config=configuration, app_config=app_configuration
+    errors = dict(
+        _get_hook_call_response(
+            hook.validate_plugin_group_config, plugin_id, config=configuration, app_config=app_configuration
+        )
     )
+
+    # On update, apply immutable-field semantics generically (plugins only declare
+    # `immutable=True`): a changed immutable field is rejected, while a plugin error for an
+    # *unchanged* immutable field is suppressed -- the user can't action it on this update
+    # (the field is locked), and it was acceptable at creation or was adopted/grandfathered
+    # from external state, so it must not block the rest of the update.
+    if old_plugin_data is not None:
+        old_configuration = _get_data_for_plugin(old_plugin_data, plugin_id).configuration
+        properties = _get_hook_call_response(
+            hook.get_plugin_group_config_properties, plugin_id, app_config=app_configuration
+        )
+        for name, prop in properties.items():
+            if not getattr(prop, "immutable", False):
+                continue
+            if old_configuration.get(name) != configuration.get(name):
+                errors.setdefault(name, f"The '{name}' field cannot be changed after creation")
+            else:
+                errors.pop(name, None)
+
+    return errors
 
 
 def get_app_group_lifecycle_plugin_app_status_properties(
--- a/api/plugins/app_group_lifecycle.py
+++ b/api/plugins/app_group_lifecycle.py
@@ -244,16 +244,36 @@ class AppGroupLifecyclePluginSpec:
         """
 
     @hookspec
-    def sync_all_group_membership(self, session: Session, app: App, plugin_id: str | None) -> None:
+    def sync_all_groups(self, session: Session, app: App, plugin_id: str | None) -> None:
         """
-        Bulk sync all group memberships for an app. This is invoked periodically by a CLI command `flask sync-app-group-memberships`.
+        Bulk reconcile all of an app's groups (membership and any external group state).
+        Invoked periodically by the `access sync-app-groups` CLI command.
 
         Args:
             session: The Access database session.
-            app: The app for which to sync all group membership.
+            app: The app for which to sync all groups.
             plugin_id: If provided, only the plugin matching this ID should respond.
                        If None, all plugins may respond.
         """
+
+    @hookspec
+    def sync_all_group_membership(self, session: Session, app: App, plugin_id: str | None) -> None:
+        """
+        Deprecated alias for ``sync_all_groups``, retained so plugins implementing the
+        older, narrower hook name keep being invoked. New plugins should implement
+        ``sync_all_groups`` instead; a plugin should implement only one of the two.
+        """
+
+
+def invoke_sync_all_groups(session: Session, app: App, plugin_id: str | None) -> None:
+    """
+    Invoke the bulk-sync hook for an app, honoring both the current ``sync_all_groups``
+    name and the deprecated ``sync_all_group_membership`` alias. A plugin should implement
+    only one of the two, so this dispatches to each exactly once.
+    """
+    hook = get_app_group_lifecycle_hook()
+    hook.sync_all_groups(session=session, app=app, plugin_id=plugin_id)
+    hook.sync_all_group_membership(session=session, app=app, plugin_id=plugin_id)
 
 
 def get_app_group_lifecycle_hook() -> pluggy.HookRelay:
--- a/api/plugins/app_group_lifecycle.py
+++ b/api/plugins/app_group_lifecycle.py
@@ -107,7 +107,7 @@ class AppGroupLifecyclePluginSpec:
 
     @hookspec
     def get_plugin_group_config_properties(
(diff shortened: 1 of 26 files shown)


barborico:
bugbot run

api/plugins/app_group_lifecycle.py
```diff
@@ -232,17 +248,37 @@ def group_members_removed(
...
+
+def invoke_sync_all_groups(session: Session, app: App, plugin_id: str | None) -> None:
+    """
+    Invoke the bulk-sync hook for an app, honoring both the current ``sync_all_groups``
+    name and the deprecated ``sync_all_group_membership`` alias. A plugin should implement
+    only one of the two, so this dispatches to each exactly once.
+    """
+    hook = get_app_group_lifecycle_hook()
+    hook.sync_all_groups(session=session, app=app, plugin_id=plugin_id)
+    hook.sync_all_group_membership(session=session, app=app, plugin_id=plugin_id)
```
Jeff-Rowell: This dispatches both the new and deprecated hook names. The docstring says implement only one, but nothing enforces it, so a plugin implementing both (e.g. mid-migration) gets every group reconciled twice. Suggestion below guards against it by raising if any plugin implements both:

```suggestion
    hook = get_app_group_lifecycle_hook()
    current_impls = {impl.plugin for impl in hook.sync_all_groups.get_hookimpls()}
    deprecated_impls = {impl.plugin for impl in hook.sync_all_group_membership.get_hookimpls()}
    both = current_impls & deprecated_impls
    if both:
        names = ", ".join(getattr(p, "__name__", repr(p)) for p in both)
        raise ValueError(
            f"Plugin(s) {names} implement both sync_all_groups and the deprecated "
            "sync_all_group_membership; implement exactly one."
        )
    hook.sync_all_groups(session=session, app=app, plugin_id=plugin_id)
    hook.sync_all_group_membership(session=session, app=app, plugin_id=plugin_id)
```

api/plugins/app_group_lifecycle.py
```diff
@@ -507,20 +583,54 @@ def validate_app_group_lifecycle_plugin_app_config(plugin_data: dict[str, Any],
...
+    # from external state, so it must not block the rest of the update.
+    if old_plugin_data is not None:
+        old_configuration = _get_data_for_plugin(old_plugin_data, plugin_id).configuration
+        properties = _get_hook_call_response(
+            hook.get_plugin_group_config_properties, plugin_id, app_config=app_configuration
+        )
+        for name, prop in properties.items():
+            if not getattr(prop, "immutable", False):
+                continue
+            if old_configuration.get(name) != configuration.get(name):
```
Jeff-Rowell, line 628: This compares the raw incoming patch against the old config, but the patch can be partial (that's the point of `ModifyGroupPluginData`'s merge). A client that PUTs `{configuration: {display_name: ...}}` and omits the immutable `email` trips this: `email` goes from set to `None`, reads as a change, and 400s even though the user never touched it. The UI hides it by resubmitting the read-only field, but partial API updates break.

Fix: only treat a field as edited when it's actually present in the patch. An omission isn't a change; an explicit value (even `null`) still is. The grandfathering `else` branch is unaffected, and a present-and-different value is still rejected.

```suggestion
            if name in configuration and old_configuration.get(name) != configuration.get(name):
```

examples/plugins/app_group_lifecycle_google/plugin.py
```diff
@@ -0,0 +1,728 @@
...
+            # Description is handled below for both directions; only push it here when Access
+            # has a (differing) non-empty description.
+            access_desc = group.description or ""
+            patch_description = access_desc if access_desc and (live.get("description") or "") != access_desc else None
+            self._patch_google_group(google_group_id, display_name=patch_display_name, description=patch_description)
+
+        # Description sync (both directions): clobber if Access has one, else backfill.
+        access_desc = group.description or ""
+        google_desc = live.get("description", "") or ""
+        if not access_desc and google_desc:
```
Jeff-Rowell, line 656: I think this makes it impossible to clear a description. If a user empties the Access description and Google still has one, `access_desc` is `""` so the enforce branch won't push the clear, and then this backfills Google's value right back (and pushes it to Okta)

api/routers/groups.py
```diff
@@ -366,16 +399,10 @@ def put_group(
...
     if new_plugin_data is not None:
-        group.plugin_data = new_plugin_data
-        if old_plugin_data and group.plugin_data != old_plugin_data:
-            for key in old_plugin_data:
-                if key not in group.plugin_data:
-                    group.plugin_data[key] = old_plugin_data[key]
-            if type(group) is AppGroup:
-                merge_app_lifecycle_plugin_data(group, old_plugin_data)
-    db.commit()
+        ModifyGroupPluginData(group=group, plugin_data=new_plugin_data).execute()
```
Jeff-Rowell, line 403: `ModifyGroupDetails` fires `group_updated` on a name/description change, and this fires it again on a config change. A PUT that changes both name and plugin config in one request runs a full reconcile twice (Google + Okta round-trips)

Proposed fix: fire `group_updated` exactly once in `put_group` after both operations apply, and suppress the per-operation fires.

1. `ModifyGroupDetails` — add an opt-out (default preserves behavior for its other callers):

```python
def __init__(self, *, group, name=None, description=None, current_user_id=None,
             validate_app_group_prefix=True, fire_lifecycle_hook=True):
    ...
    self.fire_lifecycle_hook = fire_lifecycle_hook

# in execute():
if self.fire_lifecycle_hook and (old_name != self.group.name or old_description != self.group.description):
    ...  # existing group_updated fire
```

2. `ModifyGroupPluginData` — same `fire_lifecycle_hook=True` opt-out, so `put_group` can suppress its internal fire (the merge still runs; only the hook call is gated).

3. `put_group` — capture old values before `ModifyGroupDetails`, suppress both, fire once at the end:

```python
old_name = group.name
old_description = group.description or ""
plugin_id = get_app_group_lifecycle_plugin_to_invoke(group)
...
ModifyGroupDetails(..., fire_lifecycle_hook=False).execute()
...
old_plugin_data_for_audit = copy.deepcopy(group.plugin_data) if group.plugin_data else {}
if new_plugin_data is not None:
    ModifyGroupPluginData(group=group, plugin_data=new_plugin_data, fire_lifecycle_hook=False).execute()
else:
    db.commit()

config_changed = (
    new_plugin_data is not None
    and plugin_id is not None
    and is_plugin_config_changed(old_plugin_data_for_audit, group.plugin_data or {}, plugin_id)
)
if plugin_id is not None and (name_or_desc_changed or config_changed):
    try:
        get_app_group_lifecycle_hook().group_updated(
            session=db, group=group, old_name=old_name,
            old_description=old_description, plugin_id=plugin_id,
        )
        db.commit()
    except Exception:
        logging.getLogger("api").exception(
            f"Failed to invoke group_updated hook for group {group.id} with plugin '{plugin_id}'"
        )
        db.rollback()
```

api/plugins/app_group_lifecycle.py
Jeff-Rowell, line 280: This dispatches both the new and deprecated hook names. The docstring says implement only one, but nothing enforces it, so a plugin implementing both (e.g. mid-migration) gets every group reconciled twice. Suggestion below guards against it by raising if any plugin implements both:

```suggestion
    hook = get_app_group_lifecycle_hook()
    current_impls = {impl.plugin for impl in hook.sync_all_groups.get_hookimpls()}
    deprecated_impls = {impl.plugin for impl in hook.sync_all_group_membership.get_hookimpls()}
    both = current_impls & deprecated_impls
    if both:
        names = ", ".join(getattr(p, "__name__", repr(p)) for p in both)
        raise ValueError(
            f"Plugin(s) {names} implement both sync_all_groups and the deprecated "
            "sync_all_group_membership; implement exactly one."
        )
    hook.sync_all_groups(session=session, app=app, plugin_id=plugin_id)
    hook.sync_all_group_membership(session=session, app=app, plugin_id=plugin_id)
```

Jeff-Rowell (changes requested):

somethingnew2-0:
I've rebased the PR after quite a few changes: #480 #481 #507 #511

The rebased version of the PR is in draft at https://github.com/discord/access/pull/532 so as not to completely overwrite this one. PTAL!

barborico:
Superseded by rebased branch: https://github.com/discord/access/pull/532

PR #500, opened 2026-06-25
perf!: paginate apps to reduce large ORM graphs causing OOMKilled

Jeff-Rowell:
## Summary

If an app within `/api/apps` has an excessive number of groups or excessive number of users in groups (e.g., 1,000 groups each with 5 members, 5 groups with 1,000+ members, etc), the ORM graph loaded in memory is unbounded leading to OOMKilled errors. 

This issue is unbounded and scales with the number of groups, or inlined members nested within those groups.

This PR adds two things:
1. a hard 10 groups per-page limit for apps
2. pagination for inline members of groups within apps from `/api/apps/<App-Name>`

This PR also *changes* the behavior for search. Before, search ran client-side on keystroke since all groups/users were rendered and returned at once. After this PR lands, search runs server-side on submit.

## Edge Cases

There are corner cases where even the 10 groups within a single page returned from `/api/apps/<App-Name>` can still cause memory bloat. A fix is to paginate the inline members within the groups within `/api/apps/<App-Name>` to reduce the memory footprint.

The members are the main memory amplification factor so I've added inline member pagination within the group accordions.

## Testing and Validation

App group pagination (infinite scroll), inline member pagination and search:

https://github.com/user-attachments/assets/79ec11ab-b772-4826-9ce0-f4ec28340047

Tests: 552 passed

### DIFF
--- a/api/pagination.py
+++ b/api/pagination.py
@@ -16,7 +16,7 @@ from typing import Any, TypeVar
 
 from fastapi import Query
 from fastapi_pagination import Page as _BasePage
-from fastapi_pagination.customization import CustomizedPage, UseParams
+from fastapi_pagination.customization import CustomizedPage, UseParams, UseParamsFields
 from fastapi_pagination.default import Params as _DefaultParams
 from pydantic import TypeAdapter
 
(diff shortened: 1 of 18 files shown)


somethingnew2-0:
We don't do this elsewhere in Access, but instead of requiring the user to click the pagination buttons for app groups, could we add infinite scrolling instead to load the next page?

Jeff-Rowell:
> We don't do this elsewhere in Access, but instead of requiring the user to click the pagination buttons for app groups, could we add infinite scrolling instead to load the next page?

Yeah can definitely make that change. What are your thoughts on the inlined member pagination?

somethingnew2-0:
> What are your thoughts on the inlined member pagination?

Probably a good optimization for large groups and we could probably make the default page size for those be fairly large (500 or 1000+ members?) to balance the number of clicks folks need to do and the page load time.

api/pagination.py
```diff
@@ -31,13 +36,26 @@ class PageParams(_DefaultParams):
     size: int = Query(DEFAULT_SIZE, ge=1, le=MAX_SIZE, description="Items per page")
 
 
+class AppGroupsPageParams(_DefaultParams):
+    """Params for `GET /api/apps/{id}/groups`: `size` defaults to and is capped
+    at APP_GROUPS_SIZE so a single response can't load an unbounded number of
+    groups' memberships."""
+
+    page: int = Query(1, ge=1, description="Page number")
+    size: int = Query(APP_GROUPS_SIZE, ge=1, le=APP_GROUPS_SIZE, description="Items per page")
```
somethingnew2-0: Is there another way of setting APP_GROUPS_SIZE for the `size` parameter in the route response code instead of creating one-off custom AppGroupsPageParams class?

somethingnew2-0 (dismissed):

somethingnew2-0 (approved):

PR #502, opened 2026-06-27
Plugin group config in group requests

barborico:
# Summary
Lets users supply (and resolvers edit) app-group-lifecycle plugin group config when requesting an app group, validates it at submit and approval, and applies it to the group created on approval so the plugin's `group_created` hook runs against a configured group. The detail view surfaces the config under Requested/Approved details.

**Urgency**: 5 days

# Motivation
App-group-lifecycle plugins can require per-group config (e.g. a Google Group plugin needs an email prefix + display name). Today that config is only collectable in the direct group-create/edit flow. When a user *requests* a group for such an app, there is nowhere to supply it, so approval creates the group unconfigured. This adds a generic "plugin-config-in-requests" capability.

# UI
image
image

<details>

- **Model + migration**: `requested_plugin_data` / `resolved_plugin_data` JSON columns on `GroupRequest` (JSONB on Postgres), mirroring the existing `requested_*`/`resolved_*` pairs. Config is stored in the group `plugin_data` shape `{plugin_id: {configuration: {...}}}`.
- **Create**: `CreateGroupRequest` persists `requested_plugin_data`; the POST endpoint validates it against the app's plugin (400 on invalid/missing-required).
- **Approve**: `ApproveGroupRequest` resolves `resolved_plugin_data or requested_plugin_data`, re-validates, and sets it on the new `AppGroup` *before* `CreateGroup` runs, so `group_created` sees the config. Invalid config → 400 via the PUT handler.
- **Resolve (PUT)**: accepts a resolver's `resolved_plugin_data` edits.
- **Frontend**: the create form renders the existing plugin config form for app-group requests whose app has a plugin (fetching `AppDetail` via `useAppById`, since search summaries omit the plugin id) and submits `requested_plugin_data`; the detail view shows the config under Requested and Approved details (resolver-changed values as ~~old~~ → new); the resolver can edit before approving; the config header is hidden when a plugin declares no group fields. DOM-free helpers with unit tests. Client `plugin_data` fields hand-added to `apiSchemas.ts` (the resolved codegen renames generics across the whole client; drift-check workflow permits hand-editing). Also fixes a stale `jest-dom` import that blocked vitest.

</details>

<details>

- Backend (`uv run pytest`): feature file + regressions **119 passed** (`test_group_request`, `test_group_request_plugin_config`, `test_app_group_lifecycle_plugin`). `uv run ruff check`/`ruff format` clean; `uv run ty check` clean. Migration verified up + down; single alembic head.
- Frontend: `tsc --noEmit` clean; **14** vitest unit tests pass; prettier clean.
- [x] Manual UI check: resolver form pre-fills the requester's submitted config.

</details>

# Guidance for Reviewers
Rebased onto current `main` (which had moved through the async DB/Okta refactor + Okta SDK v3); the feature and its tests were ported to the async APIs, and the migration re-chained to the current head. Review commit-by-commit — five commits by layer: (1) model + migration, (2) schemas, (3) async operations + router + backend tests, (4) TS client, (5) frontend. Most worth a look: `ApproveGroupRequest` setting `new_group.plugin_data` before `CreateGroup(...).execute()` so the async `group_created` hook sees it.

### DIFF
--- a/api/operations/approve_group_request.py
+++ b/api/operations/approve_group_request.py
@@ -10,6 +10,7 @@ from api.exceptions import ConflictError
 from api.extensions import db
 from api.models import (
     AccessRequestStatus,
+    App,
     AppGroup,
     GroupRequest,
     OktaGroup,
@@ -118,6 +119,11 @@ class ApproveGroupRequest:
             if group_request.resolved_group_tags
             else group_request.requested_group_tags
         )
+        resolved_plugin_data = (
+            group_request.resolved_plugin_data
+            if group_request.resolved_plugin_data
+            else group_request.requested_plugin_data
+        )
 
         # authorization
         access_owner_ids = {u.id for u in await get_access_owners()}
@@ -206,6 +212,24 @@ class ApproveGroupRequest:
                 description=resolved_description,
                 app_id=resolved_app_id,
             )
+            # Carry the request's plugin config onto the group so the
+            # group_created hook (fired inside CreateGroup) sees it. Re-validate
+            # defensively: the app or its config may have changed since filing.
+            if resolved_plugin_data:
+                resolved_app = await db.session.get(App, resolved_app_id)
+                plugin_id = resolved_app.app_group_lifecycle_plugin if resolved_app is not None else None
+                if plugin_id is not None:
+                    from api.plugins.app_group_lifecycle import (
+                        validate_app_group_lifecycle_plugin_group_config,
+                    )
+
+                    plugin_errors = validate_app_group_lifecycle_plugin_group_config(
+                        resolved_plugin_data,
+                        plugin_id,
+                    )
+                    if plugin_errors:
+                        raise ValueError(f"plugin_data: {plugin_errors}")
+                    new_group.plugin_data = resolved_plugin_data
         else:
             new_group = OktaGroup(
                 name=resolved_name,
--- a/api/operations/approve_group_request.py
+++ b/api/operations/approve_group_request.py
@@ -1,11 +1,11 @@
+import copy
+import logging
 from typing import Optional
 
-import logging
-
-from api.context import get_request_context
 from sqlalchemy import func, select
 from sqlalchemy.orm import joinedload, selectin_polymorphic, selectinload, with_polymorphic
 
+from api.context import get_request_context
 from api.exceptions import ConflictError
 from api.extensions import db
 from api.models import (
@@ -23,12 +23,14 @@ from api.models import (
 from api.models.access_request import get_all_possible_request_approvers
 from api.models.app_group import get_access_owners
 from api.models.tag import coalesce_ended_at
+from api.operations._fan_out import defer_notification
 from api.operations.constraints.check_for_self_add import CheckForSelfAdd
 from api.operations.create_group import CreateGroup
 from api.operations.modify_group_users import ModifyGroupUsers
-from api.operations._fan_out import defer_notification
 from api.plugins import NotificationHook
 from api.schemas import AuditLogSchema, EventType
+
+logger = logging.getLogger(__name__)
 
 
 class ApproveGroupRequest:
@@ -229,7 +231,17 @@ class ApproveGroupRequest:
                     )
                     if plugin_errors:
                         raise ValueError(f"plugin_data: {plugin_errors}")
-                    new_group.plugin_data = resolved_plugin_data
+                    # Deep-copy so the created group and the persisted request
+                    # don't share nested plugin_data objects: the group_created
+                    # hook may mutate the group's copy (e.g. writing status),
+                    # which must not leak back into the immutable request record.
+                    new_group.plugin_data = copy.deepcopy(resolved_plugin_data)
+                else:
+                    logger.warning(
+                        f"Group request {group_request.id} carried plugin_data, "
+                        f"but app {resolved_app_id} has no app_group_lifecycle_plugin; "
+                        "dropping the supplied config."
+                    )
         else:
             new_group = OktaGroup(
                 name=resolved_name,
--- a/api/operations/approve_group_request.py
+++ b/api/operations/approve_group_request.py
@@ -1,6 +1,6 @@
 import copy
 import logging
-from typing import Optional
+from typing import Any, Optional
 
 from sqlalchemy import func, select
 from sqlalchemy.orm import joinedload, selectin_polymorphic, selectinload, with_polymorphic
@@ -162,6 +162,7 @@ class ApproveGroupRequest:
             if not is_admin:
                 return group_request
 
+        # Validation
         if resolved_type != "app_group" and resolved_name.startswith(AppGroup.APP_GROUP_NAME_PREFIX):
             return group_request
 
@@ -182,6 +183,38 @@ class ApproveGroupRequest:
         ).first()
         if existing_group is not None:
             return group_request
+
+        # Resolve the plugin config the new group will carry, and re-validate it here
+        # with the rest of the validation: the app or its plugin config may have
+        # changed since the request was filed, so this can legitimately fail. Raising
+        # after the group_request_approve audit event below would leave an "approved"
+        # audit entry for an approval that never completed — re-emitted on every retry.
+        new_group_plugin_data: dict[str, Any] = {}
+        if resolved_type == "app_group" and resolved_plugin_data:
+            resolved_app = await db.session.get(App, resolved_app_id)
+            plugin_id = resolved_app.app_group_lifecycle_plugin if resolved_app is not None else None
+            if plugin_id is None:
+                logger.warning(
+                    f"Group request {group_request.id} carried plugin_data, "
+                    f"but app {resolved_app_id} has no app_group_lifecycle_plugin; "
+                    "dropping the supplied config."
+                )
+            else:
+                from api.plugins.app_group_lifecycle import (
+                    validate_app_group_lifecycle_plugin_group_config,
+                )
+
+                plugin_errors = validate_app_group_lifecycle_plugin_group_config(
+                    resolved_plugin_data,
+                    plugin_id,
+                )
+                if plugin_errors:
+                    raise ValueError(f"plugin_data: {plugin_errors}")
+                # Deep-copy so the created group and the persisted request don't share
+                # nested plugin_data objects: the group_created hook may mutate the
+                # group's copy (e.g. writing status), which must not leak back into the
+                # immutable request record.
+                new_group_plugin_data = copy.deepcopy(resolved_plugin_data)
 
         await db.session.commit()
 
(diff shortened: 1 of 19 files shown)


api/routers/group_requests.py
```diff
@@ -182,6 +182,22 @@ async def post_group_request(
...
+
+            try:
+                plugin_errors = validate_app_group_lifecycle_plugin_group_config(
+                    body.requested_plugin_data,
+                    app.app_group_lifecycle_plugin,
+                )
+            except ValueError as e:
+                raise HTTPException(400, f"plugin_data: {e}") from e
+            if plugin_errors:
+                raise HTTPException(400, f"plugin_data: {plugin_errors}")
```
Jeff-Rowell: `validate_app_group_lifecycle_plugin_group_config` raises `AppGroupLifecyclePluginFilteringError` (a subclass of `Exception`, not `ValueError`) when the app's configured plugin id isn't registered, so this `except ValueError` misses it and the request 500s instead of returning a clean 400.

```suggestion
            from api.plugins.app_group_lifecycle import (
                AppGroupLifecyclePluginFilteringError,
                validate_app_group_lifecycle_plugin_group_config,
            )

            try:
                plugin_errors = validate_app_group_lifecycle_plugin_group_config(
                    body.requested_plugin_data,
                    app.app_group_lifecycle_plugin,
                )
            except (ValueError, AppGroupLifecyclePluginFilteringError) as e:
                raise HTTPException(400, f"plugin_data: {e}") from e
            if plugin_errors:
                raise HTTPException(400, f"plugin_data: {plugin_errors}")
```

api/routers/group_requests.py
```diff
@@ -210,16 +226,23 @@ async def post_group_request(
...
+            requested_group_name=body.requested_group_name,
+            requested_group_description=body.requested_group_description or "",
+            requested_group_type=body.requested_group_type,
+            requested_app_id=requested_app_id,
+            requested_group_tags=body.requested_group_tags,
+            requested_ownership_ending_at=body.requested_ownership_ending_at,
+            request_reason=body.request_reason or "",
+            requested_plugin_data=body.requested_plugin_data if isinstance(body, _AppGroupRequestBody) else {},
+        ).execute()
+    except ValueError as e:
```
Jeff-Rowell, line 247: The approval-time re-validation raises the same `AppGroupLifecyclePluginFilteringError`, which also slips past this `except ValueError` (and the PUT catch at line 325) into a 500. Broaden these to `(ValueError, AppGroupLifecyclePluginFilteringError)` too.

api/operations/approve_group_request.py
```diff
@@ -205,6 +213,34 @@ async def execute(self) -> Optional[GroupRequest]:
...
+                    from api.plugins.app_group_lifecycle import (
+                        validate_app_group_lifecycle_plugin_group_config,
+                    )
+
+                    plugin_errors = validate_app_group_lifecycle_plugin_group_config(
+                        resolved_plugin_data,
+                        plugin_id,
+                    )
+                    if plugin_errors:
+                        raise ValueError(f"plugin_data: {plugin_errors}")
```
Jeff-Rowell: This can `raise` after the `group_request_approve` audit event was already emitted (line 190), so a config that's gone invalid since filing leaves an "approved" audit entry for an approval that never completed (re-emitted on each retry). Consider re-validating before the audit log.

src/pages/group_requests/Create.tsx
```diff
@@ -195,13 +208,16 @@ function CreateRequestContainer(props: CreateRequestContainerProps) {
...
+    const requestedPluginData = extractRequestedPluginData(formData as any);
+
     const body = {
       requested_group_name: groupName,
       requested_group_description: formData.description ?? '',
       requested_group_type: effectiveType,
       requested_app_id: appId,
       request_reason: formData.reason ?? '',
       requested_group_tags: selectedTags.map((t) => t.id),
+      ...(effectiveType === 'app_group' && requestPluginId ? {requested_plugin_data: requestedPluginData} : {}),
```
Jeff-Rowell, line 220: `requestedPluginData` is the whole RHF `plugin_data` object. If the requester fills config for one app then switches to another (RHF keeps the old fields), stale config for the previous plugin is submitted and later deep-copied onto the group. Scope it to the current plugin:

```suggestion
      ...(effectiveType === 'app_group' && requestPluginId
        ? {requested_plugin_data: requestedPluginData[requestPluginId] ? {[requestPluginId]: requestedPluginData[requestPluginId]} : {}}
        : {}),
```

api/routers/group_requests.py
```diff
@@ -286,6 +309,8 @@ async def put_group_request(
         gr.resolved_group_tags = body.resolved_group_tags
     if body.resolved_ownership_ending_at is not None:
         gr.resolved_ownership_ending_at = body.resolved_ownership_ending_at
+    if body.resolved_plugin_data is not None:
+        gr.resolved_plugin_data = body.resolved_plugin_data
```
Jeff-Rowell, line 318: `resolved_plugin_data` is persisted even on rejection and even when the resolved app has no lifecycle plugin (approval drops it with a warning), so a stored request can show config that was never applied. Consider only persisting it on approval when the app actually has a plugin.

src/components/AppGroupLifecyclePluginConfigurationForm.test.tsx
somethingnew2-0, line 1: We don't have any other frontend tests and nothing runs them in CI. Do we want to start including these or should they be removed?

api/routers/group_requests.py
somethingnew2-0, line 247: Do operations normally throw errors or is this the first one?

barborico: It looks like there's precedent for this pattern elsewhere:
https://github.com/discord/access/blob/381da8310c8b11ad30228fa6c921ec7243651df7/api/routers/groups.py#L354-L364
https://github.com/discord/access/blob/381da8310c8b11ad30228fa6c921ec7243651df7/api/routers/groups.py#L377-L382
https://github.com/discord/access/blob/381da8310c8b11ad30228fa6c921ec7243651df7/api/routers/groups.py#L204-L207
https://github.com/discord/access/blob/381da8310c8b11ad30228fa6c921ec7243651df7/api/routers/group_requests.py#L319-L326

src/components/AppGroupLifecyclePluginConfigurationForm.test.tsx
barborico: IMO tests are especially cheap and useful in the vibe coding era. I'll stack a PR to run them in CI.

api/routers/group_requests.py
somethingnew2-0: Ah okay

src/components/AppGroupLifecyclePluginConfigurationForm.test.tsx
barborico: Added: https://github.com/discord/access/pull/564

Jeff-Rowell (dismissed):

api/routers/group_requests.py
barborico: It's fine for there a group/group request to have inactive plugin config; this allows switching back and forth between plugins (unlikely scenario) without losing all prior config.

src/pages/group_requests/Create.tsx
barborico: This is as intended.

api/routers/group_requests.py
barborico: Fixed, and then reworked; the final shape is in b5e4c9f rather than the original fix.

828e805 first broadened this catch to `(ValueError, AppGroupLifecyclePluginFilteringError)`, and did the same on the two other catches in this file, on the direct group-edit path (`put_group` in `api/routers/groups.py`), and on the app-config path (`put_app` in `api/routers/apps.py`, 4ae14ea). Each of those had the identical gap, and each was confirmed to return a 500 beforehand.

Mapping the filtering error to a blanket 400 turned out to be wrong, though, so b5e4c9f replaces all of it with one shared helper lifted into `api.plugins.app_group_lifecycle`. This site is now just:

```python
validate_group_plugin_config_or_raise(
    body.requested_plugin_data,
    app.plugin_data or {},
    app.app_group_lifecycle_plugin,
)
```

and the error is split by cause instead of flattened:

- no registered plugin claims the id (bad input, or app config left behind when a plugin was dropped from the deployment) -> **400** `The plugin {id} is not known`
- the id is registered but its hook did not answer with exactly one response -> **500**, an explicit one naming the plugin rather than an unhandled stack trace

One behaviour fix fell out of routing this site through the shared helper: it had been calling the validator with two arguments, so a plugin validating group config against app-level settings saw an empty app config here while seeing a populated one on the group-create path. It now passes the app plugin_data.

Worth noting for anyone adding a plugin-config call site: `except ValueError` is not enough on its own, since `AppGroupLifecyclePluginFilteringError` subclasses `Exception` directly. Use the shared helper, or `raise_http_for_plugin_filtering_error` if you are catching the error out of an operation.

api/routers/group_requests.py
barborico: 828e805 broadened this catch and the PUT one at the `ApproveGroupRequest` call to `(ValueError, AppGroupLifecyclePluginFilteringError)`. b5e4c9f then refined both, because those two errors do not deserve the same status: `ValueError` stays a 400, while the filtering error goes through the shared `raise_http_for_plugin_filtering_error`, which answers 400 when no registered plugin claims the id and 500 when one does but its hook misbehaves.

Both of these sites catch the error propagating out of an operation rather than from a validation call, which is why the status mapping is a standalone function and not folded into the validation helper.

Tests: `test_put_approve_returns_400_when_app_plugin_is_unregistered` for the PUT catch, and the parametrized `test_post_auto_approve_filtering_error_status` for this one, covering the 400 and 500 branches through the auto-approve path. (The earlier single-status version of that test asserted a 400 while raising the error for a plugin id that *is* registered, which is precisely the 500 case, so it was replaced.)

api/operations/approve_group_request.py
barborico: Good catch; fixed in 7bbc98b. Resolving and validating the plugin config now happens with the rest of the validation, before the commit and the audit event, and the validated deep copy is stashed in a local that the `AppGroup` branch consumes. The "app has no lifecycle plugin, dropping config" warning moved up with it, and the success path is unchanged.

`test_approve_invalid_plugin_config_emits_no_approve_audit_event` asserts that no `GROUP_REQUEST_APPROVE` record is emitted when approval-time validation fails. Without the reordering it fails, logging exactly the phantom event you described.

(SHA updated from the pre-rebase 7ec8c98; the reordering itself is unchanged by the rebase and by the later validation refactor in b5e4c9f.)

Jeff-Rowell (approved):

PR #507, opened 2026-07-07
build: adopt uv and ty in place of pip/tox and mypy

somethingnew2-0:
## What

Switch the Python toolchain to Astral's tools: **uv** for dependency management (with a committed lockfile) and **ty** for type checking, replacing pip/tox and mypy.

- **Packaging** is consolidated into `pyproject.toml` — runtime deps plus `dev`/`test` dependency groups — and a committed `uv.lock` makes installs reproducible. `setup.py`, `requirements*.txt`, `constraints.txt`, and `tox.ini` are removed; `.python-version` pins the interpreter to 3.13.
- **Dev / CI / Docker** now go through `uv sync` / `uv run`. The production image installs from the frozen lockfile into `/app/.venv` (both `gunicorn` and the `access` console script run from there). CI uses `astral-sh/setup-uv`; `uv sync --locked` doubles as a lockfile-drift check.
- **Type checking** moves from mypy to `ty`, configured in `pyproject.toml`.

## Why it matters / what to know

- **ty is a preview tool and its diagnostics differ from mypy's**, so the old `.mypy.ini` per-module suppressions don't translate 1:1. The `[tool.ty]` config was derived by running ty against the tree and re-deriving the *minimum* suppressions that mirror `.mypy.ini`'s intent: the FastAPI routers, the SQLAlchemy-heavy operations layer, the Okta service wrappers, MCP tools, auth, and tests are relaxed; the core (`api/models`, `api/schemas`, `api/database`, …) stays strict. Most findings were the Okta SDK typing every field `Optional` and the `db.session.get(...).attr` pattern — things mypy tolerated via `ignore_missing_imports`, which ty surfaces because it resolves the real types.
- The example plugins under `examples/plugins/` intentionally keep their own `setup.py`/`requirements.txt` (they demonstrate the standalone pluggy packaging pattern); only the commands that install them switch to `uv pip install`.
- **`ruff` is left at its existing pin (`0.8.0`)** rather than bumped, to avoid formatting/lint churn unrelated to this migration. Worth a separate follow-up. `ty` is pinned to the current latest (`0.0.56`).

Verified locally: `uv sync`, the full test suite on both sqlite and postgres, `ruff`, a clean `ty check`, `alembic upgrade`/`check`, and a `docker build` whose image runs the `access` CLI and imports the app.


.github/workflows/ci.yml
```diff
@@ -33,63 +33,68 @@ jobs:
       - uses: actions/checkout@v7
         with:
           persist-credentials: false
-      - name: Set up Python ${{ matrix.python-version }}
-        uses: actions/setup-python@v6
+      - name: Set up uv
+        uses: astral-sh/setup-uv@v6
```
Jeff-Rowell, line 37: I saw we're adding dependency cooldowns. We should also probably pin the uv version here to protect against uv itself being compromised.

somethingnew2-0: Good call — pinned uv to `0.11.10` (matching the Dockerfile's `ghcr.io/astral-sh/uv` image) via the `version:` input on `astral-sh/setup-uv` across all four workflows, so a compromised newer uv release can't run in CI.

Jeff-Rowell (approved):

PR #509, opened 2026-07-07
feat: upgrade to Okta SDK v3 and pool the client on the server loop

somethingnew2-0:
Implements POST_MIGRATION_TODO **11a** (Okta SDK v3 upgrade) and **11b** (shared pooled Okta client). Both are contained behind the `OktaService` facade, so no call site outside the service changed — the syncer, operations, integrity checks, and CLI all keep going through the same facade methods.

## 11a — `okta` 2.9.8 → 3.4.4

v3 is the openapi-generator rewrite the TODO anticipated (`okta.api.*`, Pydantic models). Two realities cut the blast radius down from what that note assumed, and are the key things to know when reviewing:

- **The tuple contract survived.** Despite the rewrite, the generated methods still return `(data, resp, error)` / `(resp, error)` and *return* API errors rather than raising them. So the facade's tuple-based retry loop stayed intact — the main change is `response.get_status()` → `response.status_code` and reading the rate-limit reset header case-insensitively.
- **Group ownership is now native.** v3 ships `assign_group_owner` / `delete_group_owner` / `list_group_owners`, so the hand-rolled request-executor plumbing the TODO called out is gone.

Other v3 deltas absorbed inside the facade: method renames (`add_group` / `replace_group` / `assign_user_to_group` / `unassign_user_from_group`); cursor-based pagination via `PaginationHelper` (the old `resp.next()` walker is gone, and the SDK's bundled `paginate_all` helper is unusable in this build — it calls `*_with_http_info` variants that don't exist — so the facade drives the `after` cursor itself); `Group.profile` as an anyOf union reached through `actual_instance`; and already-parsed `datetime` fields (no more `dateutil` parsing). `get_user_schema` recovers the schema `_links` href from the raw response body because the v3 `UserType` model drops `_links`.

**Retry posture:** rate-limiting and timeouts are both the SDK's job. It retries 429s (honoring `X-Rate-Limit-Reset`) up to `rateLimit.maxRetries` (default 2), and `requestTimeout` bounds each request — an aiohttp per-request socket timeout plus a cumulative deadline across those retries. The facade keeps no retry loop or `wait_for` of its own; a thin `_call` wrapper just maps the SDK's "gave up" outcomes — a request timeout, or a 429 that outlived its retries — to `OktaTimeout`, which the membership/ownership fan-out catches and swallows (the syncer reconciles later). 5xx errors are intentionally not retried — left to Okta.

## 11b — pooled client

`OktaService` now holds one client (and its pooled aiohttp connector) for the server loop, created in the FastAPI lifespan and reused by `_okta_client()` when `asyncio.get_running_loop()` matches the loop it was built on. CLI invocations and per-test loops still build a fresh client per call — no pooled client is started when Okta is unconfigured, and the running-loop check prevents cross-loop session reuse. The TODO's "measure before bothering" caveat still stands; this just puts the pooling in place to measure.

## Verification note

The test suite mocks Okta at the facade boundary, so the v3 wire behavior it can't exercise — 429/rate-limit handling, multi-page pagination, `get_user_schema`'s `_links` recovery, and the group-owner endpoints (off by default behind `OKTA_USE_GROUP_OWNERS_API`) — should be validated against a live/staging org before this rolls out, as the TODO flagged.


api/services/okta_service.py
```diff
@@ -1,41 +1,106 @@
...
+    Surfaced when the SDK gives up on a request — either its ``requestTimeout``
+    is exceeded or a 429 outlives its rate-limit retries. Callers that can
+    tolerate a slow Okta call (the membership/ownership fan-out) catch this and
+    continue; the syncer reconciles the drift later.
+    """
 
     pass
 
 
+def _is_timeout_or_rate_limited(error: Any) -> bool:
```
matthew-bass, line 49: nit: this looks like it combines 429 rate-limit exhaustion and real timeouts into one `OktaTimeout`, which might be harder to debug / root cause.

matthew-bass (dismissed):
just one nit on logging, but tests / code lgtm

api/services/okta_service.py
somethingnew2-0: The old code would throw an OktaTimeout on the following error status codes, not just 429s. So the behavior is essentially the same as before

```
RETRIABLE_STATUS_CODES = [429, 500, 502, 503, 504]
```

matthew-bass (approved):

PR #510, opened 2026-07-07
fix: return MCP tool DB connections to the pool from their own task

somethingnew2-0:
## What & why

Production is emitting SQLAlchemy's *"The garbage collector is trying to clean up non-checked-in connection … which will be terminated"* error against `/mcp`. The leak was introduced by the async SQLAlchemy migration (#480).

MCP tool handlers run inside FastMCP's `StreamableHTTPSessionManager` **server task**, which is decoupled from the `/mcp` HTTP request task. Read-only tools use the request-scoped `db.session` directly and never commit, so the checked-out connection is only ever returned by `RequestIdMiddleware`'s `db.remove()`. That teardown runs in the *request* task — but the connection belongs to the *server* task, and closing an async connection from a different task than the one that opened it does not check it back in. It lingers until the garbage collector reaps it and logs the warning.

## The fix

Each tool call now gets its own DB session scope, bound and torn down (`db.remove()`) **inside the tool's own task** via a new `tool_session_scope` context manager. `requires_scope` — the wrapper every tool already passes through in the server task — enters it, so both read tools (`db.session`) and write tools (`mcp_db_session`) are covered with no per-tool boilerplate. The request-scoped session for `/mcp` is no longer relied on for the tool's connection, so there's no cross-task close.

## Reviewer notes

- Prompts return static text and there are no MCP resources, so `requires_scope` is the complete set of server-task DB entry points.
- The regression test drives a real JSON-RPC `tools/call` over the Streamable-HTTP transport against a file-backed aiosqlite engine (a real `AsyncAdaptedQueuePool`; the in-memory `StaticPool` the `db` fixture uses never returns connections and hides the bug) and asserts `pool.checkedout() == 0`. It fails on `main` and passes with this change.

### DIFF
--- a/tests/test_mcp.py
+++ b/tests/test_mcp.py
@@ -260,35 +260,40 @@ async def test_read_tool_over_http_returns_its_db_connection(
 
     engine = create_async_engine(f"sqlite+aiosqlite:///{tmp_path / 'mcp_leak.db'}")
     _db.init_app(engine=engine)
-    async with engine.begin() as conn:
-        await conn.run_sync(Base.metadata.create_all)
-
-    app = create_app(testing=True)
-    headers = {"Content-Type": "application/json", "Accept": "application/json, text/event-stream"}
-    body = {
-        "jsonrpc": "2.0",
-        "id": 1,
-        "method": "tools/call",
-        "params": {"name": "list_groups", "arguments": {}},
-    }
-
-    # A fresh /mcp request owns its session scope in production; make the
-    # ambient scope the default so RequestIdMiddleware behaves the same here.
-    token = _session_scope.set("__default__")
+    # `_db` is process-global; dispose the engine no matter what so a failed
+    # assertion below can't leak it into unrelated downstream tests (mirrors
+    # the `db` fixture's own teardown).
     try:
-        async with _mcp_client(app) as client:
-            for _ in range(3):
-                r = await client.post("/mcp", headers=headers, content=json.dumps(body))
-                assert r.status_code == 200, r.text
-    finally:
-        _session_scope.reset(token)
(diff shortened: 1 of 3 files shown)


tests/test_mcp.py
```diff
@@ -230,6 +233,64 @@ async def _call_tool(mcp_server: Any, tool_name: str, **kwargs: Any) -> str:
...
+                assert r.status_code == 200, r.text
+    finally:
+        _session_scope.reset(token)
+
+    # Force finalization of anything that a leak would have orphaned, then
+    # assert every checked-out connection made it back to the pool.
+    gc.collect()
+    await asyncio.sleep(0.05)
+    assert engine.pool.checkedout() == 0, "MCP tool call leaked a pooled DB connection"
+    await engine.dispose()
```
matthew-bass: nit, i think we might want to have a `try... finally` block here to always run `engine.dispose()` in cases where the assert fails since it seems like this test is using a process-global `db` 

> a failing assertion here could surface as confusing, unrelated failures in downstream tests. Consider moving the engine setup + `dispose()` into a small local fixture (mirroring what the `db` fixture already does in its own `finally`), or wrapping the request/assert phase in `try/finally` so `dispose()` always runs.

matthew-bass (dismissed):

tests/test_mcp.py
somethingnew2-0: Good catch — wrapped the engine setup + request/assert phase in `try/finally` so `engine.dispose()` always runs, even when the checkedout assertion fails (9b0d08a). Kept it inline rather than a fixture to stay close to the assertion it guards.

eguerrant (approved):

PR #511, opened 2026-07-07
Make plugin hook interfaces async; drop sync hooks and deprecated params

somethingnew2-0:
## What & why

`POST_MIGRATION_TODO.md` items **18** (async plugin interfaces) and **19** (drop deprecated notification params).

The app became fully async during the SQLAlchemy migration, but the four `pluggy` plugin surfaces stayed synchronous, so every call site bridged the sync hook differently: `AsyncSession.run_sync` for the session-bearing app-group-lifecycle hooks and the syncer's expiring-access notifications, and **inline on the event loop** for request-path notification/conditional-access hooks — which blocks the request whenever a plugin does network I/O.

Since 2.0 allows breaking plugin-API changes, the hooks become native `async def` and the sync variants are **removed outright** (no deprecation window). All three bridges collapse into a single `await`:

- **Lifecycle hooks** receive the request's `AsyncSession` directly (no `run_sync`).
- **Notification / conditional-access hooks** are run on the event loop.
- **metrics_reporter** is async too, so a reporter can push metrics over an async client; `batch_metrics` is now an async context manager. The pure app-group-lifecycle metadata/config/status/validation hooks intentionally **stay synchronous**.

`pluggy` never awaits hooks, so the old `@hookimpl(wrapper=True)` wrappers (which recorded the "sent" metric and swallowed plugin errors) can't survive — that cross-cutting logic moved into `send_notification` / `evaluate_conditional_access`, both routed through a shared `run_hooks_to_completion` helper. That helper uses **`asyncio.wait`, not `gather`** (following the #481 fan-out-drain review): a cancelled request doesn't tear down an in-flight hook, and one plugin failing doesn't cancel its siblings. A load-time `verify_async_impls` check fails fast if a plugin registers a *sync* impl for an async hook.

Also: completes item 19 (removes the deprecated `groups`/`roles`/`users` notification params), updates the example plugins to the async interface (the `notifications_slack` example uses Slack's native-async `AsyncWebClient`), documents the contract in the README, and **bumps `pluggy` 1.5.0 → 1.6.0** (`pyproject.toml` + `uv.lock`).

Rebased on `main` (no longer stacked): the uv/ty toolchain (#507) and the fan-out-drain fix (#481) it builds on have both merged. The request-path notifications run inside the `create_task` fan-out that #481's `drain_fan_out_tasks` awaits.


api/plugins/_async_dispatch.py
```diff
@@ -0,0 +1,90 @@
...
+
+    results: list[Any] = []
+    exceptions: list[BaseException] = []
+    for task in tasks:
+        exc = task.exception()
+        if exc is None:
+            results.append(task.result())
+        else:
+            exceptions.append(exc)
+            logger.warning("%s: plugin hook raised %r", context, exc, exc_info=exc)
```
somethingnew2-0: Open Question: Back in https://github.com/discord/access/pull/481, we modified notification plugin failures to warn instead of error. While that seems okay for notification plugins which can be noisy with connection timeout errors, this new change modifies all access' usage of all current plugins to similarly swallow errors and log as warnings. Is that something we want? Are there certain errors we expect the plugin to throw that should be reported? Or should the error reporting be done in the plugin itself?

api/plugins/app_group_lifecycle.py
somethingnew2-0, line 1: > The pure app-group-lifecycle metadata/config/status/validation hooks intentionally **stay synchronous**.

Open Question: Should these plugin hooks remain synchronous or should they allow to be async? In other words, do we ever foresee plugin implementations for metadata/config/status/validation calling external data sources that should be done async?

api/operations/approve_access_request.py
somethingnew2-0: Remove this comment and dead code

api/plugins/notifications.py
```diff
@@ -149,178 +140,33 @@ def access_group_request_completed(
...
-    except Exception:
-        # Log and do not raise since notification failures should not
-        # break the flow. Users can still manually ping approvers
-        # to process their request from the UI
-        logger.exception("Failed to execute access expiring for user notification callback")
-        return
-    _record_sent("notifications.expiring_access.sent", tags={"kind": "user"})
-    return result
-
+async def send_notification(hook_name: str, /, **kwargs: Any) -> None:
```
somethingnew2-0: Instead of passing in an arbitrary str for the hook_name, can it be an enum? Can we do this for the other plugins as well that follow this pattern of using a wrapper async function that takes a hook_name as a parameter?

api/plugins/_async_dispatch.py
somethingnew2-0: We shouldn't swallow errors thrown by plugins. So instead of logging as warning, we should log as error. This should also include the notification plugin and undo the decision made to warn in #481. Plugins themselves can and should be written to swallow their own errors.

api/plugins/app_group_lifecycle.py
somethingnew2-0: After some review with Claude, it looks like keeping these methods sync is sufficient as they're only used for retrieving static plugin config and highly unlikely to need to make network calls.

examples/plugins/notifications_slack/notifications.py
```diff
@@ -526,15 +536,15 @@ def access_expiring_user(
...
     logger.info(f"User message: {message}")
 
     # Post to the alerts channel
-    send_slack_channel_message(user, message)
+    await send_slack_channel_message(user, message)
 
 
 @notification_hook_impl
-def access_expiring_owner(
+async def access_expiring_owner(
```
eguerrant, line 545: Still uses deprecated/removed params, so does access_expiring_user in the slack example

eguerrant (dismissed):
Just the one broken example comment, looks good otherwise

examples/plugins/notifications_slack/notifications.py
somethingnew2-0: Good catch — fixed in 9c7a508. Dropped the removed `groups` param from `access_expiring_user` and `groups`/`roles`/`users` from `access_expiring_owner` so both match the current `NotificationPluginSpec`. Since pluggy dispatches hook args by name, leaving these in would have raised a `HookCallError` at call time, not just been dead params. I also AST-checked the example's other hookimpls against the spec and they're clean.

🤖 Posted by Claude Code

eguerrant (approved):

PR #520, opened 2026-07-08
feat: optional API docs in staging/prod; rename api/manage.py to api/cli.py

somethingnew2-0:
## What & why

Two small, independent changes.

### Optional OpenAPI docs outside development

Adds an `ENABLE_API_DOCS` setting (default `false`) that lets an operator expose the auto-generated OpenAPI docs (`/api/docs`) and schema (`/api/openapi.json`) in staging/production. Previously these were mounted only when `DEBUG` (i.e. `ENV=development`), so there was no way to reach them in a deployed environment — useful for debugging or client-code generation against a live deployment.

The exposure decision is centralized in a new `settings.expose_api_docs` property (`DEBUG or ENABLE_API_DOCS`) that `create_app` consults for both `docs_url` and `openapi_url`. Key safety point: both routes stay behind the app-wide `require_authenticated` gate, so flipping the flag exposes the docs to **authenticated users only** — never the public internet. Off by default so deployments don't publish their API surface unless they opt in.

### Rename `api/manage.py` → `api/cli.py`

The module is the Click-based CLI (`access sync`, `access notify`, …), not a Flask-style management shim, so the old name was misleading. The console-script entry point name (`access`) is unchanged, so cronjobs/manifests are unaffected. References updated in `pyproject.toml` (entry point + ty override), the `Dockerfile` comment, and doc strings.

## Notes for reviewers

- Rebased on latest `main`, including #507's uv/ty migration — the rename's config touchpoints landed in `pyproject.toml` (where the entry point and per-module type-checker override now live) rather than the removed `setup.py`/`.mypy.ini`.
- Verified locally with `uv run ty check .`, `uv run ruff check`/`format --check`, and the full `uv run pytest` suite (587 passing, including new coverage for the docs toggle and the `expose_api_docs` policy).


POST_MIGRATION_TODO.md
somethingnew2-0, line 1: Removing these as they're not necessary/low priority/bad ideas

eguerrant (approved):

PR #523, opened 2026-07-09
fix: paginate group members by email so the A-Z list is continuous across pages

Jeff-Rowell:
# Summary
Group/role member lists were paginated by opaque user id while the UI alphabetizes each fetched page by email, so every page independently restarted at A instead of continuing the alphabet. This orders the pagination by the member's email so the A-Z list is continuous across pages.
**Urgency**: 5 days

# Motivation
On the group detail page the member list is fetched one page at a time from `GET /api/groups/{id}/member-details` and each page is sorted alphabetically by email client-side. The backend paged over distinct users ordered by `user_id`, which is unrelated to email order, so page 1 showed an arbitrary subset sorted A-Z and page 2 showed a different arbitrary subset also sorted A-Z. The result was a list that restarts the alphabet on every page. This regressed when the member list moved from an inline (unpaginated) response to a paginated endpoint; previously the single client-side sort alphabetized the whole list.

# Description of Changes
- Order the member-details pagination query by the active user's email (with `user_id` as a stable tiebreaker) instead of by `user_id`, so page boundaries follow the same alphabetical order the UI displays.
- Order the per-page detail query by email as well, so rows within a page are also alphabetical.
- Join the active (non-deleted) user in both queries. This is required to sort by email, and it also keeps `total` aligned with the rows that actually render, since a member detail row requires an active user.
- The role detail view reuses this same endpoint, so it is covered by the same fix.

# Validation of Changes
- Added `test_get_group_member_details_paginates_in_email_order`, which assigns user ids in reverse-email order to prove pages follow email order rather than id order, and asserts the concatenated emails across two pages are globally sorted. Confirmed it fails before the change and passes after.
- Full test suite passes (565 passed). Ruff and mypy clean on the changed files.

# Guidance for Reviewers
The change is small; review `api/routers/groups.py` `get_group_member_details`. The main thing to confirm is that joining the active user to order by email is acceptable and that excluding deleted-user rows from `total` (rows that could never render anyway) is the desired behavior.

### DIFF
--- a/api/routers/groups.py
+++ b/api/routers/groups.py
@@ -18,7 +18,7 @@ from fastapi import APIRouter, HTTPException, Query
 
 from fastapi.responses import RedirectResponse
 from pydantic import TypeAdapter
-from sqlalchemy import func, nullsfirst, or_, select
+from sqlalchemy import and_, func, nullsfirst, or_, select
 from sqlalchemy.orm import joinedload, selectin_polymorphic, selectinload, with_polymorphic
 from starlette.requests import Request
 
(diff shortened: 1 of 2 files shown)


somethingnew2-0 (dismissed):

matthew-bass:
looks like this might need a rebase for the async changes but i can approve after

matthew-bass (approved):

PR #526, opened 2026-07-13
feat: cmd/ctrl+K quick switcher

somethingnew2-0:
## What

Screenshot 2026-07-13 at 12 19 16 PM

Adds a command-palette-style **quick switcher** that lets users jump to any user, group, role, or app from anywhere in the app. Open it with **⌘K / Ctrl+K** or the new search button in the app bar; results are grouped by entity type and fully keyboard-navigable (arrows to move, enter to open, esc to close).

This resolves a long-standing backlog request for a cmd+K-style switcher.

## Why / key decisions

- **Reuses the existing list endpoints** (`/api/users`, `/groups`, `/roles`, `/apps`) with their `q` param, so search behavior matches the rest of the app with no backend changes. Queries are debounced and each entity type is fetched in parallel via the generated react-query hooks.

- **Client-side relevance re-ranking.** The list endpoints return matches ordered *alphabetically*, not by relevance, and the user search additionally matches hidden Okta profile fields (e.g. a person's manager). Without ranking, searching "kevin" would surface everyone who reports to Kevin *above* Kevin himself. To fix this without touching the shared backend ordering (which the list pages depend on), the switcher over-fetches a larger candidate pool per category and re-ranks against the **visible** fields: exact > prefix > substring > matched-only-via-a-hidden-field, then trims to the top few. The API's alphabetical order is preserved as a stable tiebreak.

- **Role groups are de-duplicated** — a role group would otherwise appear under both Groups and Roles (both link to the same page), so it's shown only under Roles; app groups remain under Groups, labeled accordingly.

## Reviewer notes

- Focus handling: `autoFocus` alone is unreliable inside a MUI `Dialog` (the focus trap can reclaim focus after the open transition), so the input is explicitly focused on the dialog's `onEntered`.
- The candidate-pool size and per-category result count are small constants at the top of the component, easy to tune.
- If we later want relevance ordering to be consistent everywhere (and to work beyond the fetched pool), that would be a separate, larger change to the API sort.

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -182,7 +182,12 @@ function Dashboard({setThemeMode}: {setThemeMode: (theme: PaletteMode) => void})
   // Global cmd/ctrl+K opens the quick switcher from anywhere in the app.
   React.useEffect(() => {
     const handleKeyDown = (event: KeyboardEvent) => {
-      if ((event.metaKey || event.ctrlKey) && event.key.toLowerCase() === 'k') {
+      // Only claim the platform-native palette chord (⌘K on macOS, Ctrl+K
+      // elsewhere) so we don't clobber macOS's Ctrl+K "delete to end of line"
+      // in text fields. `event.key` can be undefined for some synthetic/IME
+      // keydowns, hence the optional chaining.
+      const modifierPressed = isMacPlatform ? event.metaKey : event.ctrlKey;
+      if (modifierPressed && event.key?.toLowerCase() === 'k') {
         event.preventDefault();
         setQuickSwitcherOpen((prev) => !prev);
       }
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -16,8 +16,10 @@ import Container from '@mui/material/Container';
 import MenuIcon from '@mui/icons-material/Menu';
 import ChevronLeftIcon from '@mui/icons-material/ChevronLeft';
 import AccountIcon from '@mui/icons-material/AccountCircle';
+import SearchIcon from '@mui/icons-material/Search';
 import {
   alpha,
+  Button,
   CssBaseline,
   PaletteMode,
   Stack,
@@ -46,6 +48,7 @@ import ListRoleRequests from './pages/role_requests/List';
 import ListTags from './pages/tags/List';
 import ListUsers from './pages/users/List';
 import NavItems from './components/NavItems';
+import QuickSwitcher from './components/QuickSwitcher';
 import NotFound from './pages/NotFound';
 import ReadApp from './pages/apps/Read';
 import ReadGroup from './pages/groups/Read';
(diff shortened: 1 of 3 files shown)


src/App.tsx
```diff
@@ -168,10 +174,23 @@ function ThemeToggle({setThemeMode, condensed}: {setThemeMode: (theme: PaletteMo
...
   const [open, setOpen] = React.useState(true);
+  const [quickSwitcherOpen, setQuickSwitcherOpen] = React.useState(false);
   const toggleDrawer = () => {
     setOpen(!open);
   };
 
+  // Global cmd/ctrl+K opens the quick switcher from anywhere in the app.
+  React.useEffect(() => {
+    const handleKeyDown = (event: KeyboardEvent) => {
+      if ((event.metaKey || event.ctrlKey) && event.key.toLowerCase() === 'k') {
```
Jeff-Rowell: On macOS, Ctrl+K is the native "delete to end of line" binding in text fields, but this handler intercepts `metaKey || ctrlKey` on every platform and calls `preventDefault()`, so Ctrl+K stops working in every input. Separately, `event.key` can be undefined for some synthetic/IME keydown events, which would throw in this window-level handler. Matching the platform-appropriate modifier plus optional chaining fixes both:

```suggestion
      const modifierPressed = isMacPlatform ? event.metaKey : event.ctrlKey;
      if (modifierPressed && event.key?.toLowerCase() === 'k') {
```

src/components/QuickSwitcher.tsx
```diff
@@ -0,0 +1,322 @@
...
+      return;
+    }
+    if (event.key === 'ArrowDown') {
+      event.preventDefault();
+      setHighlightedIndex((index) => (index + 1) % results.length);
+    } else if (event.key === 'ArrowUp') {
+      event.preventDefault();
+      setHighlightedIndex((index) => (index - 1 + results.length) % results.length);
+    } else if (event.key === 'Enter') {
+      event.preventDefault();
```
Jeff-Rowell, line 217: With `keepPreviousData`, `results` reflect `debouncedQuery`, which lags `query` by the 200ms debounce. A fast typist who highlights a row and presses Enter before the debounce fires navigates to a stale match that no longer matches the visible input. Gate Enter until the input catches up:

```suggestion
      event.preventDefault();
      // `results` reflect `debouncedQuery`; ignore Enter until the input catches up
      if (query.trim() !== trimmedQuery) {
        return;
      }
```

Jeff-Rowell (changes requested):

src/App.tsx
somethingnew2-0: Done in bc14de5 — the handler now only claims the platform-native chord (⌘K on macOS, Ctrl+K elsewhere) and uses `event.key?.toLowerCase()` to guard undefined keys. Verified in-app that Ctrl+K no longer intercepts in inputs on macOS while ⌘K still opens the palette.

src/components/QuickSwitcher.tsx
somethingnew2-0: Done in bc14de5 — Enter now returns early when `query.trim() !== trimmedQuery`, so it won't act on a stale match while the debounced query is still catching up.

Jeff-Rowell (approved):
LGTM.

PR #532, opened 2026-07-15
Add Google Group Management app-group-lifecycle plugin

somethingnew2-0:
Takes over and rebases #498 (originally by @barborico) onto current `main`. See that PR for fuller description, validation, and screenshots.

## What
Adds an opt-in example **app-group-lifecycle plugin** that mirrors each Access group to a Google group and links them via Okta group push (Okta then owns membership sync), plus the core-Access enablers it needs:

- `OktaService` gains group push-mapping create/delete/list helpers.
- Group-config plugin hooks become **app-config aware** (so a plugin's group schema/validation can reflect app-level settings, e.g. an email-prefix pattern), and gain an **`immutable`** field flag enforced host-side on update and rendered read-only in the UI.
- New `ModifyGroupPluginData` operation owns the `plugin_data` merge and fires `group_updated` **only when a plugin's configuration actually changes** — never on status-only writes, which would otherwise re-trigger reconciliation in a loop. Group plugin config is validated up front on create/update (400 instead of a later reconcile failure).
- The bulk-sync hook `sync_all_group_membership` is renamed to `sync_all_groups` (it reconciles all group state, not just membership), with a renamed `access sync-app-groups` CLI command.
- The image ships the plugin source under `/app/plugins` without installing it, so a downstream image opts in by pip-installing it.

## Why this is a fresh rebase, not a merge
#498 was branched before three things now on `main`: the async-SQLAlchemy migration (#480/#481), the `uv`/`ty` toolchain (#507), and — most relevant — the **async plugin interface** (#511). So this isn't a mechanical rebase: every backend surface #498 added had to be reworked for async. The plugin's lifecycle hooks are now `async def` and receive the request `AsyncSession`; Okta calls are awaited; the blocking Cloud Identity client runs off the event loop via `asyncio.to_thread`; and the operation/router/CLI paths were converted to awaited async SQLAlchemy. The 19 original commits are re-expressed as 6 logically-grouped, async-native commits.

The plugin's own behavior (idempotent reconcile, Access-as-source-of-truth with backfill from live Google/Okta state, immutable group email, per-group sync status) is unchanged from #498 — see that PR's description and screenshots for the functional walkthrough.

_Rebase and async conversion by Claude; original plugin design and implementation by @barborico in #498._


api/operations/modify_group_plugin_data.py
```diff
@@ -0,0 +1,83 @@
...
+        changed = bool(old_plugin_data) and merged_plugin_data != old_plugin_data
+        if changed:
+            for key in old_plugin_data:
+                if key not in merged_plugin_data:
+                    merged_plugin_data[key] = old_plugin_data[key]
+
+        # Decide whether to fire the hook before the per-plugin merge below, which
+        # mutates old_plugin_data in place (it shares the configuration/status dicts)
+        # and would otherwise erase the difference we're testing for.
+        config_changed = plugin_id is not None and is_plugin_config_changed(
```
Jeff-Rowell: `config_changed` is computed against `merged_plugin_data` *before* `merge_app_lifecycle_plugin_data` restores the per-plugin config keys the patch omitted. So a valid partial patch (for example one that omits the immutable `email`, which is now allowed by the present-only immutable check) reads as a config change and fires a redundant `group_updated` reconcile even though the effective merged config is identical. It contradicts this class's stated invariant that the hook fires only on real configuration changes.

Fix: snapshot the old config, run the merge first, then compare the fully-merged effective config against the snapshot. Requires `import copy` at the top of the file.

```python
        old_plugin_data = self.group.plugin_data or {}
        old_name = self.group.name
        old_description = self.group.description or ""
        plugin_id = get_app_group_lifecycle_plugin_to_invoke(self.group)

        # merge_app_lifecycle_plugin_data below mutates old_plugin_data in place (it shares
        # the configuration/status dicts), so snapshot what we compare against up front.
        old_config_snapshot = copy.deepcopy(old_plugin_data)

        # Apply the patch: start from the incoming data, preserving top-level plugin
        # entries it didn't mention.
        merged_plugin_data = dict(self.plugin_data)
        changed = bool(old_plugin_data) and merged_plugin_data != old_plugin_data
        if changed:
            for key in old_plugin_data:
                if key not in merged_plugin_data:
                    merged_plugin_data[key] = old_plugin_data[key]

        self.group.plugin_data = merged_plugin_data

        # Preserve per-plugin configuration/status keys the patch omitted for app group
        # lifecycle plugins (partial-patch semantics).
        if changed and type(self.group) is AppGroup:
            merge_app_lifecycle_plugin_data(self.group, old_plugin_data)

        # Compare the fully-merged effective configuration against the pre-merge snapshot, so a
        # partial patch that only omits an unchanged config key is not misread as a change
        # (which would fire a redundant reconcile).
        config_changed = plugin_id is not None and is_plugin_config_changed(
            old_config_snapshot, self.group.plugin_data or {}, plugin_id
        )
        self.config_changed = config_changed
```

api/routers/groups.py
```diff
@@ -382,16 +427,29 @@ async def put_group(
...
+        config_changed = plugin_op.config_changed
+    else:
+        await db.commit()
+
+    # Fire group_updated exactly once, covering the name/description change (ModifyGroupDetails)
+    # and/or the config change (ModifyGroupPluginData), whose per-operation fires were suppressed
+    # above. A type conversion emits its own group_created/group_deleted via ModifyGroupType; this
+    # fire simply no-ops when no lifecycle plugin applies to the group's final type.
+    name_or_desc_changed = group.name != old_name or (group.description or "") != old_description
+    if name_or_desc_changed or config_changed:
```
Jeff-Rowell: When a PUT converts a group to `app_group` via `ModifyGroupType` (which fires its own `group_created`) and also renames it, `name_or_desc_changed` is true, so this block fires `group_updated` as well. The plugin then reconciles twice for one request (duplicate Google/Okta round-trips; harmless only because reconcile is idempotent).

Fix: capture whether the type changed and skip the consolidated fire when it did, since `ModifyGroupType` already emitted the right lifecycle event.

At the `body_type` assignment:

```python
    body_type = body.type
    type_changed = body_type != group.type
    if type_changed:
```

At this fire block:

```python
    if not type_changed and (name_or_desc_changed or config_changed):
        await invoke_app_group_lifecycle_hook(
            AppGroupLifecycleHook.GROUP_UPDATED,
            group=group,
            old_name=old_name,
            old_description=old_description,
        )
```

api/routers/groups.py
```diff
@@ -109,6 +109,34 @@ async def _load_group_with_options(db: DbSession, group_id: str) -> OktaGroup |
...
+    if plugin_id is None:
+        return
+
+    from api.plugins.app_group_lifecycle import validate_app_group_lifecycle_plugin_group_config
+
+    try:
+        errors = validate_app_group_lifecycle_plugin_group_config(
+            plugin_data, plugin_id, app_plugin_data, old_plugin_data=old_plugin_data
+        )
+    except ValueError as e:
```
Jeff-Rowell, line 137: `validate_app_group_lifecycle_plugin_group_config` can raise `AppGroupLifecyclePluginFilteringError` (a plain `Exception`, not `ValueError`) when the configured plugin does not return exactly one response for a hook. The update path now makes a second such call (`get_plugin_group_config_properties` for immutable-field handling), so a plugin that implements `validate_plugin_group_config` but not `get_plugin_group_config_properties` turns a PUT into an unhandled 500 instead of a clear error. Catch it explicitly:

```python
    from api.plugins.app_group_lifecycle import (
        AppGroupLifecyclePluginFilteringError,
        validate_app_group_lifecycle_plugin_group_config,
    )

    try:
        errors = validate_app_group_lifecycle_plugin_group_config(
            plugin_data, plugin_id, app_plugin_data, old_plugin_data=old_plugin_data
        )
    except ValueError as e:
        raise HTTPException(400, f"plugin_data: {e}") from e
    except AppGroupLifecyclePluginFilteringError as e:
        raise HTTPException(500, f"Misconfigured app group lifecycle plugin '{plugin_id}': {e}") from e
    if errors:
        raise HTTPException(400, f"plugin_data: {errors}")
```

A filtering error is a server-side plugin misconfiguration rather than bad client input, so a 500 with a clear message is more honest than a 400; either is better than an unhandled stack trace.

api/services/okta_service.py
```diff
@@ -492,6 +492,95 @@ async def list_owners_for_group(self, groupId: str) -> list[User]:
         # loose accessor other okta reads return so callers stay uniform.
         return list(map(lambda owner: User(owner), owners))
 
+    # https://developer.okta.com/docs/api/openapi/okta-management/management/tag/GroupPushMapping/#tag/GroupPushMapping/operation/createGroupPushMapping
+    async def create_group_push_mapping(self, appId: str, sourceGroupId: str, targetGroupId: str) -> dict[str, Any]:
```
Jeff-Rowell: These three new push-mapping methods each re-inline the same open-client, `create_request`, check-error, `_call(execute)`, check-error dance; only `delete_group_push_mapping` factored it into a nested `_request`. Lift that to a shared private method so the request contract lives in one place:

```python
    async def _group_push_request(
        self, method: str, url: str, body: dict[str, Any], *, ignore_statuses: tuple[int, ...] = ()
    ) -> Any:
        """Issue one Group Push Mapping API request. Raises on error unless the returned status
        is in ignore_statuses (used for idempotent delete on a 404). Returns the response, or
        None when an ignored status was returned."""
        async with self._okta_client() as client:
            request_executor = client.get_request_executor()
            request, error = await request_executor.create_request(
                method=method, url=url, body=body, headers={}, oauth=False
            )
            if error is not None:
                raise Exception(error)
            response, error = await OktaService._call(request_executor.execute(request))
        if error is not None:
            if getattr(error, "status", None) in ignore_statuses:
                return None
            raise Exception(error)
        return response
```

The three methods then collapse to their validation plus policy:

```python
    async def create_group_push_mapping(self, appId, sourceGroupId, targetGroupId) -> dict[str, Any]:
        # ... existing appId/sourceGroupId/targetGroupId validation ...
        response = await self._group_push_request(
            "POST",
            f"/api/v1/apps/{appId}/group-push/mappings",
            {"sourceGroupId": sourceGroupId, "targetGroupId": targetGroupId, "status": "ACTIVE"},
        )
        assert response is not None
        return response.get_body()

    async def delete_group_push_mapping(self, appId, mappingId, deleteTargetGroup=False) -> None:
        # ... existing appId/mappingId validation ...
        base = f"/api/v1/apps/{appId}/group-push/mappings/{mappingId}"
        # Deactivate first; Okta rejects deleting an ACTIVE mapping. 404 means already gone.
        await self._group_push_request("PATCH", base, {"status": "INACTIVE"}, ignore_statuses=(404,))
        await self._group_push_request(
            "DELETE", f"{base}?deleteTargetGroup={str(deleteTargetGroup).lower()}", {}, ignore_statuses=(404,)
        )

    async def list_group_push_mappings(self, appId) -> list[dict[str, Any]]:
        # ... existing appId validation ...
        response = await self._group_push_request("GET", f"/api/v1/apps/{appId}/group-push/mappings", {})
        assert response is not None
        mappings = list(response.get_body() or [])
        while response.has_next():
            more, error = await OktaService._call(response.next())
            if error is not None:
                raise Exception(error)
            mappings.extend(more or [])
        return mappings
```

api/cli.py
```diff
@@ -425,5 +421,12 @@ async def sync_app_group_memberships() -> None:
     click.echo("Completed app group lifecycle plugin sync")
 
 
+@cli.command("sync-app-groups")
```
Jeff-Rowell, line 425: Heads up on backward compatibility: `sync-app-group-memberships` (this command) and the `sync_all_group_membership` hook both exist on `main` today, so this is a rename of already-shipped surface, not a net-new name. There is no compatibility alias, so:

- any external scheduler invoking `access sync-app-group-memberships` (for example a periodic job defined outside this repo) breaks with "No such command" on deploy; and
- any third-party lifecycle plugin still implementing `sync_all_group_membership` silently stops being invoked, since it no longer matches a hookspec.

examples/plugins/app_group_lifecycle_google/plugin.py
```diff
@@ -0,0 +1,746 @@
...
+            "displayName": display_name,
+            "description": description or "",
+            "labels": {GROUP_DISCUSSION_FORUM_LABEL: ""},
+        }
+        # initialGroupConfig=EMPTY: the admin-role service account creates the group without
+        # being added as an owner; Okta group push owns membership and ownership.
+        operation = await asyncio.to_thread(
+            lambda: self._groups_api.create(body=body, initialGroupConfig="EMPTY").execute()
+        )
+        created = operation.get("response") or {}
```
Jeff-Rowell: Minor / non-blocking, and scoped to this example plugin (opt-in, not installed), so the blast radius is limited to downstream adopters: `groups.create(...).execute()` returns a Cloud Identity long-running Operation, and this reads `operation["response"]` assuming `done: true`. Group creation is normally synchronous so this usually holds, but if the API ever returns a still-running operation, `response` is absent, `name` is None, and the create raises and marks the group `SYNC_ERROR` even though the group was created. If you want to harden the example, poll the operation until `done` before reading `response`.

Jeff-Rowell (changes requested):

api/operations/modify_group_plugin_data.py
barborico: Fixed in 439fdd2. `config_changed` is now computed against the fully-merged effective config: I snapshot the pre-merge config (the partial-patch merge mutates it in place), run the merge first, then compare. So a partial patch that only omits an unchanged config key no longer misreads as a change and fires a redundant reconcile. Added a regression test (`test_partial_patch_omitting_unchanged_config_key_does_not_fire`).

api/routers/groups.py
barborico: Reworked in 439fdd2 to fire exactly one app group lifecycle hook per PUT, after every field is applied. `ModifyGroupType` gained a `fire_created_hook` flag so `put_group` can defer the `group_created` fire to the end — after `plugin_data` is applied — and choose the event by the transition:

- convert **to** app group → `group_created` (deferred; previously it fired inside `ModifyGroupType` *before* the config was applied, and the consolidated block then also fired `group_updated` — a double reconcile that also hid the new config from `group_created`);
- same-type edit with name/description/config change → `group_updated`;
- convert **away** from app group → `group_deleted`, unchanged, still fired inside `ModifyGroupType` (it needs the pre-conversion AppGroup row, so it can't be deferred).

Tests cover all three transitions (single fire; `group_created` observes the final config).

api/routers/groups.py
barborico: Fixed in 439fdd2. `_validate_group_plugin_config` now catches `AppGroupLifecyclePluginFilteringError` and raises a clean `HTTPException(500, "Misconfigured app group lifecycle plugin '<id>': ...")` — a server-side plugin misconfiguration rather than an unhandled stack trace. Added a test asserting the 500 and the message.

api/services/okta_service.py
barborico: Obsolete after the rebase onto `main`: these three methods were refactored to go through the typed `GroupPushMappingApi` client (commit 834d026) instead of hand-rolling `create_request` + `_call(execute)`. The duplicated request dance no longer exists, and the methods now differ enough (create parses a raw-body fallback, delete does deactivate-then-delete with 404 tolerance, list paginates) that a shared `_group_push_request` helper wouldn't factor cleanly. Resolving.

api/cli.py
barborico: Accepting this rename without a compatibility alias, intentionally. This ships as part of Access 2.0, which introduces a number of breaking changes (e.g. the async hook specs), so we're deliberately omitting the deprecated command/hook shims that would preserve backward compatibility. External schedulers and third-party plugins will need to move to the new `sync-app-groups` command and hook name as part of the 2.0 upgrade.

examples/plugins/app_group_lifecycle_google/plugin.py
barborico: Obsolete after the rebase: group creation no longer goes through the Cloud Identity create API. The plugin now creates the Google group via Okta group push (commit 74acdec), which creates the target group and the downstream Google group in one step, so `_create_google_group` and the `operation["response"]` read were removed. The long-running-operation concern no longer applies. Resolving.

examples/plugins/app_group_lifecycle_google/plugin.py
```diff
@@ -0,0 +1,786 @@
...
+            return
+        update_mask = ",".join(sorted(body))
+        await asyncio.to_thread(
+            lambda: self._groups_api.patch(
+                name=self._resource_name(google_group_id), body=body, updateMask=update_mask
+            ).execute()
+        )
+
+    async def _delete_google_group(self, google_group_id: str) -> None:
+        await asyncio.to_thread(lambda: self._groups_api.delete(name=self._resource_name(google_group_id)).execute())
```
Jeff-Rowell: `_delete_google_group` is the only Google-side path that isn't tolerant of an already-absent group. Every read path (`_lookup_google_group_id`, `_email_from_status`, `_owned_group_id`) tolerates absence, but this bare `delete().execute()` does not.

In `group_deleted` the mapping unlink is wrapped in a best-effort `try/except`, but this call runs outside it. So if the Google group was already deleted out of band (or on a delete replay/retry, which the reconcile design intends to be safe), `delete()` raises, the `group_deleted` hook raises, and `invoke_app_group_lifecycle_hook` rolls back instead of committing. The "authoritative cleanup" the surrounding comment describes then fails on a group that is already gone.

Note it should NOT reuse `_is_group_absent_error` here: that treats 403 as absent, whose rationale ("a genuine permission problem surfaces on the subsequent create call") does not hold on the delete path. A 403 on delete most likely means the service account lacks delete permission on a group that still exists; swallowing it would log a phantom successful deletion and leak the group. Only a true 404 is an unambiguous "already gone":

```suggestion
    async def _delete_google_group(self, google_group_id: str) -> None:
        try:
            await asyncio.to_thread(lambda: self._groups_api.delete(name=self._resource_name(google_group_id)).execute())
        except HttpError as e:
            # 404 -> already gone, so a delete replay is a no-op. Deliberately 404-only (not
            # _is_group_absent_error): a 403 on delete more likely means we lack permission on a
            # group that still exists, which must surface rather than masquerade as a deletion.
            if getattr(getattr(e, "resp", None), "status", None) != 404:
                raise
```

(One thing to confirm: that Cloud Identity's `groups.delete` returns 404, not a conflated 403, for an already-deleted group. If it returns 403, a replay would still error here, which is the safe direction.)

Dockerfile
```diff
@@ -74,6 +74,12 @@ COPY ./api ./api
 COPY ./migrations ./migrations
 COPY ./config ./config
 COPY alembic.ini ./
+
+# Ship prod-ready plugin source into the image WITHOUT installing it here.
+# Downstream images opt in by pip installing the plugins, which keeps the source
+# single-homed in this repo while leaving activation to the consumer.
+COPY ./examples/plugins/app_group_lifecycle_google ./plugins/app_group_lifecycle_google
```
somethingnew2-0: Can we rebase this off of https://github.com/discord/access/pull/552 and add this plugin install as optional, but enable it via a build argument in the [docker-image.yml](https://github.com/discord/access/blob/main/.github/workflows/docker-image.yml) Github Action?

src/components/AppGroupLifecyclePluginConfigurationForm.test.tsx
somethingnew2-0, line 1: We don't have any other frontend tests and nothing runs them in CI. Do we want to start including these or should they be removed?

api/operations/modify_group_plugin_data.py
```diff
@@ -0,0 +1,89 @@
...
+from api.plugins.app_group_lifecycle import (
+    AppGroupLifecycleHook,
+    get_app_group_lifecycle_plugin_to_invoke,
+    invoke_app_group_lifecycle_hook,
+    is_plugin_config_changed,
+    merge_app_lifecycle_plugin_data,
+)
+
+
+class ModifyGroupPluginData:
```
somethingnew2-0, line 15: Should ModifyGroupDetails call into this new operation instead of it running `invoke_app_group_lifecycle_hook(AppGroupLifecycleHook.GROUP_UPDATED...)` itself?

api/plugins/app_group_lifecycle.py
```diff
@@ -248,13 +277,14 @@ async def group_members_removed(
         """
 
     @hookspec
-    async def sync_all_group_membership(self, session: AsyncSession, app: App, plugin_id: str | None) -> None:
+    async def sync_all_groups(self, session: AsyncSession, app: App, plugin_id: str | None) -> None:
```
somethingnew2-0, line 292: Let's make sure any internal app group plugins are updated with this change to the plugin spec

examples/plugins/app_group_lifecycle_google/plugin.py
```diff
@@ -0,0 +1,786 @@
...
+    AppGroupLifecyclePluginConfigProperty,
+    AppGroupLifecyclePluginMetadata,
+    AppGroupLifecyclePluginStatusProperty,
+    get_config_value,
+    get_status_value,
+    hookimpl,
+    set_config_value,
+    set_status_value,
+)
+from api.services import okta  # used by the push-mapping/discovery helpers
```
somethingnew2-0: Not sure how much I like plugins being able to call Access API methods directly. I wonder if there's a way we can expose the Okta group mapping methods via the plugin interface?

examples/plugins/app_group_lifecycle_google/plugin.py
```diff
@@ -0,0 +1,786 @@
...
+                )
+        await self._delete_google_group(google_group_id)
+        logger.info(f"Deleted Google group {google_group_id} for Access group {group.name}")
+
+    @hookimpl
+    async def sync_all_groups(self, session: AsyncSession, app: App, plugin_id: str | None) -> None:
+        if plugin_id is not None and plugin_id != PLUGIN_ID:
+            return
+        groups = (
+            await session.scalars(
```
somethingnew2-0: Feels weird to allow the plugin to query the Access database directly, should the app groups be passed into this plugin method instead or maybe accessible from the `app` parameter?

api/plugins/app_group_lifecycle.py
barborico: I confirmed we don't have any today; GitHub handles the group syncing rather than the plugin.

src/components/AppGroupLifecyclePluginConfigurationForm.test.tsx
barborico: Adding frontend test framework in https://github.com/discord/access/pull/564.

barborico: Added frontend test command and CI job in https://github.com/discord/access/pull/564

examples/plugins/app_group_lifecycle_google/plugin.py
barborico: I just tested, and the API endpoint returns a 403 on a missing group. So we should use the helper here to handle 403s and 404s.

api/operations/modify_group_plugin_data.py
barborico: Given the sequencing of operations in put_group (modify details, then type, then plugin config), I think it's best to keep them separate. The plugin config validation depends on the group type.

examples/plugins/app_group_lifecycle_google/plugin.py
barborico: Done. Moved the Okta group-push operations — create-by-name, link an already-imported target, discover an existing mapping, and delete/unlink — behind the plugin interface as functions on `api/plugins/app_group_lifecycle.py`, alongside the existing config/status helpers (41560dc), and reworked the plugin to import those instead (5414dcb). The example plugin no longer imports `api.services.okta` at all.

examples/plugins/app_group_lifecycle_google/plugin.py
barborico: Done in 5414dcb — `sync_all_groups` now reads `app.active_app_groups` instead of running its own `select(AppGroup)` query, so the app groups come from the `app` parameter as you suggested. One direct query remains in `_get_google_group_owner` (the cross-app ownership check): it has no `app` to read from and deliberately spans every app configured with this plugin, so it stays a scoped SQL lookup rather than an iteration over loaded state. We can refactor later if needed.

examples/plugins/app_group_lifecycle_google/plugin.py
barborico: Implemented in 5414dcb: `_delete_google_group` now wraps the delete in a `try/except HttpError` and treats 403/404 as absent via the same `_is_group_absent_error` helper the read paths use, so an out-of-band deletion or a delete replay no longer raises out of `group_deleted` (which would otherwise roll the surrounding host operation back). Added regression tests for the 403/404 and the non-absent (500) cases.

Dockerfile
barborico: Done in 505b9c7. Folded the plugin into the shared opt-in build-arg pattern from #552: added `INSTALL_GOOGLE_GROUP_LIFECYCLE_PLUGIN` (default `false`) plus its install line, dropped the one-off copy-but-don't-install `COPY`, enabled the arg in the `docker-image.yml` release build so the published image includes it, and documented it in `examples/plugins/README.md`. Default `docker build .` stays byte-identical. One heads-up: the plugin instantiates at import (reads `GOOGLE_WORKSPACE_*` env vars and Google ADC), so any image built with the arg enabled must supply those at runtime.

barborico (approved):

Jeff-Rowell (approved):

examples/plugins/app_group_lifecycle_google/plugin.py
```diff
@@ -0,0 +1,739 @@
...
+from typing import Any
+
+from google.auth import default
+from googleapiclient.discovery import build
+from googleapiclient.errors import HttpError
+from sqlalchemy import select, text
+from sqlalchemy.ext.asyncio import AsyncSession
+
+from api.models import App, AppGroup
+from api.operations import ModifyGroupDetails
```
somethingnew2-0, line 22: Similar to my previous comments: plugins should not call into Access api methods directly and instead (if those methods are necessary) add them as apart of the plugin interface.

@barborico please fix in a follow-up PR

somethingnew2-0: In general, plugins should be fully self-contained and relying purely on the plugin interface for integration with Access. They should never tightly integrate with internal Access functionality. We should update the AGENTS.md and README to reflect this

barborico: Addressed in https://github.com/discord/access/pull/577

PR #533, opened 2026-07-15
Defer all request-path notifications (follow-up to #512)

somethingnew2-0:
Follow-up to #512 (item 10, merged). #512 deferred the Okta/notification fan-out for the *fan-out operations* (`ModifyGroupUsers`/`ModifyRoleGroups`) — i.e. approving an access/role request. But every **other** request-path notification still dispatched inline via `await send_notification(...)`, so the HTTP response blocked on the notification's network I/O until it had been sent. That was every **reject** (access/role/group), the request-**created** notifications, and **group-request approval** — which is what made the UI spinner hang.

### What changed
A new `defer_notification(session, hook_name, *, detach=..., **kwargs)` helper (`api/operations/_fan_out.py`): it `expunge`s the hook payload (when the request opted into deferral), spawns `send_notification` as a task, and hands it to the post-response `BackgroundTask` drain — falling back to an inline drain outside an HTTP request (CLI, syncer, direct `execute()`).

All request-path notification sites now route through it:
- the 3 reject ops, the 3 create ops, and `ApproveGroupRequest` (previously inline);
- `ModifyGroupUsers`/`ModifyRoleGroups` migrated onto it too (dropping their bespoke `_notify_*` methods), so there's a single deferral path.

The syncer's `access_expiring_*` notifications stay inline — they're a batch CronJob with no response to return.

### Reviewer notes
- **Detached-snapshot contract.** The deferred hook runs after the router's `db.expire_all()` and session teardown, so `defer_notification` expunges the payload (request, group/role, requester, approvers) first; the async hook may read only already-loaded attributes. Documented in `NotificationPluginSpec`, the README, and `.claude/CLAUDE.md`.
- A few tests that reused a request ORM object after an HTTP resolve now re-fetch it, since the deferred notification expunges it.
- Full suite green; `ruff` + `ty` clean.


api/operations/modify_group_users.py
```diff
@@ -741,24 +741,26 @@ async def execute(self) -> OktaGroup:
...
                 await db.session.refresh(access_request, attribute_names=["resolved_at"])
                 requester = await db.session.get(OktaUser, access_request.requester_user_id)
                 approvers = await get_all_possible_request_approvers(access_request)
-                notification_payload.extend([access_request, group, requester, *approvers])
-                async_tasks.append(
-                    asyncio.create_task(self._notify_access_request(access_request, group, requester, approvers))
+                # Deferred to the post-response drain; `defer_notification` expunges
+                # the payload so the async hook can read it after the router's
+                # `db.expire_all()` and session teardown.
+                await defer_notification(
```
eguerrant: Here and in modify_role_groups:

Calling defer_notification per-iteration here means that outside a deferred HTTP request (CLI/syncer/direct execute()), each notification is awaited to completion before the next iteration even starts — these used to run concurrently via a single asyncio.wait at the end. Splitting defer_notification into a prepare_notification_task (detach + create_task, no drain) + the existing single-task defer_notification wrapper, then have this loop (and the two analogous loops in modify_role_groups.py) append into async_tasks and drain once after the loop changes the bhavior to be the same as the old _notify_access_request code. Doesn't affect the HTTP path (deferred calls don't block either way), just the non-deferred one.

eguerrant (dismissed):

api/operations/modify_group_users.py
somethingnew2-0: Good catch — split out `prepare_notification_task` (detach + create_task, no drain); both fan-out loops now append into `async_tasks` and drain once with the Okta tasks, so on the non-deferred path (CLI/syncer/direct-`execute()`) the notifications run concurrently again. The single-site callers keep the `defer_notification` wrapper. No change to the deferred HTTP path. (27b6f58)

eguerrant (approved):

PR #540, opened 2026-07-16
perf: trim eager loads that routes never emit (POST_MIGRATION_TODO #4)

somethingnew2-0:
Completes the remaining bullet of POST_MIGRATION_TODO item 4 (eager-loading
hygiene): a couple of read routes eager-loaded relationships their response
schema never serializes and no handler reads, so each request paid for dead
JOINs / selectin round-trips.

- **`group_requests._load_options()`** — dropped `active_requester`,
  `active_resolver`, and the `approved_group` relationship. `GroupRequestDetail`
  (the schema for both the detail response and the paginated list) exposes only
  `requester` / `resolver` and the `approved_group_id` *column*, so those three
  JOINs were discarded on every row of every page. The approve/resolve
  operations re-query their own object graph, so no handler needed them.
- **`roles.get_role`** — stopped reusing groups' polymorphic
  `DEFAULT_LOAD_OPTIONS` and now uses a role-specific loader. `RoleGroupDetail`
  serializes only tags and the role→group *association* mappings; the shared
  tuple additionally selectin-loaded the own-group
  `active_role_member/owner_mappings` (structurally empty for a role — a role
  can't be the target of a `RoleGroupMap`, so two wasted round-trips) plus
  AppGroup-only loaders that never apply to a role row.

The strict serialization path (`validated` / `dump_orm`, which raises on any
unloaded `lazy="raise_on_sql"` relationship a response touches) is the safety
net: the full suite stays green, so nothing actually serialized was trimmed.
Routers that already split detail vs. summary loaders (`access_requests`,
`role_requests`, `groups`, `apps`, `users`, `tags`, `audit`) were audited and
left unchanged.

### DIFF
--- a/api/routers/group_requests.py
+++ b/api/routers/group_requests.py
@@ -28,12 +28,15 @@ router = APIRouter(prefix="/api/group-requests", tags=["group-requests"], depend
 
 
 def _load_options() -> tuple:
+    # `GroupRequestDetail` — the response schema for every route below (list
+    # included) — serializes only `requester` / `resolver` and the
+    # `approved_group_id` *column*. It exposes neither `active_requester`,
+    # `active_resolver`, nor the `approved_group` relationship, and no handler
+    # reads them (the approve/resolve operations re-query their own graph), so
+    # eager-loading them just adds dead JOINs to every row of every page.
     return (
         joinedload(GroupRequest.requester),
-        joinedload(GroupRequest.active_requester),
         joinedload(GroupRequest.resolver),
-        joinedload(GroupRequest.active_resolver),
-        joinedload(GroupRequest.approved_group),
     )
 
 
--- a/POST_MIGRATION_TODO.md
+++ b/POST_MIGRATION_TODO.md
@@ -6,26 +6,6 @@ Items grouped roughly by surface area; ordering within each group is rough prior
 
 ---
 
-## Database / SQLAlchemy
-
-### 4. Eager-loading hygiene
-
-**Strict serialization landed.** `api/schemas/_serialize.py` exposes
-`dump_orm`, which runs each adapter with `from_attributes=True` and lets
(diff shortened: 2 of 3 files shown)


api/routers/group_requests.py
```diff
@@ -28,12 +28,15 @@
 
 
 def _load_options() -> tuple:
+    # `GroupRequestDetail` — the response schema for every route below (list
+    # included) — serializes only `requester` / `resolver` and the
+    # `approved_group_id` *column*. It exposes neither `active_requester`,
+    # `active_resolver`, nor the `approved_group` relationship, and no handler
+    # reads them (the approve/resolve operations re-query their own graph), so
```
somethingnew2-0, line 35: Can you verify the client never reads these fields?

somethingnew2-0: Verified — yes, on two independent fronts:

1. **Response schema.** `GroupRequestDetail` (the only shape the `/api/group-requests` routes return, list and detail) serializes just `requester`, `resolver`, and the `approved_group_id` *column* — not `active_requester`, `active_resolver`, or the `approved_group` relationship. This PR doesn't touch that schema (no `api/schemas/` changes), only the eager loads, so the response body is byte-identical before and after — these fields were never on the wire for group requests.

2. **Frontend.** The generated `GroupRequestDetail` type in `src/api/apiSchemas.ts` likewise carries only `approved_group_id` / `requester` / `resolver`, and no component reads `active_requester` / `active_resolver` / `approved_group` off a group request — grep across `src/` finds those names only in the generated `AccessRequestDetail` / `AccessRequestSummary` types, which are a different (untouched) endpoint that does still serialize them.

So the loads I removed were populating ORM relationships the serializer never read; there's no client-visible change.

🤖 Posted by Claude Code

matthew-bass (approved):

PR #541, opened 2026-07-16
feat: nonce-based CSP for the SPA (POST_MIGRATION_TODO #20)

somethingnew2-0:
Completes POST_MIGRATION_TODO item 20. Drops `'unsafe-inline'` from `script-src`
and `style-src` in the production CSP and authorizes the app's inline content
with a per-response nonce instead — closing the main XSS-mitigation gap left by
the migration.

**How the nonce is threaded** (the non-obvious part):

- `SecurityHeadersMiddleware` generates a per-response nonce *before* the route
  runs and stashes it on `request.state`, then emits it in the CSP header.
- The SPA catch-all (`serve_spa`) reads that nonce and injects it into the
  served `index.html`: onto every inline `<script>`/`<style>`, plus a
  `window.__webpack_nonce__` setter as the first thing in `<head>`. MUI's
  styling engine is styled-components (`@mui/styled-engine-sc`), which reads
  `__webpack_nonce__` and stamps the nonce onto the `<style>` tags it injects
  at runtime — so the app's styles survive the strict policy. `index.html` is
  already served `no-cache`, so per-response injection is safe; the `assets/*`
  cache invariants are untouched.

**Docs exception:** FastAPI's auto-generated Swagger page bootstraps with an
inline script and CDN assets that can't carry our nonce, so `/api/docs` and
`/api/openapi.json` keep the relaxed CSP (which also still applies in
development, where the SPA is served by Vite rather than this route).

Verified against a real production build in a browser: the SPA renders fully
styled under the strict policy with zero CSP violations, and the
styled-components `<style>` element carries the matching nonce. Backend tests
cover the injection helper, the prod nonce policy, and the docs-path exception.

### DIFF
--- a/api/app.py
+++ b/api/app.py
@@ -43,8 +43,8 @@ _SCRIPT_OR_STYLE_OPEN = re.compile(r"<(script|style)(?![^>]*\bnonce=)", re.IGNOR
 def _inject_csp_nonce(html: str, nonce: str) -> str:
     """Stamp the request's CSP nonce into the SPA shell.
 
-    With `'unsafe-inline'` dropped from `script-src`/`style-src` (see
-    `api.middleware.build_csp`), every inline `<script>`/`<style>` the build
+    `script-src`/`style-src` authorize inline content by nonce (see
+    `api.middleware.build_csp`), so every inline `<script>`/`<style>` the build
     emitted (e.g. Vite's module-preload polyfill) must carry the nonce, and
     styled-components' runtime `<style>` injections are authorized by seeding
     `window.__webpack_nonce__` before the app bundle runs. External same-origin
--- a/api/app.py
+++ b/api/app.py
@@ -8,7 +8,9 @@ present.
 
 from __future__ import annotations
 
+import json
 import logging
+import re
 import sys
 from contextlib import asynccontextmanager
 from os import environ
@@ -18,6 +20,8 @@ from typing import Any, AsyncIterator, Optional
 from fastapi import FastAPI
 from fastapi.routing import APIRoute
 from starlette.middleware.cors import CORSMiddleware
+from starlette.requests import Request
+from starlette.responses import Response
 
 from api import exception_handlers, middleware
 from api.config import settings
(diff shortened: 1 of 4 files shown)


api/app.py
```diff
@@ -31,6 +35,26 @@
...
 
+# Matches the opening of any inline `<script>`/`<style>` that isn't already
+# carrying a nonce, so `_inject_csp_nonce` can stamp one on.
+_SCRIPT_OR_STYLE_OPEN = re.compile(r"<(script|style)(?![^>]*\bnonce=)", re.IGNORECASE)
+
+
+def _inject_csp_nonce(html: str, nonce: str) -> str:
+    """Stamp the request's CSP nonce into the SPA shell.
+
+    With `'unsafe-inline'` dropped from `script-src`/`style-src` (see
```
somethingnew2-0: Do not mention the past config of unsafe-inline, only focus on the current config

somethingnew2-0: Done — reworded both docstrings (`build_csp`, `_inject_csp_nonce`) to describe the current nonce-based policy directly rather than framing it as "unsafe-inline dropped". The remaining `'unsafe-inline'` reference is in the `DEBUG_CSP` comment, which is the docs/Swagger policy's actual current config.

🤖 Posted by Claude Code

tests/test_spa.py
```diff
@@ -64,3 +66,68 @@ async def test_unknown_route_falls_back_to_spa_shell_without_caching(spa_client:
...
+        '<body><div id="root"></div>'
+        '<script type="module" crossorigin src="/assets/index-abc.js"></script></body></html>'
+    )
+    resp = await spa_client.get("/apps/Github")
+    assert resp.status_code == 200
+    body = resp.text
+
+    match = re.search(r'window\.__webpack_nonce__="([^"]+)"', body)
+    assert match is not None, body
+    nonce = match.group(1)
```
matthew-bass, line 102: Optional: tie the header nonce to the body nonce here, so the security invariant — the CSP header's `'nonce-…'` matches the nonce stamped into the shell — is checked in one request. Today the header is covered by `test_prod_csp_uses_nonce_and_drops_unsafe_inline` and the body by this test, but nothing asserts they're equal. They can't diverge given the current design; this just locks that in against a future refactor that regenerates the nonce in the wrong place. No monkeypatch needed — the harness runs with `DEBUG` false, so the header already carries the nonce policy.

```suggestion
    nonce = match.group(1)
    # CSP header must authorize the exact nonce stamped into the shell.
    assert f"'nonce-{nonce}'" in resp.headers["content-security-policy"]
```

matthew-bass (dismissed):

tests/test_spa.py
matthew-bass: claude went crazy with the emdash here, but yeah, thought it could be worth adding this

somethingnew2-0: Added (`3c61afb`) — the shell test now asserts the header's `'nonce-…'` equals the nonce stamped into the served shell, so header-vs-body is checked in one request. And you're right: the harness runs with `ENV=test` (DEBUG off), so the header already carries the nonce policy — no monkeypatch needed.

(On the em-dashes: I scrubbed those from this file in the earlier comment cleanup, so there are none left here now.)

Also rebased onto current main.

🤖 Posted by Claude Code

matthew-bass (approved):

PR #542, opened 2026-07-16
test: replace factory_boy with polyfactory (POST_MIGRATION_TODO #16)

somethingnew2-0:
Completes POST_MIGRATION_TODO item 16: the test data factories move from
`factory_boy` to [`polyfactory`](https://polyfactory.litestar.dev/), and
`factory_boy` / `pytest-factoryboy` / `types-factory-boy` are dropped.

**ORM-model factories** are now `SQLAlchemyFactory` subclasses. The one wrinkle
worth flagging: polyfactory populates *every* column with random data by
default, which would clobber soft-delete (`deleted_at`), timestamps,
discriminators, FKs, and JSON config — the columns the old factories left to
model/DB defaults. So a small base factory whitelists just the columns each
factory sets (`should_set_field_value` + `__set_foreign_keys__` /
`__set_relationships__` off) and leaves the rest alone. Explicit
`build()`/`create()` kwargs still override any column (the fixtures and many
tests rely on that). To avoid churning the ~600 existing call sites and the
registered instance fixtures, factory_boy-style method names
(`create`/`create_batch`/`build_batch`) are kept as thin aliases over
polyfactory's `build`/`batch` — semantically identical, since a plain
`factory.Factory.create()` never persisted either.

**Okta SDK model factories** (`Group`/`User`/`UserSchema`) build the
openapi-generated SDK Pydantic models through `from_dict` (camelCase aliases,
the `anyOf` `profile` union). polyfactory's generic generation can't reproduce
that shape, and these mock external Okta responses rather than our own schemas,
so they become plain hand-rolled builders (no longer factory_boy) that still
honor the `config=` override the syncer tests pass.

`conftest` swaps `pytest_factoryboy.register(...)` for equivalent instance
fixtures; the `faker` fixture now comes from faker's own pytest plugin (`faker`
is promoted to a direct dependency, since the factories import it).

Item 16 was the last entry left in `POST_MIGRATION_TODO.md` (items 4, 14, and
20 have merged), so this PR also removes the tracker file — the post-migration
follow-up work is complete.

Full suite green on the default sqlite path.


tests/factories.py
```diff
@@ -1,12 +1,39 @@
+"""Test data factories.
+
+**ORM model factories** are `polyfactory` `SQLAlchemyFactory` subclasses
+(POST_MIGRATION_TODO #16 — replaces the old `factory_boy` factories). polyfactory
```
somethingnew2-0: Remove this comment about completed POST_MIGRATION_TODO, only discuss the current code in comments

tests/factories.py
```diff
@@ -1,12 +1,39 @@
...
+The `factory_boy`-style method names (`create`, `create_batch`, `build_batch`)
+are kept as thin aliases over polyfactory's `build`/`batch` so the existing call
+sites are untouched; on a plain `factory.Factory` `create()` never persisted
+either, so the semantics match.
+
+**Okta SDK model factories** (`Group`/`User`/`UserSchema`) build the
+openapi-generated SDK Pydantic models through `from_dict` (camelCase aliases, the
+`anyOf` `profile` union). polyfactory's generic generation can't reproduce that
+shape, and these mock external Okta responses rather than our own schemas, so
+they stay hand-rolled builders — just no longer via `factory_boy`.
```
somethingnew2-0: Remove all references to factory_boy unless they are still relevant to the code after this PR is merged

tests/factories.py
```diff
@@ -1,12 +1,39 @@
...
+**ORM model factories** are `polyfactory` `SQLAlchemyFactory` subclasses
+(POST_MIGRATION_TODO #16 — replaces the old `factory_boy` factories). polyfactory
+otherwise populates *every* column with random data; these factories instead
+whitelist just the columns each one sets (via `should_set_field_value` +
+`__set_foreign_keys__` / `__set_relationships__` off) and leave the rest — soft
+deletes, timestamps, discriminators, FKs, JSON config — to the model/DB defaults,
+mirroring what the old factories did. Explicit `build(...)`/`create(...)` kwargs
+still override any column (managed or not), which the fixtures and tests rely on.
+
+The `factory_boy`-style method names (`create`, `create_batch`, `build_batch`)
```
somethingnew2-0, line 11: Keep this factory_boy comment as it's relevant to the current code

tests/factories.py
somethingnew2-0: Done (b04683b) — removed the POST_MIGRATION_TODO reference; the docstring now describes the current polyfactory design directly, with no historical framing.

🤖 Posted by Claude Code

tests/factories.py
somethingnew2-0: Done (b04683b) — dropped "just no longer via `factory_boy`" and the other historical mentions ("replaces the old factory_boy factories", "matching the old factory_boy config attribute", "as in the old factory"). The only `factory_boy` references left are the ones describing the current `create`/`create_batch`/`build_batch` aliases.

🤖 Posted by Claude Code

tests/factories.py
somethingnew2-0: Kept, thanks. Also applied the general guidance across the other PRs — removed a pre-migration Flask reference from a comment I touched in #540 and a temporal wording in #545; #539/#541 code comments were already clean.

🤖 Posted by Claude Code

matthew-bass (dismissed):
Was sitting with claude trying to figure out what polyfactory gets us, and if there's a better longer-term structure:

I haven't spent much time with access tests, but looked and this seems reasonable/true:

> **The ORM factories are effectively plain builders, and I think that's worth leaning into rather than away from.** In this codebase neither library's headline feature is actually exercised: the old factories were plain `factory.Factory` (not `SQLAlchemyModelFactory`), so `create` never persisted or built relationship graphs — and the new ones disable polyfactory's schema-driven generation via `_managed`. So both before and after are functionally `Model(id=<random>, **overrides)`.

> That makes `_managed` a bit of an awkward middle: it's manual column enumeration *plus* the library's machinery. It also has a sharp edge — add a `NOT NULL`-without-default column to a model and every test on that factory fails with an `IntegrityError` until someone adds it to `_managed`. Plain builder functions have the identical failure mode with none of the `should_set_field_value` indirection or the dependency. The Okta-SDK-as-hand-rolled-builder decision here is really the same conclusion; it just wasn't extended to the ORM layer.

> **Where polyfactory would genuinely earn its keep is the Pydantic request schemas.** There are ~250 places in the suite that hand-build request bodies as raw dicts (`json={...}`), and those are exactly the thing that drifts when `api/schemas/requests_schemas.py` changes. A factory bound to those models generates valid payloads that stay in sync automatically.

Curious on if you had explored that approach, approving regardless to unblock as I don't think I have too strong of opinions.

somethingnew2-0:
@matthew-bass we could definitely lean more into polyfactory with regards to the Pydantic request schemas and using polyfactory how it should be used instead of how we previously used factory_boy

matthew-bass (dismissed):

matthew-bass (approved):

PR #543, opened 2026-07-16
perf(syncer): batch concurrent Okta reads and reuse group scans

somethingnew2-0:
## What & why

The group **membership** and **ownership** sync passes issued one Okta list call per group, serially, so a sync run's wall-clock scaled linearly with the number of groups — the dominant cost of a run. This makes those per-group Okta reads concurrent.

Each pass now fetches a batch of groups' member/owner lists concurrently via `asyncio.gather`, then reconciles each result against the DB **sequentially**. The serial reconciliation is deliberate: the syncer shares a single `AsyncSession`, which must never be used concurrently, so only the network reads fan out. Batching in bounded groups (rather than prefetching every group at once) keeps peak memory to one batch of lists and keeps the number of concurrent Okta requests modest, so rate-limit pressure stays low (the SDK still retries any transient errors). A per-group fetch failure is captured and skipped rather than aborting the batch.

The batch size is set by the `sync` command's `--group-batch-size` flag (default 10, minimum 1).

Supporting changes in the `sync` command:
- Pool one Okta client (and its aiohttp connector) for the whole run so the concurrent fan-out reuses connections.
- Fetch the group-rule listing and the group listing once each, reused across the membership and ownership passes.

`sync_groups` keeps its own group-list fetch on purpose — in authoritative mode it creates/deletes groups in Okta, so the shared snapshot is taken *after* it runs to keep the later passes' behavior identical. `sync_group_memberships` / `sync_group_ownerships` take the shared `groups` and group-rule data as optional parameters (defaulting to fetching them) plus a required `batch_size`, so standalone and test callers are unaffected.

### DIFF
--- a/api/syncer.py
+++ b/api/syncer.py
@@ -1,5 +1,7 @@
+import asyncio
 import logging
 from collections import defaultdict
+from collections.abc import AsyncIterator, Awaitable, Callable
 from datetime import date, datetime, timedelta, timezone
 
 from sqlalchemy import and_, func, or_, select
@@ -10,6 +12,8 @@ from sqlalchemy.orm import (
     selectinload,
     with_polymorphic,
 )
+
+from okta.models.group_rule import GroupRule as OktaGroupRuleType
 
 from api.extensions import db
 from api.models import (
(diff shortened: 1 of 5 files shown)


api/syncer.py
```diff
@@ -185,26 +221,41 @@ async def sync_groups(act_as_authority: bool) -> None:
...
 
     # Hydrate all groups into sql alchemy context at once
     # to avoid a roundtrip for each group
     _ = (await db.session.scalars(select(with_polymorphic(OktaGroup, [AppGroup, RoleGroup])))).all()
 
-    group_ids_with_group_rules = await okta.list_groups_with_active_rules()
+    if group_ids_with_group_rules is None:
+        group_ids_with_group_rules = await okta.list_groups_with_active_rules()
+
+    if batch_size is None:
```
somethingnew2-0: Instead of doing this check here as an optional parameter, can we do the check in cli.py and then pass down the default (or custom configured size) to sync_group_memberships and sync_group_ownerships?

somethingnew2-0: Done in a666b34. `sync_group_memberships` / `sync_group_ownerships` now take a required keyword-only `batch_size`, and the `sync` command resolves `--group-batch-size` against `SYNC_GROUP_BATCH_SIZE` once and passes the concrete value down. The `if batch_size is None` fallback is gone from the sync functions.

🤖 Posted by Claude Code

Jeff-Rowell (dismissed):

Jeff-Rowell (approved):

PR #545, opened 2026-07-16
feat: strict ty type-checking on api/operations (POST_MIGRATION_TODO #14)

somethingnew2-0:
Finishes tightening `ty` over the backend: removes the last relaxed source path
(`api/operations/**`) from the `[[tool.ty.overrides]]` block, so the **entire
`api/` package is now strictly type-checked**. Only tests, examples, and Alembic
migrations remain relaxed.

Stacked on #539 (routers / services / leaf modules) — review/merge that first;
this diff is just the operations layer.

Resolved ~291 diagnostics, all **type-only** (no runtime behavior change; the
full suite stays green and **no `# ty: ignore` suppressions were added**):

- **Optional dereferences** (the bulk): a row from `scalars(...).first()` /
  `session.get(...)` is typed `Model | None` then used as non-None, since the
  router/caller already guarantees it exists. Encoded that invariant with
  `assert <row> is not None` right after the fetch. This also cleared the
  dependent `invalid-return-type` / `invalid-argument-type` and the
  `func.now()` / enum column-assignment errors — all secondary to the same
  Optional union.
- **`Sequence` → `list`**: `.all()` returns `Sequence[X]`; wrapped in `list(...)`
  where a `list[X]` was expected. (These wraps re-indent the multi-line queries,
  which inflates the line count — the query logic is unchanged.)
- **TypedDict-union constructors** (`T | GroupDict`, `App | AppDict`,
  `Tag | TagDict`): `ty` can't narrow the dict branch through `isinstance(dict)`,
  so the dict/model branches use `cast(...)` (a runtime no-op).
- **Two signature/typing touch-ups**: widened `CheckForReason` /
  `CheckForSelfAdd` member/owner params to `list[str] | list[OktaUser]` to match
  what the group-modify paths already pass at runtime; and `cast(Table,
  Model.__table__)` for the core `delete()`/`insert()` in `ModifyGroupType`
  (keeps the emitted SQL identical — deliberately *not* rewritten to
  `delete(Model)`, which would change the SQL).

The `assert`s encode invariants the code already relied on, so they're strictly
behavior-preserving (even stripped under `-O`, behavior reverts to today's).

### DIFF
--- a/api/operations/constraints/check_for_reason.py
+++ b/api/operations/constraints/check_for_reason.py
@@ -7,7 +7,7 @@ from sqlalchemy.orm import (
 )
 
 from api.extensions import db
-from api.models import AppGroup, OktaGroup, OktaGroupTagMap, RoleGroup, RoleGroupMap, Tag
+from api.models import AppGroup, OktaGroup, OktaGroupTagMap, OktaUser, RoleGroup, RoleGroupMap, Tag
 from api.models.tag import coalesce_constraints
 
 
@@ -16,8 +16,8 @@ class CheckForReason:
         self,
         group: OktaGroup | str,
         reason: Optional[str],
-        members_to_add: list[str] = [],
-        owners_to_add: list[str] = [],
+        members_to_add: list[str] | list[OktaUser] = [],
+        owners_to_add: list[str] | list[OktaUser] = [],
     ):
         self.group_id = group if isinstance(group, str) else group.id
 
@@ -50,6 +50,7 @@ class CheckForReason:
                 .where(OktaGroup.id == self.group_id)
             )
         ).first()
+        assert group is not None
 
         if self.invalid_reason(self.reason):
             tags = [tag_map.active_tag for tag_map in group.active_group_tags]
--- a/api/operations/constraints/check_for_reason.py
+++ b/api/operations/constraints/check_for_reason.py
@@ -7,7 +7,7 @@ from sqlalchemy.orm import (
 )
 
 from api.extensions import db
-from api.models import AppGroup, OktaGroup, OktaGroupTagMap, OktaUser, RoleGroup, RoleGroupMap, Tag
+from api.models import AppGroup, OktaGroup, OktaGroupTagMap, RoleGroup, RoleGroupMap, Tag
 from api.models.tag import coalesce_constraints
 
 
@@ -16,9 +16,12 @@ class CheckForReason:
         self,
         group: OktaGroup | str,
         reason: Optional[str],
-        members_to_add: list[str] | list[OktaUser] = [],
-        owners_to_add: list[str] | list[OktaUser] = [],
+        members_to_add: list[str] = [],
+        owners_to_add: list[str] = [],
     ):
+        # `members_to_add` / `owners_to_add` are id lists: OktaUser ids for the
+        # direct group path (execute_for_group), or the group ids a role is
+        # added to for the role path (execute_for_role).
         self.group_id = group if isinstance(group, str) else group.id
 
         self.reason = reason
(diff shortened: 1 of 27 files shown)


api/operations/constraints/check_for_reason.py
somethingnew2-0, line 133: Will this still work if members_to_add is a list of OktaUser?

api/operations/constraints/check_for_reason.py
```diff
@@ -16,8 +16,8 @@ def __init__(
         self,
         group: OktaGroup | str,
         reason: Optional[str],
-        members_to_add: list[str] = [],
-        owners_to_add: list[str] = [],
+        members_to_add: list[str] | list[OktaUser] = [],
```
somethingnew2-0: Should members/owners_to_add in both this and the check_for_self_add.py constraint instead be lists of ids and the callers fixed to send lists of ids instead?

somethingnew2-0: Done (022ca91). Narrowed both `members_to_add` / `owners_to_add` back to `list[str]` on `CheckForReason` and `CheckForSelfAdd`, and fixed the only caller that was passing objects — `ModifyGroupUsers` now passes `[u.id for u in members_to_add]`. Every other caller (the group/role PUT-members routers, the approve/modify-role paths) already passed id lists, so `ty` confirmed those four `ModifyGroupUsers` args were the only ones to change.

🤖 Posted by Claude Code

api/operations/constraints/check_for_reason.py
somethingnew2-0: Good catch — it wouldn't. Here `.where(OktaGroup.id.in_(self.members_to_add))` needs group-id strings, and `execute_for_group`'s `current_user.id in self.members_to_add` needs user-id strings; a list of `OktaUser` objects matches neither. `ModifyGroupUsers` was in fact passing objects, so that self-add check was a silent no-op on that path (HTTP was still covered because the PUT-members routers run the check with ids first). Fixed by passing ids, so it now evaluates correctly. Full suite still green.

🤖 Posted by Claude Code

api/operations/modify_group_users.py
```diff
@@ -179,8 +196,8 @@ async def execute(self) -> OktaGroup:
         valid, _ = await CheckForSelfAdd(
             group=group,
             current_user=self.current_user_id,
-            members_to_add=members_to_add,
-            owners_to_add=owners_to_add,
+            members_to_add=[u.id for u in members_to_add],
```
matthew-bass, line 199: This path isn't type-only — it changes enforcement, and it's untested:
- `modify_group_users` was the only caller passing `OktaUser` objects (not ids) to `CheckForSelfAdd`, whose `execute_for_group()` gates on `current_user.id in self.owners_to_add` — so the `DISALLOW_SELF_ADD_*` constraints were silently a no-op here. `.id` correctly enables them.
- `test_disallow_self_add_modify_group_users` only covers the router path; its direct `ModifyGroupUsers` calls are setup against an ownership tag set to `False`, so they don't catch this.

Could we add an operation-level test — a non-admin self-adding via `ModifyGroupUsers(...).execute()` on a group tagged `DISALLOW_SELF_ADD_*`, asserting nothing gets added?

somethingnew2-0: Added an operation-level test (`a2615b9`): `test_disallow_self_add_membership_blocks_direct_modify_group_users` drives `ModifyGroupUsers(...).execute()` directly — a non-admin self-adding as a member on a group tagged `DISALLOW_SELF_ADD_MEMBERSHIP` — and asserts no membership is created. I confirmed it fails on the pre-fix code (objects passed to `CheckForSelfAdd`, the check no-ops, membership gets added), and passes with the id-list fix.

Also rebased onto the updated #539 (this PR is stacked on it).

🤖 Posted by Claude Code

matthew-bass (approved):

PR #549, opened 2026-07-17
perf(syncer): stream group prefetch as a bounded sliding window

somethingnew2-0:
## What & why

`_prefetch_group_okta_lists` previously used a **chunked barrier**: fetch a batch of N groups' member/owner lists concurrently, wait for *all* N, reconcile all N against the DB sequentially, then fetch the next N. Two costs fell out of that:

- The Okta connections sit idle during each batch's reconciliation phase (no fetching overlaps it).
- A single slow or large group head-of-line-blocks the other 9 in its batch — nothing in the batch is reconciled until the slowest fetch returns.

This replaces it with a **bounded sliding window**: start N fetches up front, and each time one completes, yield its result and immediately start the next group's fetch to refill the window. The Okta round trips now overlap the caller's (sequential) DB reconciliation, and a slow group no longer stalls the rest.

Preserved invariants:
- **Concurrency bound** — at most N fetches in flight (same rate-limit ceiling; driven by the `--group-fetch-concurrency` flag).
- **Peak-memory bound** — at most N in-flight lists at once (the bound that motivated batching in the first place).
- **Failure isolation** — a failing fetch surfaces its exception as the paired value instead of aborting the run; the caller skips it and the next reconcile picks it up.

One behavioral change: pairs are now yielded in **completion order** rather than input order. Each group is reconciled independently against the DB, so order doesn't matter (only the per-group log ordering shifts). The single-`AsyncSession` rule is intact — the windowed fetches do network I/O only; all `db.session` access stays on the caller's sequential path.

The caller loops (`sync_group_memberships` / `sync_group_ownerships`) are unchanged — the generator's `(group, result)` interface is the same.

## Naming (breaking CLI change)

Since the value is a bound on how many fetches run concurrently — not a chunk size — the flag and parameters are renamed to match: `--group-batch-size` → **`--group-fetch-concurrency`**, and the internal `batch_size` params → `concurrency`. The old `--group-batch-size` flag is **dropped** (no alias); it shipped only days ago in #543, so adoption is minimal. Any cronjob passing `--group-batch-size` must switch to `--group-fetch-concurrency` (the default of 10 is unchanged).

### DIFF
--- a/api/syncer.py
+++ b/api/syncer.py
@@ -50,16 +50,16 @@ _GroupRulesByGroupId = dict[str, list[OktaGroupRuleType]]
 async def _prefetch_group_okta_lists(
     groups: list[Group],
     fetch: Callable[[str], Awaitable[list[User]]],
-    batch_size: int,
+    concurrency: int,
 ) -> AsyncIterator[tuple[Group, list[User] | BaseException]]:
-    """Yield ``(group, okta_list)`` pairs, keeping up to ``batch_size`` Okta fetches in flight.
+    """Yield ``(group, okta_list)`` pairs, keeping up to ``concurrency`` Okta fetches in flight.
 
-    A bounded sliding window: ``batch_size`` fetches are started up front, and each
+    A bounded sliding window: ``concurrency`` fetches are started up front, and each
     time one finishes its result is yielded and a fetch for the next group is
     started to refill the window. This overlaps the Okta round trips with the
     caller's (sequential) DB reconciliation and keeps each fetch independent, so a
     slow group does not hold up the others, while bounding both the load on Okta's
-    rate limits and the peak memory held (at most ``batch_size`` in-flight lists).
+    rate limits and the peak memory held (at most ``concurrency`` in-flight lists).
     Pairs are yielded in completion order; each group is reconciled independently,
     so order does not matter.
 
@@ -78,7 +78,7 @@ async def _prefetch_group_okta_lists(
             in_flight[asyncio.ensure_future(fetch(group.id))] = group
 
     try:
-        for _ in range(batch_size):
+        for _ in range(concurrency):
             _start_next()
 
         while in_flight:
@@ -256,7 +256,7 @@ async def sync_group_memberships(
     groups: list[Group] | None = None,
     group_ids_with_group_rules: _GroupRulesByGroupId | None = None,
     *,
-    batch_size: int,
+    concurrency: int,
 ) -> None:
     logger.info("Membership sync started.")
     if groups is None:
@@ -269,7 +269,7 @@ async def sync_group_memberships(
     if group_ids_with_group_rules is None:
         group_ids_with_group_rules = await okta.list_groups_with_active_rules()
 
-    async for group, members in _prefetch_group_okta_lists(groups, okta.list_users_for_group, batch_size):
+    async for group, members in _prefetch_group_okta_lists(groups, okta.list_users_for_group, concurrency):
         if isinstance(members, OktaTransientError):
             logger.warning(f"Transient Okta error listing members for group {group.id}, skipping.", exc_info=members)
             continue
@@ -365,7 +365,7 @@ async def sync_group_ownerships(
     groups: list[Group] | None = None,
     group_ids_with_group_rules: _GroupRulesByGroupId | None = None,
     *,
-    batch_size: int,
+    concurrency: int,
 ) -> None:
     logger.info("Ownership sync started.")
     if groups is None:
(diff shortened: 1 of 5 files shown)


api/syncer.py
```diff
@@ -52,23 +52,54 @@ async def _prefetch_group_okta_lists(
...
-    These fetch coroutines do network I/O only and must never touch ``db.session``
-    (the concurrency rule in ``api/extensions.py``); the caller reconciles each
-    yielded pair against the DB on its own sequential code path.
+    """Yield ``(group, okta_list)`` pairs, keeping up to ``batch_size`` Okta fetches in flight.
+
+    A bounded sliding window: ``batch_size`` fetches are started up front, and each
+    time one finishes its result is yielded and a fetch for the next group is
+    started to refill the window. This overlaps the Okta round trips with the
+    caller's (sequential) DB reconciliation and avoids head-of-line blocking — a
+    slow group no longer holds up the rest of a batch — while still bounding both
```
somethingnew2-0: Change to describe the current code, don't comment on previous behavior
```suggestion
    slow group doesn't holds up the rest of a batch — while still bounding both
```

somethingnew2-0: Done in a978f47. Rewrote the docstring to describe only the current behavior — dropped "no longer holds up …" and the "mirroring `asyncio.gather(return_exceptions=True)`" reference (and "not input order"). I didn't take the suggestion verbatim since it read "doesn't holds"; the rephrase says a slow group "does not hold up the others."

🤖 Posted by Claude Code

api/syncer.py
```diff
@@ -50,25 +50,55 @@
...
+        for _ in range(concurrency):
+            _start_next()
+
+        while in_flight:
+            done, _ = await asyncio.wait(set(in_flight), return_when=asyncio.FIRST_COMPLETED)
+            for task in done:
+                group = in_flight.pop(task)
+                try:
+                    result: list[User] | BaseException = task.result()
+                except Exception as exc:
```
Jeff-Rowell, line 94: `except Exception` narrows the failure-isolation contract compared to the old `asyncio.gather(return_exceptions=True)`, which captured `BaseException`. The docstring, the yielded type (`list[User] | BaseException`), and the caller's `if isinstance(members, BaseException)` branch all promise that *any* failing fetch surfaces as the paired value and is skipped. With `except Exception`, a non-`Exception` `BaseException` from a fetch task (most plausibly a `CancelledError` if the Okta client's timeout path cancels the coroutine) now propagates out of the generator and aborts the entire sync run mid-stream instead of being isolated to that one group.

This is borderline: swallowing `CancelledError` is itself an anti-pattern, so narrowing to `Exception` may be intentional and arguably better. But then the docstring, the `BaseException` return annotation, and the caller's `isinstance(..., BaseException)` branch are overclaiming/partly dead. Suggest picking one: either widen the catch back to `BaseException` (re-raising `CancelledError`) to honor the stated contract, or keep `except Exception` and tighten the annotation/docstring/caller to `Exception`.

Jeff-Rowell (changes requested):

tests/test_syncer_prefetch.py
```diff
@@ -0,0 +1,66 @@
...
+
+    out: dict[str, Any] = {}
+    async for group, result in _prefetch_group_okta_lists(groups, fetch, 10):
+        out[group.id] = result
+
+    assert len(out) == 25
+    assert all(out[f"g{i}"] == [f"g{i}"] for i in range(25))
+    # Never exceeds the window, and the window actually fills (real concurrency).
+    assert peak <= 10
+    assert peak == 10
```
matthew-bass, line 34: think this is a redundant line

api/syncer.py
somethingnew2-0: Good catch — fixed in 5b156fa. I went with your second option: kept `except Exception` (so a `CancelledError`/shutdown propagates and stops the run rather than being swallowed) and tightened the contract to match — the local annotation, the generator's yield type, and both callers' `isinstance(...)` checks are now `Exception` instead of `BaseException`. Added a short comment at the catch site documenting why it's `Exception` and not `BaseException`, so it doesn't get "widened" back later.

🤖 Posted by Claude Code

tests/test_syncer_prefetch.py
somethingnew2-0: Fixed in 5b156fa — dropped the redundant `assert peak <= 10` and kept `assert peak == 10` (it subsumes the bound and also proves the window actually fills). Thanks!

🤖 Posted by Claude Code

Jeff-Rowell (approved):

PR #551, opened 2026-07-17
fix(sentry): stop alerting on ignored errors wrapped by the MCP tool layer

Jeff-Rowell:
# Summary
MCP tool errors are paging Sentry even though we explicitly ignore the underlying error types. This adds a `before_send` hook that drops events whose `__cause__` chain contains any type already in `ignore_errors`, closing the gap left by the framework wrapping those errors in a `ToolError`.
**Urgency**: 3 days

# Motivation
`_configure_sentry` lists client/control-flow error types in `ignore_errors` so they don't alert. But `ignore_errors` only matches the *top-level* exception type, and the MCP server wraps every exception a tool raises in a `ToolError` (`raise ToolError(...) from e` in FastMCP's `tools/base.py`). So an ignored error raised inside a tool surfaces as a non-ignored `ToolError` and leaks through.

Observed in production: a client called the `get_group` tool with `group_id` instead of its declared `group_id_or_name`. Argument validation raised a pydantic `ValidationError` (which we ignore), FastMCP re-raised it as a `ToolError` (which we don't), and it alerted.

Auditing the rest of the `ignore_errors` list: today only `ValidationError` is actually wrapped this way (the other types are raised in the HTTP/auth layers, outside tool bodies). But the same leak would silently reappear for any other ignored type raised inside a tool, so the filter checks the whole set rather than special-casing `ValidationError`.

# Description of Changes
- Add `_drop_wrapped_ignored_errors`, a Sentry `before_send` hook that walks the explicit `__cause__` chain and drops the event if any exception in it is an instance of a type in `ignore_errors`.
- Follows `__cause__` only (set by `raise ... from`), not implicit `__context__`, so a genuine failure that merely ran while handling an ignored error still alerts.
- Guards against a cyclic cause chain with a visited set.

# Validation of Changes
- New unit tests in `tests/test_sentry_filter.py`: the production case (`ToolError` wrapping `ValidationError` -> dropped), another ignored type wrapped (`ToolError` wrapping `HTTPException` -> dropped), bare ignored errors (dropped), a generic exception (kept), a `ToolError` wrapping a real server-side error (kept, so genuine bugs still alert), missing `exc_info` (kept), and a cyclic chain (terminates).
- `uv run ruff check` / `ty check` / `pytest tests/test_sentry_filter.py` all pass (7 passed).

# Guidance for Reviewers
The key design choice is doing this in `before_send` rather than adding `ToolError` to `ignore_errors`: FastMCP wraps *every* exception raised in *any* tool body in a `ToolError`, so ignoring `ToolError` outright would also silence genuine server-side failures. The `before_send` approach only suppresses events actually rooted in an already-ignored type. Note the ignored-type set is intentionally duplicated between `ignore_errors` and the hook (with a "keep in sync" comment) to keep the diff minimal; happy to hoist it into a shared constant if preferred.

### DIFF
--- a/api/app.py
+++ b/api/app.py
@@ -118,12 +118,12 @@ def _drop_wrapped_validation_errors(event: dict, hint: dict) -> Optional[dict]:
     from pydantic import ValidationError
 
     exc = (hint or {}).get("exc_info", (None, None, None))[1]
-    depth = 0
-    while exc is not None and depth < 20:
+    seen: set[int] = set()
+    while exc is not None and id(exc) not in seen:
         if isinstance(exc, ValidationError):
             return None
+        seen.add(id(exc))
         exc = exc.__cause__
-        depth += 1
     return event
 
 
--- a/api/app.py
+++ b/api/app.py
@@ -105,6 +105,28 @@ def _configure_logging() -> None:
     logging.root.addFilter(token_filter)
 
 
+def _drop_wrapped_validation_errors(event: dict, hint: dict) -> Optional[dict]:
+    """Sentry before_send: drop events whose root cause is a pydantic ValidationError.
+
+    ``ignore_errors`` only matches the top-level exception type. The MCP server
+    re-raises tool argument-validation failures as a ToolError via
+    ``raise ToolError(...) from e``, so a client passing a bad tool argument
+    surfaces as ToolError (not ignored) wrapping the ValidationError we do want
+    to ignore. Walk the explicit cause chain and drop the event when a
+    ValidationError is its root.
+    """
+    from pydantic import ValidationError
+
+    exc = (hint or {}).get("exc_info", (None, None, None))[1]
+    depth = 0
+    while exc is not None and depth < 20:
+        if isinstance(exc, ValidationError):
+            return None
+        exc = exc.__cause__
+        depth += 1
+    return event
+
+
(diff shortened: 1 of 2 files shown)


api/app.py
```diff
@@ -140,6 +162,7 @@ def _configure_sentry() -> None:
             ValidationError,
             OIDCRedirectRequired,
```
somethingnew2-0, line 178: Don't we just need to add a ToolError as ValidationError is already ignored? Or does the MCP server wrap other errors besides ValidationError in ToolError?

```suggestion
            OIDCRedirectRequired,
            ToolError
```

api/app.py
```diff
@@ -105,6 +105,28 @@ def _configure_logging() -> None:
...
+    surfaces as ToolError (not ignored) wrapping the ValidationError we do want
+    to ignore. Walk the explicit cause chain and drop the event when a
+    ValidationError is its root.
+    """
+    from pydantic import ValidationError
+
+    exc = (hint or {}).get("exc_info", (None, None, None))[1]
+    seen: set[int] = set()
+    while exc is not None and id(exc) not in seen:
+        if isinstance(exc, ValidationError):
```
somethingnew2-0: Should we include the other errors in the `ignore_errors` list below?

somethingnew2-0 (approved):

PR #552, opened 2026-07-19
Load the audit-logger example plugin in local dev; fix plugin install docs

barborico:
# Summary
Two related dev/docs fixes around example plugins after the migration to `uv`:
1. Wire `examples/plugins/app_group_lifecycle_audit_logger` into the `dev` dependency group (editable path source) so `uv sync` / `make run-backend` install and register it automatically, to exercise the app group lifecycle plugin behavior without any external dependencies.
2. Correct every example plugin README's install instructions, which still said `pip install` and no longer work against the `uv`-based image.

**Urgency**: MEDIUM; one week

# Motivation
After the `uv` migration, the plugin READMEs are inaccurate: the Access image builds with `uv`, its virtualenv is `/app/.venv` (on `PATH`, with **no `pip`**), and the app runs from that venv. So a derived image's `RUN pip install ./plugin` installs into the *system* interpreter, where the running app never finds it — and locally, `pip install -e …` lands in a different env than the `uv run`-launched server (and `make run-backend`'s `uv sync` would prune it anyway). The net symptom is "Registered 0 app group lifecycle plugin(s)".

<details>

- `pyproject.toml`: add `app_group_lifecycle_audit_logger` to the `dev` group + a `[tool.uv.sources]` editable path entry. `uv.lock` regenerated (prod build is `--frozen`). Dev-group only ⇒ excluded from prod via `uv sync --no-dev`.
- Example plugin READMEs (`app_group_lifecycle_audit_logger`, `datadog_metrics_reporter`, `health_check_plugin`, `notifications_slack`): change the Dockerfile install snippets from `pip install` to `uv pip install`, with a one-line note on why (`.venv` has no `pip`). The audit-logger's local-dev / testing instructions now point at `make run-backend` (auto-loaded via the dev group) instead of a manual `pip install -e`.

</details>

<details>

- `uv sync` builds/installs the plugin editable; `entry_points().select(group='access_app_group_lifecycle')` returns `audit_logger`.
- `make run-backend` logs `Registered 1 app group lifecycle plugin(s): ['audit_logger']` and `Application startup complete`.
- Confirmed the base image ships no `pip` in `.venv` (so the old `RUN pip install` guidance was installing into the wrong interpreter).

</details>

# Guidance for Reviewers
Two commits: (1) the dev-dependency wiring, (2) README install-instruction fixes. Open question: whether the example plugin belongs in the default `dev` group (loaded for every local `make run-backend`) or an opt-in group — easy to move.

### DIFF
--- a/examples/plugins/app_group_lifecycle_audit_logger/README.md
+++ b/examples/plugins/app_group_lifecycle_audit_logger/README.md
@@ -38,23 +38,23 @@ The plugin doesn't integrate with any external systems—it simply logs events, 
 
 ## Installation
 
-To bake the plugin into an operator's App container image, add these lines to
-the Dockerfile. The Access image is built with `uv` and its virtualenv lives at
-`/app/.venv` (which has no `pip`), so install with `uv pip install` — plain
-`pip` would install into the system interpreter, where the running app won't
-find the plugin:
+This plugin's source ships in the Access build context under
+`examples/plugins/`, but the default image does **not** install it. Enable it at
+build time with its build arg (default `false`):
 
-```dockerfile
-# Install the audit logger plugin
-WORKDIR /app/plugins
-ADD ./examples/plugins/app_group_lifecycle_audit_logger ./app_group_lifecycle_audit_logger
-RUN uv pip install ./app_group_lifecycle_audit_logger
-
-# Reset working directory
-WORKDIR /app
+```bash
+docker build --build-arg INSTALL_AUDIT_LOGGER_PLUGIN=true .
+# or, with docker compose:
+docker compose build --build-arg INSTALL_AUDIT_LOGGER_PLUGIN=true
 ```
 
-For local development in this repo, this plugin is already wired into the `dev`
+The image installs the plugin into its `uv` virtualenv (`/app/.venv`) with
+`uv pip install` — the venv has no `pip`, and plain `pip` would install into the
+system interpreter where the running app won't find it. See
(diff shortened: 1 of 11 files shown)


examples/plugins/app_group_lifecycle_audit_logger/README.md
```diff
@@ -38,23 +38,26 @@ The plugin doesn't integrate with any external systems—it simply logs events,
 
 ## Installation
 
-To install the plugin, add these lines to the App container Dockerfile:
+To bake the plugin into an operator's App container image, add these lines to
```
somethingnew2-0: Instead of requiring the Dockerfile to be modified we may want to add a build arg instead to the Dockerfile which installs it

somethingnew2-0: Something like

```
ARG INSTALL_APP_GROUP_LOGGER_PLUGIN=false

# 2. Evaluate using standard shell logic
RUN if [ "$INSTALL_APP_GROUP_LOGGER_PLUGIN" = "true" ] ; then \
        uv pip install ./app_group_logger_plugin ; \
    else \
        echo "Skipping app group logger audit logger plugin" ; \
    fi
```

somethingnew2-0 (dismissed):

somethingnew2-0 (dismissed):

somethingnew2-0 (approved):

PR #563, opened 2026-07-23
feat: search group/role members by name or email

somethingnew2-0:
https://github.com/user-attachments/assets/6b2ab415-5c4d-476d-a26e-ec8019e58fe6

## What & why

[PR #500](https://github.com/discord/access/pull/500) added pagination to the member list on the group/role Read page. For large groups that span several pages, that made it hard to find a specific user — you'd have to page through to spot them. This adds a member search so you can filter the whole group by name or email, mirroring the search the apps page already has.

Because the member list is paginated server-side, the search is computed server-side too: `GET /api/groups/{id}/member-details` gains an optional `q` that filters the paginated set by name/email in SQL. Client-side filtering wouldn't work — it would only ever see the current page. Searching resets to page 1, and the page control reflects the filtered result count, so search and pagination compose (paging through filtered results keeps the query, and totals/pages track the filter).

The debounced search field is factored into a shared `DebouncedSearchField` component and reused by the apps page, which previously carried the same debounce logic inline.

### DIFF
--- a/src/pages/groups/Read.tsx
+++ b/src/pages/groups/Read.tsx
@@ -770,7 +770,16 @@ export default function ReadGroup() {
                           </Typography>
                         </Stack>
                         <Stack direction="row" spacing={1} alignItems="center" sx={{flexWrap: 'nowrap'}}>
+                          {/*
+                           * The field is an uncontrolled Autocomplete, so MUI owns its text.
+                           * /groups/:id and /roles/:id render the same ReadGroup element, so
+                           * navigating between groups re-renders rather than remounts — keying
+                           * on id remounts the field so its text clears in lockstep with the
+                           * memberSearchQuery reset above (otherwise stale text desyncs from the
+                           * results). The apps page keys its search group the same way.
+                           */}
                           <DebouncedSearchField
+                            key={id}
                             label="Search members"
                             onSearchChange={handleMemberSearchChange}
                             sx={{width: 220}}
--- a/src/pages/groups/Read.tsx
+++ b/src/pages/groups/Read.tsx
@@ -71,6 +71,7 @@ import AppLinkButton from './AppLinkButton';
 import AvatarButton from '../../components/AvatarButton';
 import MembershipChip from '../../components/MembershipChip';
 import AppGroupLifecyclePluginData from '../../components/AppGroupLifecyclePluginData';
+import DebouncedSearchField from '../../components/DebouncedSearchField';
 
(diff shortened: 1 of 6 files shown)


src/pages/groups/Read.tsx
```diff
@@ -735,21 +754,31 @@ export default function ReadGroup() {
...
+                        <Stack direction="row" spacing={1} alignItems="center" sx={{flexWrap: 'wrap'}}>
+                          <Typography variant="h6" color="text.accent">
+                            {group.type == 'role_group' ? 'Role Members' : 'Group Members'}
+                          </Typography>
+                          <Typography variant="body1" color="text.secondary">
+                            {group.type == 'role_group' ? 'Members of Okta Group for Role' : 'Members of Okta Group'}
+                          </Typography>
+                        </Stack>
+                        <Stack direction="row" spacing={1} alignItems="center" sx={{flexWrap: 'nowrap'}}>
+                          <DebouncedSearchField
```
Jeff-Rowell, line 781: The search box can retain stale text after navigating to a different group/role, desyncing it from the results list.

`DebouncedSearchField` is an uncontrolled `Autocomplete` (only `onInputChange`; no `value`/`inputValue`/`key`), so MUI owns the displayed text. Both `/groups/:id` and `/roles/:id` render the same `<ReadGroup />` element (App.tsx), so navigating between them re-renders rather than remounts so the field instance and its text survive.

Smallest fix is to remount the field on id change so its internal text clears in lockstep with the state reset:

```suggestion
                          <DebouncedSearchField
                            key={id}
```

Jeff-Rowell (changes requested):

src/pages/groups/Read.tsx
somethingnew2-0: Good catch — fixed in 360aba2, using your `key={id}` suggestion. The field now remounts on group/role change, so its (MUI-owned) text clears in lockstep with the `memberSearchQuery` reset instead of surviving the re-render.

Verified via the quick-switcher, which navigates client-side (`navigate()`), so `ReadGroup` re-renders rather than remounts — reproducing the exact scenario: I set a `window` marker before jumping, confirmed it survived the navigation (i.e. no full reload), and the destination group's search box came up empty with the list in sync. Keyed it the same way the apps page keys its search group.

🤖 Posted by Claude Code

Jeff-Rowell (approved):

PR #568, opened 2026-07-29
Syncer bugfix: skip mid-run-deleted groups without an ERROR

barborico:
# Summary

The syncer logs an ERROR (and pages) when a group is deleted from Okta partway through a sync run, even though the situation is expected, harmless, and self-healing. Classify that 404 as `OktaResourceNotFoundError` and skip it at WARNING, alongside the existing `OktaTransientError` branch.

**Urgency**: 5 days

# Motivation

Fixes 24 occurrences of a Sentry error over ~11 weeks, 0 users impacted, every one benign.

The syncer takes one Okta group snapshot per run (`api/cli.py:295`) and reuses it for the whole membership and ownership fan-out. In production those passes run 20–68 minutes, so the snapshot ages substantially before the fan-out reaches the tail of the list. A group deleted after the snapshot was taken is still in the list being walked but already gone from Okta, so its member/owner fetch 404s. The facade raised a bare `Exception`, which the fan-out logged at ERROR.

I confirmed the diagnosis against production rather than assuming it. All 24 group IDs from Sentry are present in the Access DB with `deleted_at` set, and in **24 out of 24 cases the deletion landed during the run that 404'd, after that run's snapshot was taken** (decoding each event's cronjob pod name to recover the run's start time). No group ever 404'd twice, which rules out a stale/dangling DB reference — that failure mode would repeat on every run forever.

The deletions came from Access itself: `DeleteGroup` soft-deletes the row and removes the group from Okta, and `DeleteApp` cascades across an app's groups — which is why the 404s arrive in bursts of same-app groups.

Nothing is lost by skipping: the next run's snapshot won't contain the group, and `sync_groups` reconciles it as deleted locally.

<details>

Two commits, described in order.

### `fix(syncer)` — the alert-noise fix

**`api/services/okta_service.py`**
- New `OktaResourceNotFoundError`, documenting the snapshot-vs-authoritative-read skew that produces it. Deliberately distinct from `OktaTransientError`: nothing was flaky and a retry would not help, so it should not map to a 503 on the request path.
- New `_is_not_found_okta_error` plus a `NOT_FOUND_STATUS_CODE` constant next to the existing `TRANSIENT_STATUS_CODES`.

**`api/syncer.py`**
- A WARNING-level skip branch in both `sync_group_memberships` and `sync_group_ownerships`, ordered ahead of the generic `isinstance(..., Exception)` branch (both new and existing types are `Exception` subclasses, so order matters).

### `refactor(okta)` — uniform classification

- Classification moved from `_paginate` up into `_call`, so it covers every proxied SDK method rather than only paginated list endpoints. This mirrors the existing `OktaTransientError` handling and matches the rationale `_WrapperClient` already gives for centralizing dispatch in `_call`.
- **`delete_group_push_mapping` had to change as a direct consequence.** It treats a 404 at either the deactivate or the delete step as success, so deletion stays idempotent across a retry or replay of a partial attempt — and it did that by reading the 404 out of the returned tuple. Once `_call` raises first, that tuple never arrives. It now catches `OktaResourceNotFoundError` around both steps, which also preserves skipping the delete when the mapping already vanished at deactivate.
- **`_fake_okta_client` fidelity fix.** The fixture yielded a bare `MagicMock` where the real `_okta_client` yields `_WrapperClient(client)`, so every test using it bypassed `_call` entirely. See the reviewer callout below.
</details>

<details>

- Full suite green: 769 passed. `ruff check`, `ruff format --check`, and `ty check` all clean.
- New tests in `tests/test_okta_service.py`: a 404 surfaces as `OktaResourceNotFoundError` on both a list endpoint and a single-resource endpoint (pinning that classification lives in `_call`, not `_paginate`); a 403 does not, so genuine failures stay loud.
- New tests in `tests/test_group_membership_sync.py` and `tests/test_group_ownership_sync.py`: the affected group is skipped with a WARNING naming it, nothing is logged at ERROR from the `api.syncer` logger, and the *other* groups in the run still reconcile. Confirmed to fail before the fix and pass after (verified by reverting `api/syncer.py` and re-running — both failed on exactly the ERROR-level assertion).
- The push-mapping regression was demonstrated, not argued: with the fixture corrected, the two idempotency tests passed on the old code, failed the moment classification moved into `_call`, and passed again once the method caught `OktaResourceNotFoundError`.
- Audited every other place a 404 could leak differently — all ~15 remaining `raise Exception(error)` sites (type change only; `Exception` subclass, message preserved), `.args[]` and error-object introspection (none in `api/`), `exception_handlers.py` (no handler for the new type, so it falls to the `Exception` catch-all → 500, unchanged), `_WrapperClient`'s retry (catches `OktaTransientError` only), the syncer's mutation-path catchers, and `add_owner_to_group`'s "already assigned" tolerance (message on a 400/409, not a 404).
- `tests/test_okta_transient_errors.py::test_non_transient_http_error_not_mapped` retargeted from 404 to 403: its point is that non-transient errors pass through unmapped, and a 404 is now classified, so it would have passed for the wrong reason.
</details>

# Guidance for Reviewers

Best to review file-by-file; the second commit reverses a scoping decision made in the first, so the combined diff reads as though `_paginate` was never involved.

The callout that most merits attention is in `refactor(okta)`:

**`_fake_okta_client` was yielding a bare `MagicMock` where production yields `_WrapperClient(client)`, so every test using it bypassed `_call`.** That is why the push-mapping idempotency tests would have stayed green through a real regression: they were exercising error handling against errors that in production never reach the code under test. The fixture now yields the wrapper. It is shared by 11 tests — all still pass, but it is a fidelity change with reach beyond this PR and worth a look on its own.

Two smaller ones:

1. **Centralizing in `_call` revokes each facade method's ability to *see* a 404.** `delete_group_push_mapping` was the only call site that depended on that (checked for `== 404`/`!= 404` and error-object introspection across `api/`), but it is the kind of thing no type checker will catch, so a second pair of eyes on that grep is worthwhile.
2. **Only 404 is reclassified.** A 403 or any other error still raises loudly, with a test pinning it so it can't be widened by accident.

Also worth confirming WARNING is the level you want. It matches the neighbouring `OktaTransientError` branch and keeps the event out of Sentry's alerting, but INFO would be quieter still if this is considered fully routine.

Out of scope, noted for the record: `group_ids_with_group_rules` is fetched once per run (`api/cli.py:285`) and reused across all three passes, so `is_managed` — which gates authoritative membership deletes — is decided on a snapshot that ages the same way. Real but separate, and a behavior change rather than alert-noise cleanup.

### DIFF
--- a/api/services/okta_service.py
+++ b/api/services/okta_service.py
@@ -32,6 +32,10 @@ REQUEST_TIMEOUT = 30
 # may swallow them and let the next reconcile retry. Covers rate limiting (429)
 # and upstream/gateway blips (5xx, e.g. a 502 Bad Gateway from the load balancer).
 TRANSIENT_STATUS_CODES = frozenset({429, 500, 502, 503, 504})
+# The status a list endpoint returns when the resource it lists no longer exists —
+# chiefly a group deleted after the syncer took its group snapshot. See
+# ``OktaResourceNotFoundError``.
+NOT_FOUND_STATUS_CODE = 404
 # Page size for cursor-paginated list endpoints. Okta caps most list endpoints
 # at 200 per page; the facade drives the ``after`` cursor to walk pages.
 LIST_PAGE_LIMIT = 200
@@ -78,6 +82,23 @@ class OktaTransientError(Exception):
         self.retryable = retryable
 
 
+class OktaResourceNotFoundError(Exception):
+    """Raised when a list endpoint 404s because the resource it lists is gone.
+
+    The syncer takes one group snapshot per run (``api/cli.py``) and reuses it for
+    the whole membership and ownership fan-out, which can take tens of minutes. A
+    group deleted — usually *through Access itself*, whose ``DeleteGroup`` removes
+    it from Okta — after that snapshot was taken is still in the list being walked
+    but already absent from Okta, so its member/owner fetch 404s. That race is
+    expected and self-healing: the next run's snapshot won't contain the group and
+    ``sync_groups`` reconciles it as deleted locally. Callers in the fan-out skip
+    it at WARNING rather than ERROR so it doesn't alert.
+
+    Distinct from ``OktaTransientError``: nothing was flaky and a retry would not
+    help — the resource is genuinely gone.
+    """
+
+
 # On a gateway failure that yields no response object, the SDK returns
 # ``(None, None, error)`` from the request executor, then dereferences
 # ``response.status`` before checking the error on its list/get endpoints. That
@@ -107,6 +128,16 @@ def _is_transient_okta_error(error: Any) -> bool:
     if isinstance(error, AttributeError) and str(error) == _NONE_RESPONSE_ATTRIBUTE_ERROR:
         return True
     return str(error) == "Request Timeout exceeded."
+
+
+def _is_not_found_okta_error(error: Any) -> bool:
+    """Whether an SDK error is a 404 for a resource that no longer exists.
+
+    Both ``OktaAPIError`` (JSON error bodies) and ``HTTPError`` expose ``.status``.
+    Classified only for list endpoints (see ``_paginate``), where a 404 means the
+    parent resource being listed is gone rather than that a single lookup missed.
+    """
+    return error is not None and getattr(error, "status", None) == NOT_FOUND_STATUS_CODE
 
 
 def _is_retryable_okta_error(error: Any) -> bool:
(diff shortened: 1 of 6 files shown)


Jeff-Rowell (dismissed):

Jeff-Rowell (approved):

somethingnew2-0:
Thanks for fixing this!

PR #574, opened 2026-08-04
Fix sync-app-groups handing plugins an unloaded App, then crashing on the error path

barborico:
# Summary

A lifecycle plugin that raises took down the operation that fired it, not just its own side effect. Two shapes of that: `access sync-app-groups` handed plugins an `App` with no relationships loaded and then crashed on its own error path, and every other hook call site reads ORM state immediately after firing — which a plugin failure invalidates. Fixes both, plus the missing host-side test coverage that let the first one ship.

**Urgency**: 1 day
**Expected review effort**: MEDIUM (~30 min) — four commits, but only two to read: **commit 1** (`sync-app-groups`) and **commit 4** (the SAVEPOINT). Commits 2 and 3 are a change and its full revert, kept so the approach pivot is legible; skip them.

> **Approach pivot since the last review.** Commit 2 fixed the post-hook ORM reads one call site at a time. Review found three more of the same shape that it did not fix, two of them live 500s. Rather than keep enumerating call sites, commit 3 reverts that work and commit 4 fixes it once in the host with a SAVEPOINT. Every review comment on this PR was against commit 2 and is resolved by its revert. See "The pivot" below for the reasoning, including a correction to this PR's own earlier rejection of the SAVEPOINT idea.

# Motivation

Fix two Sentry issues firing together out of the staging `sync-app-groups` cronjob, 7 occurrences each, one pair per run. They are one cascade:

1. `_sync_all_app_groups` selected apps with a bare `select(App)` and passed each to the hook. Every `App` relationship is `lazy="raise_on_sql"`, so a plugin reading `app.active_app_groups` (the natural way to bulk reconcile) raised `InvalidRequestError` rather than querying. The Google example plugin does exactly that, so its periodic sync has never succeeded for any app.
2. The error handler rolled back and then formatted `app.name` into the failure message. A top-level rollback expires **every** instance in the identity map regardless of `expire_on_commit=False`, so that read became an implicit refresh needing sync IO on an `AsyncSession`: `MissingGreenlet`, unhandled, job dead.

That expiry also invalidated the apps still queued in the loop, so the `continue` intended to isolate one app's failure was already broken; app 2 would have re-raised regardless of app 1's outcome.

**Chasing that second point found the same defect at every other hook call site.** `invoke_app_group_lifecycle_hook` rolls the session back on a plugin failure, and callers read ORM state on the very next line:

- The **membership loops** in `ModifyGroupUsers` and `ModifyRoleGroups` fire once per affected group, so the per-group isolation they appear to provide was nominal — one bad plugin took down every group behind it, and the rest of the operation with it.
- **`CreateGroup`** re-reads `self.group.id` after firing `group_created`, to load the group for its audit log.
- **`PUT /api/groups/{id}`** reads `group.id` after its fire, to apply tags and re-load the response body — so one plugin's failure turned an otherwise-successful PUT into a 500.
- **`ModifyGroupType`** reads `group.id` right after firing `group_deleted`, and `group.type` after firing `group_created` (reachable from `CreateApp`, which leaves `fire_created_hook` at its default).
- **`ModifyGroupDetails`** reads `self.group.name` right after its own `group_updated` fire, and `routers/apps.py` calls it in a loop over an app's groups on rename — so `PUT /api/apps/{id}` 500s and leaves the app renamed with only some of its groups renamed.
- **`ModifyGroupUsers` returns** the group it operated on, so a caller reading that instance is the same post-hook read one frame out.

Reachable today for any operator running a lifecycle plugin.

# The pivot to savepoint

The last three bullets above are the ones commit 2 missed. That is the argument for fixing this in the host rather than at each call site: the list of sites is not closed, and a convention that has to be re-applied correctly at every future one will be got wrong again.

<details>

## Commit 1 — `sync-app-groups` (unchanged; previously approved)

**`api/cli.py`** — the batch loop no longer holds ORM instances across a transaction boundary. It scans plain column values (`App.id`, `App.name`, plugin id), keeping its own bookkeeping and error messages independent of session state, and loads each `App` inside the iteration with `selectinload(App.active_app_groups)`. Loading per iteration rather than once up front is what makes per-app isolation actually work: a preceding app's rollback expires the whole identity map, so a graph loaded before the loop would leave later apps' collections unloaded again. Apps are ordered by name so a run's sequence, and its output, is reproducible. The helper returns a failure count and the command exits non-zero if anything failed.

Note that only the collection needs an eager load. `AppGroup.app` is a many-to-one, and `raise_on_sql` checks the identity map before deciding SQL is required, so each group's `.app` resolves back to the `App` we already hold for free.

**`api/plugins/app_group_lifecycle.py`** — the `sync_all_groups` hookspec now states what arrives loaded and that anything wider must be eager-loaded by the caller rather than lazily read in the hook, mirroring the existing rule for the notification hooks.

**`tests/test_app_group_lifecycle_plugin.py`** — `DummyPlugin` had no `sync_all_groups` implementation at all, which is the direct reason this shipped: the only hook invoked exclusively by the CLI was the only one without host-side integration coverage, while the Google plugin's own unit tests assign `app.active_app_groups` on a `Mock(spec=App)` and so hand-feed the exact thing that was broken. `DummyPlugin` now reads the collection the way a real bulk-reconcile plugin does, with three tests covering the loaded graph, per-app failure isolation, and the no-plugin filter.

## Commits 2 and 3 — the per-call-site fix, and its revert

Nothing to review. Commit 2 is the previously-reviewed change; commit 3 reverts it whole, so the tree after commit 3 is byte-identical to the tree after commit 1. Kept rather than squashed so the approach and the reasons it was abandoned stay in history.

## Commit 4 — the SAVEPOINT

**`api/plugins/app_group_lifecycle.py`** — `invoke_app_group_lifecycle_hook` runs the hook inside `db.session.begin_nested()`. A plugin failure rolls back to the savepoint, which passes `dirty_only=True` to `_restore_snapshot` and expires only what the plugin modified, instead of `dirty_only=False` expiring the whole identity map. Callers keep reading their own instances across a fire; the entire class of defect in the Motivation section goes away at once, including the sites nobody enumerated.

Two edges are handled explicitly and commented at the call site:

- `begin_nested()`'s flush can raise on the **caller's own** pending state. Previously that surfaced as autoflush inside the hook and was caught by `run_hooks_to_completion`; it is caught here so it still cannot escape a function documented never to propagate.
- A hook that commits ends the savepoint itself, so `savepoint.is_active` is `False` and there is nothing to roll back to. That branch falls back to a session rollback — exactly the pre-savepoint behavior, cascading expiry and all — rather than silently skipping the rollback.

**`tests/test_app_group_lifecycle_plugin.py`** — ten tests, all of which fail without the change. Six carried over from commit 2 (the four membership cascade tests and the two single-fire tests); they are behavioral, so they pass here for a different reason. Four are new, one per call site commit 2 missed: `ModifyGroupType` on both its fires, the `PUT /api/apps/{id}` rename loop, and the operation return value.

Net effect on `api/` is 134 changed lines, against roughly 250 for the approach it replaces.

</details>

<details>

- Both original Sentry errors were reproduced by the new tests against unfixed code, matching the production traces frame for frame (including the crash site at the `app.name` read in the failure branch), and both are gone after the fix.
- **All ten isolation tests fail with commit 4's production change stashed and pass with it.** The four new ones were each first reproduced as a live failure: two 500s (`PUT /api/groups/{id}` converting away from an app group, `PUT /api/apps/{id}` rename) and two `MissingGreenlet`s out of the operation.
- The four membership-cascade tests inject the failure on group A **and** on group B. Iteration follows dict insertion order, which follows an unordered query, so failing only the group that happens to come second would let the regression pass by luck.
- Full backend suite green on **both backends**: 805 passed on SQLite and on PostgreSQL 15.6. `ruff check`, `ruff format --check`, and `ty check` clean.
- Verified against SQLAlchemy 2.0.50's source rather than assumed: that a top-level rollback expires unconditionally via `_restore_snapshot(dirty_only=False)`, that a *nested* rollback passes `dirty_only=transaction.nested`, that `_take_snapshot()` flushes for any non-`BEGIN`/`AUTOBEGIN` origin, and the identity-map shortcut that precedes the `raise_on_sql` check.
- Advisory-lock scope measured on real PostgreSQL via `pg_locks`: held inside the hook, released by `ROLLBACK TO SAVEPOINT`, transferred to the parent by `RELEASE`.
- Confirmed `SystemExit` propagates through the `asyncio.run` boundary in `_with_app_context` with the session teardown still running (exit code 1).

</details>

# Guidance for Reviewers

**Read commits 1 and 4; skip 2 and 3.** Commit 3 reverts commit 2 in full, so the pair is a no-op on the tree.

**One judgment call worth a look.** The non-zero exit in commit 1 is a deliberate behavior change. Fixing the crash would otherwise convert a loud failure into stderr output plus exit 0, leaving a periodic job reporting success after reconciling nothing. Every app is still attempted; only the final exit code changed.

### DIFF
--- a/api/operations/create_group.py
+++ b/api/operations/create_group.py
@@ -120,6 +120,11 @@ class CreateGroup:
                 )
             await db.session.commit()
 
+        # Hold the id as a plain string across the hook fire below. A failing plugin rolls the
+        # session back, and a top-level rollback expires every instance in the identity map, so
+        # re-reading `self.group.id` afterwards triggers a refresh and raises MissingGreenlet.
+        group_id = self.group.id
+
         # Invoke app group lifecycle plugin hook, if configured
         await invoke_app_group_lifecycle_hook(AppGroupLifecycleHook.GROUP_CREATED, group=self.group)
 
@@ -133,7 +138,7 @@ class CreateGroup:
                 select(OktaGroup)
                 .options(selectin_polymorphic(OktaGroup, [AppGroup, RoleGroup]), joinedload(AppGroup.app))
                 .where(OktaGroup.deleted_at.is_(None))
-                .where(OktaGroup.id == self.group.id)
+                .where(OktaGroup.id == group_id)
             )
         ).first()
 
--- a/api/operations/create_group.py
+++ b/api/operations/create_group.py
@@ -120,11 +120,6 @@ class CreateGroup:
                 )
             await db.session.commit()
 
-        # Hold the id as a plain string across the hook fire below. A failing plugin rolls the
-        # session back, and a top-level rollback expires every instance in the identity map, so
-        # re-reading `self.group.id` afterwards triggers a refresh and raises MissingGreenlet.
-        group_id = self.group.id
-
         # Invoke app group lifecycle plugin hook, if configured
         await invoke_app_group_lifecycle_hook(AppGroupLifecycleHook.GROUP_CREATED, group=self.group)
 
@@ -138,7 +133,7 @@ class CreateGroup:
                 select(OktaGroup)
                 .options(selectin_polymorphic(OktaGroup, [AppGroup, RoleGroup]), joinedload(AppGroup.app))
                 .where(OktaGroup.deleted_at.is_(None))
-                .where(OktaGroup.id == group_id)
+                .where(OktaGroup.id == self.group.id)
             )
         ).first()
 
--- a/api/operations/modify_group_users.py
+++ b/api/operations/modify_group_users.py
@@ -26,7 +26,7 @@ from api.models.access_request import get_all_possible_request_approvers
 from api.models.tag import coalesce_ended_at
 from api.operations.constraints import CheckForReason, CheckForSelfAdd
 from api.plugins import NotificationHook
-from api.plugins.app_group_lifecycle import AppGroupLifecycleHook, invoke_app_group_lifecycle_membership_hook
+from api.plugins.app_group_lifecycle import AppGroupLifecycleHook, invoke_app_group_lifecycle_hook
 from api.services import okta
 from api.schemas import AuditLogSchema, EventType
 
@@ -125,7 +125,7 @@ class ModifyGroupUsers:
                     await db.session.scalars(
                         select(OktaUserGroupMember)
                         .where(OktaUserGroupMember.id.in_(self.member_should_expire_ids))
-                        .where(OktaUserGroupMember.group_id == self.group_id)
+                        .where(OktaUserGroupMember.group_id == group.id)
                         .where(OktaUserGroupMember.ended_at > func.now())
                         .where(OktaUserGroupMember.is_owner.is_(False))
                     )
@@ -139,7 +139,7 @@ class ModifyGroupUsers:
                     await db.session.scalars(
                         select(OktaUserGroupMember)
                         .where(OktaUserGroupMember.id.in_(self.owner_should_expire_ids))
-                        .where(OktaUserGroupMember.group_id == self.group_id)
+                        .where(OktaUserGroupMember.group_id == group.id)
                         .where(OktaUserGroupMember.ended_at > func.now())
                         .where(OktaUserGroupMember.is_owner.is_(True))
                     )
@@ -250,12 +250,9 @@ class ModifyGroupUsers:
         remove_changed_members = members_to_remove + members_to_add
         remove_changed_owners = owners_to_add + owners_to_remove
 
-        # Track access overall lost and gained to pass along to app group lifecycle plugin hooks.
-        # Keyed by group id holding member ids rather than ORM instances: the hooks fire in a loop
-        # further down, and a failing plugin rolls the session back, which expires every instance in
-        # the identity map -- see invoke_app_group_lifecycle_membership_hook.
-        members_lost_by_group: dict[str, list[str]] = {}
-        members_gained_by_group: dict[str, list[str]] = {}
+        # Track access overall lost and gained to pass along to app group lifecycle plugin hooks
+        members_lost_by_group: dict[OktaGroup | RoleGroup | None, list[OktaUser]] = {}
+        members_gained_by_group: dict[OktaGroup | RoleGroup | None, list[OktaUser]] = {}
 
         # End group memberships and ownerships
         if len(remove_changed_members) > 0 or len(remove_changed_owners) > 0:
@@ -267,7 +264,7 @@ class ModifyGroupUsers:
                         OktaUserGroupMember.ended_at > func.now(),
                     )
                 )
-                .where(OktaUserGroupMember.group_id == self.group_id)
+                .where(OktaUserGroupMember.group_id == group.id)
                 .where(OktaUserGroupMember.is_owner.is_(False))
                 .where(OktaUserGroupMember.user_id.in_([m.id for m in remove_changed_members]))
                 .where(OktaUserGroupMember.role_group_map_id.is_(None))
@@ -285,7 +282,7 @@ class ModifyGroupUsers:
                         OktaUserGroupMember.ended_at > func.now(),
                     )
                 )
-                .where(OktaUserGroupMember.group_id == self.group_id)
+                .where(OktaUserGroupMember.group_id == group.id)
                 .where(OktaUserGroupMember.is_owner.is_(True))
                 .where(OktaUserGroupMember.user_id.in_([m.id for m in remove_changed_owners]))
                 .where(OktaUserGroupMember.role_group_map_id.is_(None))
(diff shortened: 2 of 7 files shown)


Jeff-Rowell (dismissed):

matthew-bass (dismissed):

api/operations/create_group.py
```diff
@@ -120,6 +120,11 @@ async def execute(self, *, _group: Optional[T] = None) -> T:
                 )
             await db.session.commit()
 
+        # Hold the id as a plain string across the hook fire below. A failing plugin rolls the
+        # session back, and a top-level rollback expires every instance in the identity map, so
```
somethingnew2-0: Should a failing plugin be able to affect the database session state for Access itself?

api/routers/groups.py
```diff
@@ -456,13 +461,13 @@ async def put_group(
         )
 
     await ModifyGroupTags(
-        group=group,
+        group=group_id,
```
somethingnew2-0: Is it generally safe to still pass in the OktaGroup object instead of the group_id into this ModifyGroupTags operation or just not safe in this place in the route?

api/operations/modify_group_users.py
```diff
@@ -125,7 +125,7 @@ async def execute(self) -> OktaGroup:
                     await db.session.scalars(
                         select(OktaUserGroupMember)
                         .where(OktaUserGroupMember.id.in_(self.member_should_expire_ids))
-                        .where(OktaUserGroupMember.group_id == group.id)
+                        .where(OktaUserGroupMember.group_id == self.group_id)
```
somethingnew2-0: This is a subtle operation coding anti-pattern. We want to validate the group passed into the constructor exists and is still active before using it in the operation. For this and all the other operations similarly changed in this PR we're now trusting unvalidated inputs in the operations execution method which they should be validated first.

somethingnew2-0 (changes requested):

api/operations/modify_group_users.py
```diff
@@ -783,6 +778,25 @@ async def execute(self) -> OktaGroup:
                     )
                 )
 
+        # Invoke app group lifecycle plugin hooks last, once every DB change is committed and every
+        # ORM read this operation needs is done. A failing plugin rolls the session back, which
+        # expires the whole identity map, so anything that reads ORM state after a hook fires would
+        # raise MissingGreenlet -- keeping the hooks at the tail means there is nothing left to
+        # poison. The helper re-loads each group and its members by id for the same reason, so one
+        # group's failing plugin cannot strand the groups queued behind it.
```
eguerrant: ModifyGroupDetails has this same loop shape and isn't fixed in this PR. It reads self.group.name at line 100, right after its fire at line 92, and apps.py:379 calls it in a loop over an app's groups on rename with fire_lifecycle_hook defaulting True. It commits before firing, so a failing plugin leaves the app renamed and only some of its groups renamed, then 500s out of the loop. Same defect as the two loops fixed here

api/routers/groups.py
```diff
@@ -445,6 +445,11 @@ async def put_group(
     # needs the pre-conversion AppGroup row, so it can't be deferred); the final type has no
     # lifecycle plugin, so nothing is fired here for that case.
     name_or_desc_changed = group.name != old_name or (group.description or "") != old_description
+    # Hold the id as a plain string across the hook fires below. A failing plugin rolls the session
+    # back, and a top-level rollback expires every instance in the identity map, so re-reading
+    # `group.id` afterwards triggers a refresh and raises MissingGreenlet — turning one plugin's
+    # failure into a 500 on an otherwise-successful PUT.
+    group_id = group.id
```
eguerrant: This same fix is missing at two other spots (both in modify_group_type.py). Line 123 fires group_deleted, then the next statement reads group.id at 134, reachable from PUT /api/groups/{id} converting away from app_group. Line 284 fires group_created, then line 287 reads group.type, suppressed from put_group by fire_created_hook=False but not from CreateApp

api/operations/create_group.py
barborico: Good call. Ideally no, and I've introduced a savepoint to codify this behavior and narrow the impact of rollbacks. This issue of plugin failure blast radius should be eliminated between the 4th commit on this PR and the isolation WIP in https://github.com/discord/access/pull/577.

api/operations/modify_group_users.py
barborico: This code is only reachable after the `self.group_id` has been validated, but that's a good pattern to enforce. I'll add a note to that effect in agent context in a follow on PR.

api/operations/modify_group_users.py
barborico: Good catch! I've changed the guard for this kind of bug by isolating rollbacks for plugin operations, rather than requiring each call site to be robust.

api/routers/groups.py
barborico: Just here. ModifyGroupTags accepts `OktaGroup | str` and the instance is fine anywhere no rollback can have intervened; the hazard is specifically "across a hook fire." But I'm removing that hazard.

api/operations/modify_group_users.py
barborico: Added memory in https://github.com/discord/access/pull/579

somethingnew2-0 (approved):

PR #577, opened 2026-08-06
Hand app-group-lifecycle plugins a capability context instead of a session

barborico:
# Summary

App-group-lifecycle plugins no longer receive a database session. They get an `AppGroupLifecycleContext` — a bounded capability surface — so a plugin integrates with Access only through the plugin interface. The Google example plugin's Access imports drop to `api.models.AppGroup` and `api.plugins.app_group_lifecycle`; it no longer imports `api.operations`, builds a query, takes a raw advisory lock, or commits.

**Urgency**: 3 days
**Expected review effort**: MEDIUM (~45 min)

> **#574 has merged**, so this now targets `main` directly. **#587 is stacked on this one** and adds the example Kubernetes CronJob for `sync-app-groups`; it can land right after.

# Motivation

Two open review comments on #532:

> Similar to my previous comments: plugins should not call into Access api methods directly and instead (if those methods are necessary) add them as apart of the plugin interface.

> In general, plugins should be fully self-contained and relying purely on the plugin interface for integration with Access. They should never tightly integrate with internal Access functionality. We should update the AGENTS.md and README to reflect this

The root cause is that the six async lifecycle hooks were handed the request `AsyncSession`. Given a session, a plugin will write SQL — and the Google plugin did: a hand-built cross-app query with a JSONB path predicate, a raw `pg_advisory_xact_lock` with backend sniffing, a mid-hook `session.commit()`, and a call into `api.operations`.

Growing the module-level helper list instead would have left that shape intact: 14 names a plugin imported individually, each needing `plugin_id` threaded in by hand, each implicitly bound to the ambient `db.session` ContextVar. A context object replaces the session parameter 1:1, binds `plugin_id` once, and gives the host one place to own transaction policy.

Breaking, and intentionally so — it ships with the Access 2.0 async hookspecs.

<details>

**`AppGroupLifecycleContext`** (`api/plugins/app_group_lifecycle.py`) — 13 capabilities on one class, replacing 4 loose accessors plus 4 loose coroutines. Bound to the invoking plugin's id, so plugin code never passes `plugin_id` and cannot reach another plugin's `plugin_data` namespace. `lock` generalizes the plugin's advisory lock; `find_groups_by_status` generalizes its ownership query, keeping the JSON path in SQL so it stays a point lookup; `set_group_description` replaces the `api.operations` call; the four Okta group-push calls mean a plugin never imports `api.services.okta`.

**`list_group_members` and `find_user_groups`** (added in review) — the membership reads a plugin needs when it provisions users itself rather than delegating to Okta group push. Neither is answerable from what a hook is handed: `sync_group` gets a group with only `.app` eager-loaded, and `group_members_removed` gets the removed delta, not the membership that remains. Where the external system attaches one combined permission set per user (Google Play Console, App Store Connect), removal has to write the union recomputed over the user's *remaining* groups. Same shape as `find_groups_by_status` — predicate in SQL, no session and no query in plugin code. `find_user_groups` scopes itself to this plugin's apps, since it names no group; `list_group_members` takes one, which is the only argument on the surface that could reach past the plugin id the context is bound to, so it re-establishes that binding and raises `ValueError` for a group whose app names a different lifecycle plugin or none. Membership is not namespaced per plugin the way `plugin_data` is, so nothing else would hold that line. Both go through one `_active_membership` helper that matches on one side with `IN` rather than joining: a user can hold the same group both directly and through a role, so the join multiplies rows, and `SELECT DISTINCT` over the entity would compare `AppGroup.plugin_data` / `OktaUser.profile`, which are `JSON` rather than `JSONB` off Postgres and have no equality operator.

Not added: the by-email `find_user_groups` variant, since `group_members_removed` already receives `OktaUser` objects and `sync_group` can reach users through `list_group_members`. Happy to add it if a plugin needs to map an identifier out of its own status back onto a user.

`lock` is a plain `await`, not an async context manager: `pg_advisory_xact_lock` releases only at transaction end and the transaction belongs to the host, so a `with` block would advertise a scope it cannot honor — in the one place where misreading the lifetime is a correctness bug.

**Host transaction policy** — `invoke_app_group_lifecycle_hook` builds the context, commits on success, and returns `list[BaseException]` so a batch caller can count failures. On failure it rolls back and then re-applies the status writes the plugin marked `durable_on_failure`. That flag is the subtle part; see Guidance below.

**`ModifyGroupDetails` gains `commit_db_changes: bool = True`** — `ctx.set_group_description` reuses the operation (so the ORM update, Okta push, and name-collision checks stay consistent with the API's own edit path) with both re-entrancy hazards disabled. The default keeps the two request-path callers byte-identical.

**`sync_all_groups(app)` → `sync_group(group)`**, and `_sync_all_app_groups` in `api/cli.py` iterates groups rather than apps, calling `invoke_app_group_lifecycle_hook` per group instead of hand-rolling `run_hooks_to_completion`. Plugins stop implementing their own batch loop and error isolation.

**Both example plugins migrated.** The audit logger is the smaller reference (config reads, status writes on both `App` and `AppGroup`). The Google plugin loses `_lock_claim`, `_get_google_group_owner`, its `session.add`/`commit`, and its batch loop.

**Comment and docstring pass (added in review)** — the interface no longer narrates its own history (no Access 1.x framing, no comparison to shapes a reader can't see), and every capability documents its parameters, return value, and exceptions the same way the module functions and hookspecs already did. Two comments were stale rather than merely retrospective: the Google plugin's `_reconcile` claimed to commit sync status inside the hook, which a plugin can no longer do, and `run_hooks_to_completion` justified `asyncio.wait` only for post-commit network-I/O hooks, which is no longer its only caller. Click command docstrings are untouched — they are `--help` text.

**Docs** — `.claude/CLAUDE.md` (the plugin-isolation principle, kept general; per-capability contract lives in docstrings), `.claude/access-dev-sync.md` (a section for the `sync-app-groups` cronjob and why its transaction boundary is per-group), `README.md` (the plugin-author contract). Peter asked for AGENTS.md; the repo has none and agent context lives under `.claude/`, so the rule went there plus the public README rather than starting a parallel doc tree.

</details>

<details>

- Full suite green at **every commit**: 829 passed (811 at the original commits, plus 13 new membership-query tests and 5 picked up from `main`). `ruff check`, `ruff format --check`, `ty check` clean throughout.
- **The isolation is asserted, not eyeballed.** `test_plugin_only_imports_the_access_plugin_interface` parses `plugin.py`'s imports and pins the boundary. Verified it fails — naming the offending module — when an `api.operations` import is reintroduced.
- **The `durable_on_failure` split is pinned by a test that fails against a host replaying every recorded write**, verified by making the host do exactly that: the ownership token persists and the assertion catches it.
- Both example plugins register against the real hookspec and pass `verify_async_impls`, confirming pluggy's argument validation accepts the new signatures and the renamed `sync_group`.
- **13 new tests for the membership reads**, covering owner exclusion, both temporal filters, soft-deleted users/apps/groups, cross-plugin scoping, the `app=` narrowing, `.app` loaded on a **cold** session, and `list_group_members` refusing both a foreign plugin's group and an app with no plugin configured. Mutation-checked: swapping the `IN` subquery for a plain join breaks 3 of them, so the dedupe and `ended_at` assertions bite rather than passing incidentally.
- New coverage for `find_groups_by_status` (spans apps sharing a plugin id, excludes soft-deleted rows, honors `exclude_group`/`limit`, and a returned group's `.app` is readable **on a cold session** — a warm one hides a missing eager-load, since many-to-one resolves from the identity map without SQL), `ctx.lock` on both backends, the durable replay against a vanished row, and `GROUP_DELETED` skipping the replay.
- `TestSyncGroupHook` is per-group, including that a failing group's *siblings* still sync.
- I manually did some E2E smoke testing (group create, rename, delete) to sanity check.
- **Not done:** the end-to-end run against real Postgres + Google + Okta (`docker compose up`, configure an app with the plugin, `access sync-app-groups`, confirm `SYNC_ERROR` persists across a rolled-back hook). Needs credentials I don't have. Everything covering that path is unit/integration-level.

</details>

# Guidance for Reviewers

<details>
<summary>There's some churn between commits, so going file-by-file may be most efficient. I recommend hiding whitespace. Also, a lot of the diff is comments and docstrings. Here are the main things to focus on:</summary>

- The CLI command to sync downstream groups in `api/cli.py`: Refactor for better failure isolation by processing groups individually rather than all together.
- The plugin interface in `api/plugins/app_group_lifecycle.py`: This is essentially a big refactor to enforce the boundary between plugin implementations and the rest of Access--particularly the DB. The most complex bits are ensuring that DB operations are rolled back on failure, but that failure messages are persisted to show up in the UI.
  > **The one design decision that most merits scrutiny is `durable_on_failure` on `set_status`.** A plugin's sync status must outlive its own failure or an operator sees a group stuck with no explanation — but an *ownership token* must not. The token is only sound when committed in the transaction that held the `lock` during the ownership check; replaying it in a fresh transaction after the lock released would let a second Access group, which has meanwhile passed its own check, claim the same Google group. Both then record it, and deleting either would delete the shared one. That is why the flag exists rather than the host replaying everything.
- The plugin implementations in `examples/plugins/*`: Refactor the plugins to use the new interface. Functionally the same.
- Tests: the logic should all be essentially the same, just adapted to the new class and method interfaces.
</details>

<details>
<summary>Alternatively, review commit-by-commit to understand logically grouped changes; each commit is green and self-contained.</summary>

**Two structural constraints worth knowing, because they shaped the commit sequence:**

1. Pluggy validates hookimpl arguments against the spec at *registration*, so renaming `session` → `ctx` outright breaks both example plugins at plugin-load. An impl may declare a *subset* of spec args, so `ctx` was added alongside a deprecated `session`, the plugins migrated one at a time, and `session` came off in commit 4. Nothing transitional escapes the PR.
2. Privatizing the superseded module functions had to wait until no plugin imported them, or the commit would have shipped a red suite.

</details>


.claude/access-dev-sync.md
```diff
@@ -16,6 +16,23 @@ and whether a group `is_managed` (`act_authoritatively = act_as_authority and is
 - **Unmanaged group, or `act_as_authority=False`**: Okta is authoritative. Changes in Okta
   are reflected into the DB; DB-only records are removed.
 
+## App group lifecycle sync (`access sync-app-groups`)
+
+A separate cronjob from `syncer.py`, driving the app-group-lifecycle plugin hooks rather than Okta
```
somethingnew2-0, line 21: Let's add an example Kubernetes deployment of this cronjob in https://github.com/discord/access/blob/main/examples/kubernetes

somethingnew2-0:
The capability boundary is the right call, and `durable_on_failure` looks correct to me — a sync status must outlive the failure, an ownership token must not, and I don't think a host that replayed everything could be made safe. No argument there.

One gap I'd like to resolve before this lands, because it blocks two plugins I'm writing: **the context has no way to read group membership.**

The surface is `plugin_id`, `lock`, `find_groups_by_status`, `get`/`set_config`, `get`/`set_status`, `set_group_description`, and the four push-mapping calls. Nothing answers either of:

1. **Who are this group's current members?** Needed by `sync_group` to converge. The hookspec says `group.app` is eager-loaded and every other relationship is `lazy="raise_on_sql"`, and `_sync_all_app_groups` loads only `.app` — so `group.active_user_memberships` raises, and there's no `ctx` verb to go get it.
2. **Which other groups is this user in, among the apps using this plugin?** Needed by `group_members_removed`.

`find_groups_by_status` doesn't cover either: it matches one status property against one exact string, so it can't express "groups this user is a member of."

### Why (2) matters, and why it isn't just a `sync_group` eager-load

The example plugin doesn't hit this because it delegates membership to Okta group push, so it never reads Access membership and implements no `group_members_added`/`group_members_removed` at all. That works whenever the external system is an Okta push target.

It breaks when the external system provisions **individual users with a single per-user permission set** and its API is the only integration surface. Both plugins I'm working on are that shape:

- **Google Play Console** (Android Publisher `users` API): one user resource per email, one permission set on it.
- **App Store Connect**: roles are tenant-global, attached to the user, not to any group.

In both, a user in two Access groups must hold the **union** of both groups' grants. So on removal from one group the correct action is to recompute the union from their *remaining* groups and write the reduced set — deleting the user, or blindly writing the departing group's grants, silently revokes access they still legitimately hold via the other group.

That recompute is a cross-group read from inside `group_members_removed`. `members` gives me the delta, not the remaining set. And because that's a request-path hook, the "have the caller eager-load it" escape hatch in the `sync_group` docstring doesn't apply — the caller isn't `_sync_all_app_groups`.

### Concretely

Two additions would close it:

```python
async def list_group_members(self, group: AppGroup) -> list[OktaUser]:
    """Active, non-owner members of an app group."""

async def find_user_groups(self, user: OktaUser, *, app: App | None = None) -> list[AppGroup]:
    """Active app groups the user is a member of, scoped to apps configured with this
    plugin id (optionally to one app)."""
```

Both are the same shape as `find_groups_by_status` — predicate stays in SQL, plugin-id scoping implicit, plugin still never sees a session or builds a query. `find_user_groups` is the one that actually unblocks the union case; `list_group_members` just spares every membership-provisioning plugin a CLI-side eager-load. (A by-email variant of `find_user_groups` would also let a plugin map identifiers it recorded in its own status back onto users, but that's minor next to the two above.)

If you'd rather not grow the surface here, the alternative I can live with is eager-loading members for `sync_group` plus a follow-up for the cross-group read — but a member-provisioning plugin can't correctly implement `group_members_removed` in the meantime, so I'd want that follow-up tracked before this ships.

🤖 Posted by Claude Code

api/plugins/app_group_lifecycle.py
```diff
@@ -116,6 +118,372 @@ class AppGroupLifecyclePluginData:
...
+    entity_type: Literal["app", "group"]
+    entity_id: str
+    property_name: str
+    value: Any
+
+
+class AppGroupLifecycleContext:
+    """The capability surface an app group lifecycle hook may use to talk to Access.
+
+    Replaces the raw ``AsyncSession`` the Access 1.x lifecycle hooks received. A plugin gets exactly
```
somethingnew2-0: Access 1.x never used `async` or AsyncSession, so we can remove this part of the comment. Let's update the comments to only describe current intent and behavior, not prior behavior.

somethingnew2-0:
Much cleaner plugin interface for self-contained plugins, thanks for refactoring this!

Left a few comments, the important one being the suggestion for adding query methods for app group plugins that sync membership themselves (instead of via Okta push group mappings) and need to provide the union of permissions for an individual across multiple app groups in order to sync those memberships.

.claude/access-dev-sync.md
barborico: Added in #587, stacked on this one: `examples/kubernetes/cron-job-sync-app-groups.yaml`.

api/plugins/app_group_lifecycle.py
barborico: Removed, and taken as a general pass over the plugin's comments and docstrings in ad4d97b.

barborico:
> The capability boundary is the right call, and `durable_on_failure` looks correct to me — a sync status must outlive the failure, an ownership token must not, and I don't think a host that replayed everything could be made safe. No argument there.
> 
> One gap I'd like to resolve before this lands, because it blocks two plugins I'm writing: **the context has no way to read group membership.**
> 
> The surface is `plugin_id`, `lock`, `find_groups_by_status`, `get`/`set_config`, `get`/`set_status`, `set_group_description`, and the four push-mapping calls. Nothing answers either of:
> 
> 1. **Who are this group's current members?** Needed by `sync_group` to converge. The hookspec says `group.app` is eager-loaded and every other relationship is `lazy="raise_on_sql"`, and `_sync_all_app_groups` loads only `.app` — so `group.active_user_memberships` raises, and there's no `ctx` verb to go get it.
> 2. **Which other groups is this user in, among the apps using this plugin?** Needed by `group_members_removed`.
> 
> `find_groups_by_status` doesn't cover either: it matches one status property against one exact string, so it can't express "groups this user is a member of."
> 
> ### Why (2) matters, and why it isn't just a `sync_group` eager-load
> The example plugin doesn't hit this because it delegates membership to Okta group push, so it never reads Access membership and implements no `group_members_added`/`group_members_removed` at all. That works whenever the external system is an Okta push target.
> 
> It breaks when the external system provisions **individual users with a single per-user permission set** and its API is the only integration surface. Both plugins I'm working on are that shape:
> 
> * **Google Play Console** (Android Publisher `users` API): one user resource per email, one permission set on it.
> * **App Store Connect**: roles are tenant-global, attached to the user, not to any group.
> 
> In both, a user in two Access groups must hold the **union** of both groups' grants. So on removal from one group the correct action is to recompute the union from their _remaining_ groups and write the reduced set — deleting the user, or blindly writing the departing group's grants, silently revokes access they still legitimately hold via the other group.
> 
> That recompute is a cross-group read from inside `group_members_removed`. `members` gives me the delta, not the remaining set. And because that's a request-path hook, the "have the caller eager-load it" escape hatch in the `sync_group` docstring doesn't apply — the caller isn't `_sync_all_app_groups`.
> 
> ### Concretely
> Two additions would close it:
> 
> ```python
> async def list_group_members(self, group: AppGroup) -> list[OktaUser]:
>     """Active, non-owner members of an app group."""
> 
> async def find_user_groups(self, user: OktaUser, *, app: App | None = None) -> list[AppGroup]:
>     """Active app groups the user is a member of, scoped to apps configured with this
>     plugin id (optionally to one app)."""
> ```
> 
> Both are the same shape as `find_groups_by_status` — predicate stays in SQL, plugin-id scoping implicit, plugin still never sees a session or builds a query. `find_user_groups` is the one that actually unblocks the union case; `list_group_members` just spares every membership-provisioning plugin a CLI-side eager-load. (A by-email variant of `find_user_groups` would also let a plugin map identifiers it recorded in its own status back onto users, but that's minor next to the two above.)
> 
> If you'd rather not grow the surface here, the alternative I can live with is eager-loading members for `sync_group` plus a follow-up for the cross-group read — but a member-provisioning plugin can't correctly implement `group_members_removed` in the meantime, so I'd want that follow-up tracked before this ships.
> 
> 🤖 Posted by Claude Code

Both added in d4ced11, with the signatures you specified.

Two deltas worth flagging: `list_group_members` raises `ValueError` if the group's app isn't configured with the calling plugin, and I skipped the by-email variant since nothing's blocked on it. Happy to add the latter if you want it.

somethingnew2-0:
`list_group_members` and `find_user_groups` cover what I asked for — thanks. The ownership guard on `list_group_members` and the `IN`-subquery dedup for role-held membership are both better than what I sketched.

One residual case, narrower than the first: **there's no way to get from an identifier a plugin stored to the `OktaUser` that `find_user_groups` takes.**

Where that bites is `group_deleted`. `delete_group` end-dates every membership before invoking the hook, so `list_group_members(group)` returns empty there — which is why a plugin that has to clean up after itself keeps its own record of who it provisioned in group status. That record is emails (or ids), and `find_user_groups` takes an `OktaUser`, so the union recompute that works fine in `group_members_removed` can't be done in `group_deleted`.

Without it the choices are to delete the user outright, which drops roles another group still grants, or to infer the remaining set from the external system's own state, which is wrong as soon as two groups grant overlapping roles. Both are worse than the over-broad revocation this interface is otherwise good at preventing.

Either shape would close it:

```python
async def find_users_by_email(self, emails: Sequence[str]) -> list[OktaUser]: ...
```

or `find_user_groups` accepting an email alongside an `OktaUser`.

I called this minor when I raised it, which was right for `sync_group` and `group_members_removed` and wrong for `group_deleted` — that hook can't be implemented correctly without it. Fine by me as a follow-up rather than more scope here, as long as it's tracked.

🤖 Posted by Claude Code

barborico:
> `list_group_members` and `find_user_groups` cover what I asked for — thanks. The ownership guard on `list_group_members` and the `IN`-subquery dedup for role-held membership are both better than what I sketched.
> 
> One residual case, narrower than the first: **there's no way to get from an identifier a plugin stored to the `OktaUser` that `find_user_groups` takes.**
> 
> Where that bites is `group_deleted`. `delete_group` end-dates every membership before invoking the hook, so `list_group_members(group)` returns empty there — which is why a plugin that has to clean up after itself keeps its own record of who it provisioned in group status. That record is emails (or ids), and `find_user_groups` takes an `OktaUser`, so the union recompute that works fine in `group_members_removed` can't be done in `group_deleted`.
> 
> Without it the choices are to delete the user outright, which drops roles another group still grants, or to infer the remaining set from the external system's own state, which is wrong as soon as two groups grant overlapping roles. Both are worse than the over-broad revocation this interface is otherwise good at preventing.
> 
> Either shape would close it:
> 
> ```python
> async def find_users_by_email(self, emails: Sequence[str]) -> list[OktaUser]: ...
> ```
> 
> or `find_user_groups` accepting an email alongside an `OktaUser`.
> 
> I called this minor when I raised it, which was right for `sync_group` and `group_members_removed` and wrong for `group_deleted` — that hook can't be implemented correctly without it. Fine by me as a follow-up rather than more scope here, as long as it's tracked.
> 
> 🤖 Posted by Claude Code

This feels like a bit of an XY problem; it seems cleaner to have Access pass `members` as a new parameter to the `group_deleted` hook, so that it's not necessary to also serialize that as plugin state and then perform a lookup. I added that in a stacked PR: https://github.com/discord/access/pull/588

somethingnew2-0 (approved):

PR #585, opened 2026-08-12
Run app-group-lifecycle plugin hooks after the response

somethingnew2-0:
# Summary

App-group-lifecycle plugin hooks no longer run on the request path. A fire is recorded as plain data and replayed from a post-response `BackgroundTask`, so a group create, rename, or membership change commits and answers the request without waiting on the plugin's reconciliation against an external system.

> **Rebased onto `main`** now that #577 and #588 have merged; no longer stacked. The rebase onto #588 surfaced a real bug in the deferral — an empty `members` list collapsed to "no members argument", which would have failed `group_deleted` for any group nobody was in. Fixed in its own commit at the end, with the DeleteGroup interaction #588 cares about pinned by a test.

# Motivation

These hooks were the slowest thing left on the request path. `group_created` on a plugin-managed app group does Okta plus whatever the plugin talks to (for the Google reference plugin: a lookup, a push-mapping create, and a wait for Okta to import), all before the user gets a response. Notifications and Okta calls were moved off the response path in #481; lifecycle hooks were the remaining holdout, and the reason they were is that they *write*, so the detached-snapshot trick notifications use does not apply to them.

#577 is what makes this tractable rather than an ecosystem migration. Before it, hooks received the raw `AsyncSession` and plugins used it directly, so deferring meant every operator's plugin had to survive running on a different session. After it, the only session reference on this path is inside `invoke_app_group_lifecycle_hook`, and the change is a host concern.

# Design notes

**A deferred fire carries ids, not objects.** This deliberately does not reuse `_fan_out.py`: its primitive spawns with `create_task` at defer time, which would put a second coroutine on the live request session; and one request's fires are order-dependent. Replay re-loads the group by id, which is better than a snapshot would have been — a hook sees committed truth, and a group deleted (or converted out of being an app group) in the meantime is skipped rather than reconciled against a row that is no longer there.

**The drain binds `_session_scope`, not just a session.** `ctx.set_group_description` delegates to `ModifyGroupDetails`, which reaches for the ambient `db.session`. If that resolved to a different session than the one the context holds, its writes would land in a transaction nobody commits — silently persisting nothing. There is a test whose only job is to catch that.

**`group_deleted` stays on the request, at both of its fire sites.** Deferral trades the chance of losing a fire against a faster response, and nothing comes back for a delete: `sync_group` sweeps live app groups, so a soft-deleted one is invisible to it, and a dropped delete would leave the external group and its members alive while Access shows the access as revoked. That is not a latency cost, and deleting a group is rare and nobody is waiting on it. `ModifyGroupType` also has a structural reason — it hard-deletes the `app_group` row, so a replay would find no app group at all. `defer_or_invoke_lifecycle_hook` routes `GROUP_DELETED` inline itself rather than trusting call sites to remember.

**Ordering is unchanged.** The lifecycle drain is registered ahead of the Okta/notification drain. Those Okta tasks are already in flight — operations spawn them with `create_task` and only the await was ever deferred — so hooks keep racing the pushes rather than newly trailing them. Making them trail is arguably more correct and is a separate question.

**A fire can be lost, and what that costs is the plugin's property, not the host's.** A worker killed between response and drain drops one, as does a request that fails after queuing one. A plugin whose `sync_group` is a full idempotent reconciliation re-converges on the next `sync-app-groups` run — that is what the Google plugin does, and it is the shape the interface asks for. But `sync_group` is optional and Access does not police what it does, so a plugin that omits it or implements it as something else does not recover; the audit-logger example in this repo is exactly that case. Two gaps close for nobody: the sweep filters `AppGroup.deleted_at.is_(None)`, so a dropped `group_deleted` is unrecoverable, and a sweep sees only current membership, so a delta-based revoke cannot recover a lost `group_members_removed`.

**Deferral is therefore a semantic change to the plugin interface, not a scheduling one.** Hooks observe re-loaded state, are unordered across requests, and can be lost — so the delivery contract now lives on `AppGroupLifecyclePluginSpec`, with short caveats on the delta arguments themselves and a matching entry in the README's plugin-author list.

# The plugin interface addition, and why it's here

Deferring means a create/update response no longer carries status the hook writes, so the group page has to poll. Gating that on "pending" would have hardcoded the Google plugin's vocabulary into a client every operator shares — Access has no opinion on what a plugin's status values mean.

So `AppGroupLifecyclePluginStatusProperty` gains `pending_values`: the plugin names the values that mean "still converging", and the group page refreshes while a status sits on one. It is additive and defaults to opting out. Landing it now rather than as a follow-up because 2.0 is the window for interface changes, and because the polling it enables also fixes a case that predates this PR — a group parked in the Google plugin's "awaiting Okta import" state never updated without a manual reload.

The client also polls when a plugin that declares `pending_values` has written no status at all, which is what a freshly created group looks like before its first hook lands. Both signals are derived from data rather than navigation state, since two of the three post-submit paths do a full page reload.

# Second round — changes from @barborico's review

The review's core point was that several safety arguments were properties of the Google plugin stated as properties of the host. That was right, and four commits address it:

- **`54d8c68`** — the durability fix. `ModifyRoleGroups` fired `GROUP_MEMBERS_ADDED`/`REMOVED` mid-loop, with the commit that persisted those rows further down the function; both now accumulate per group and fire after their commit, as `ModifyGroupUsers` already did. Inline that was survivable — the hook's own commit flushed the caller's pending state — so deferral is what turned it into a gap. `defer_fan_out`'s error branch now *drops* queued fires instead of replaying them: replay re-loads the group but carries `members` verbatim, so a rolled-back removal still arrived as fact, and a plugin acting on it deprovisions someone who still has access.
- **`f0336ec`** — the delivery contract on the hookspec and in the README, including `ctx.lock` as the answer for plugins needing mutual exclusion across concurrent drains.
- **`c33da15`** — the Google plugin records a terminal `skipped` state at its three "nothing to do" returns. An empty status is indistinguishable from a hook that has not run, so a group predating plugin enablement was polling for the full window on every page view.
- **`5b97f4b`** — the safety-net claim scoped back here, in `.claude/access-dev-sync.md`, and in the module docstring.
- **`c5fd33d`** — and then the conclusion that follows from it: `group_deleted` is no longer deferred at all. Once "a lost fire costs latency" stops being true for a hook, deferring it is not a trade worth making. Both fire sites run inline, the deferring helper enforces it, and the docs from `5b97f4b` say so.

One correction worth recording: point 4 cited `modify_group_users.py:474`, but that site commits at :473 before firing at :477. The genuinely fire-before-commit sites were `modify_role_groups.py:327` and `:470`, whose commits land at :373 and :545 — which is what `54d8c68` fixes.

# Notes for reviewers

- The most consequential single line is the scope binding in `run_deferred_lifecycle`; the accompanying comment explains what breaks without it.
- The generated API client is regenerated in the same commit as the schema change, which the `openapi-client-drift` job requires. I learned that the hard way — the first push failed on it — so `.claude/access-dev-testing.md` now documents the gate and the offline regeneration CI uses, since `npm run codegen` otherwise wants a dev backend on 6060.
- The group-page poll is driven from state rather than a ref. The ref version looked right and did nothing: react-query re-reads `refetchInterval` when its options change, and a ref mutation does not re-render. It took watching the requests against a running backend to catch, and there is no page-level test infrastructure here that would have.
- Commit-by-commit works well here: commit 3 is the only one that changes behavior.


src/components/pluginStatus.ts
barborico: Taking just the example of the Google plugin, there are a couple cases where this will poll for a while without success:
1. Groups without plugin config which therefore cannot be reconciled will have unset status until the user adds that config.
2. Groups that are pending reconciliation via the periodic sync will have a pending status for up to the sync period.

src/components/pluginStatus.ts
```diff
@@ -0,0 +1,57 @@
...
+ * True when a plugin-managed entity looks like it is mid-reconcile.
+ *
+ * Two ways that happens:
+ *  - a status property is sitting on one of the values its plugin declared as in-progress; or
+ *  - the plugin declares in-progress values but has written no status at all yet, which is what a
+ *    just-created group looks like before its first hook lands.
+ *
+ * A plugin that declares no `pending_values` anywhere has opted out, and never polls.
+ */
+export function isPluginStatusPending(
```
barborico, line 36: This treats "plugin declares `pending_values` but has written no status at all" as pending, since that's what a freshly created group looks like before its first hook lands. But it's also what a group looks like after a hook ran and deliberately declined to write, which happens in a few cases for the Google plugin:
- plugin disabled for the app — plugin.py:536 (PR-branch line numbers; main's are ~6 lower)
- no email config, nothing to adopt — plugin.py:594-596, the Case 4 bail: if not configured_email_prefix or not configured_email: logger.info("Skipping … missing required config."); return
- no resolvable email when linking an adopted group — plugin.py:647-649, same bare return
So an app group that predates plugin enablement (no `google_group_email` config, since only the create path enforces it) polls every 5s for the full 60s window on every page view, not just after a mutation.

So if we go this route, we probably need to set a clear expectation that pending values are always written on reconcile for this type of plugin. And add a new sync status (e.g. SKIPPED) for the Google implementation.

.claude/access-dev-sync.md
```diff
@@ -37,6 +37,12 @@ deliberate, not incidental: a rolled-back group expires the entire identity map,
 across an iteration boundary is unusable — the same hazard described under `lazy="raise_on_sql"` in
 `.claude/CLAUDE.md`.
 
+This cronjob is also what makes the request path's post-response lifecycle drain
+(`api/operations/_lifecycle_fan_out.py`) safe to lose. A pod killed between a response and its drain
+drops that fire; because `sync_group` is per-group and idempotent, the next run re-converges it. The
+cost of a dropped fire is reconciliation latency, not divergence — so if this job is disabled or
+failing, that safety net is gone.
```
barborico: I think this description of the safety net (also given in the PR description) is too strong: It states a host-level guarantee that only the reference plugin provides.

`sync_group` is optional: `verify_async_impls` only checks that implemented hooks are `async def`, so a plugin omitting it loads clean and `_sync_all_app_groups` prints ✓ for every group and exits 0. And nothing in the hookspec asks `sync_group` to subsume the event hooks; the Google plugin routes it to the same `_reconcile()`, which is where "re-converges it" actually comes from. For example, our GitHub plugin doesn't implement this hook at all.

Deletes can't be re-converged by any plugin: the sweep filters `AppGroup.deleted_at.is_(None)`, so a soft-deleted group is structurally invisible to it. A dropped `group_deleted` leaves the external group and its members alive with nothing scanning for them. (Pre-existing — an inline hook that raised was already unrecoverable — but this paragraph is what claims the net covers it.)

Same limit for membership deltas: `group_members_removed` carries the delta, while
`sync_group` sees only current membership, so a plugin whose external system takes revoke calls rather than a desired-state write can't recover a lost removal either.

At a minimum, let's update the documentation to clarify the net covers what a plugin's own `sync_group` chooses to reconcile, and that the hookspec make idempotent full reconciliation an explicit expectation.

barborico (commented):
Overall, I think the mechanism is sound and the hard parts are handled well, e.g. not reusing `_fan_out.py`, `allow_deleted` for the one hook that must see a soft-deleted row, `_snapshot_kwargs` failing open to inline, and the `_session_scope` binding.

My only concern is that several safety arguments are properties of the Google plugin expressed as properties of the host. That plugin is a full-state idempotent reconciler, the one shape for which deferral is free, and it implements no membership hooks at all; so the machinery this PR most changes is exercised by tests and by the audit-logger example, but not by the plugin the design was validated against. Four places that shows up:

1. **"Losing a fire costs latency, not correctness" is false for some implementations, such as the other example plugin in this repo.** `app_group_lifecycle_audit_logger` records events: every hook calls `_log(...)` and `_increment_event_count(...)`, and its `sync_group` only logs that a sweep happened. A dropped fire is a permanently missing audit line and a wrong count. Stated unconditionally at `_lifecycle_fan_out.py:29` and `.claude/access-dev-sync.md:40`.

2. **`group_updated`'s before/after pair may describe a transition that never happened.** `old_name` is snapshotted; the group is re-loaded at replay. Two renames landing close together from different requests give the first drain `old_name` from before update 1 and `group.name` from after update 2. Inline execution couldn't produce that, because the hook ran inside the transaction that made the change.

3. **Nothing orders fires across requests.** `:224` and `test_replays_in_the_order_they_were_fired` cover ordering within a request; each request has its own `BackgroundTask`, and two drains race. "Add X" and "remove X" from two requests can replay reversed.

4. **Fires are recorded before the commit that makes them true, and the error path replays them anyway.** `modify_group_users.py:474` defers `GROUP_MEMBERS_REMOVED` and then commits; if the endpoint raises in between, `defer_fan_out`'s `except` branch drains regardless and the plugin is told about a removal that did not happen. Suggest clearing the lifecycle collector on that branch rather than replaying it.

2 and 3 are the same shape: the hookspecs hand plugins deltas, which invites delta-shaped implementations, while deferral is only sound for state-shaped ones. That is a semantic change to the plugin interface rather than a scheduling change, and worth stating outright to clarify the responsibilities of individual implementations. Most of 1 to 4 collapse into one missing contract, which belongs in the hookspec rather than only in the module docstring:

> A lifecycle hook may run after the response has been sent, against a group re-loaded at run time, and is not ordered relative to fires from other requests. Implement hooks as idempotent reconciliations of current state rather than applications of the delta they are handed; treat delta arguments as a hint about what changed, not as an instruction. Delivery is not guaranteed: `sync_group` is the recovery path, and a plugin that cannot converge without the events (e.g. an audit or notification plugin) should expect gaps.

Delete fragility may merit an additional callout, unless we change the shape of the "sync app groups" command to handle soft-deleted groups; see the inline comment on `.claude/access-dev-sync.md`.

api/operations/modify_role_groups.py
```diff
@@ -544,6 +553,14 @@ async def execute(self) -> RoleGroup:
             # Commit changes so far, so we can reference OktaUserGroupMember in approved AccessRequests
             await db.session.commit()
 
+            # Now that the additions are durable, tell the lifecycle plugin about them.
+            for affected_group, gained_members in members_gained_by_group.items():
```
somethingnew2-0, line 557: Is there ever a case where a member is both in members_lost_by_group and members_gained_by_group? What is the plugin behavior then? Since the plugins are deferred and running concurrently would this lead to inconsistent calling of the plugin hooks?

README.md
somethingnew2-0, line 1: Should this change also be added to the https://github.com/discord/access/blob/main/.claude/CLAUDE.md or similar agent specific documentation?

api/operations/modify_role_groups.py
somethingnew2-0: Yes, it's reachable — `ModifyRoleGroups` takes `groups_to_add` and `groups_to_remove` independently and doesn't dedupe them, so naming the same group in both puts the same user in both dicts.

It doesn't produce inconsistent hook calls *within* a request, though. The two fires are collected in the order the DB applies them — removals during the removal pass, fired after its commit; additions after theirs — and `run_deferred_lifecycle` replays them sequentially, one at a time, in collection order. So the plugin sees `group_members_removed` then `group_members_added`, matching the terminal DB state. They're never handed over concurrently or reversed.

(That ordering is only true as of 54d8c68 in this PR. Before it, `ModifyRoleGroups` fired mid-loop, *before* the commit that persisted those membership rows — which was a worse problem than ordering, since a failure in between meant the fire described a rollback.)

Where the concern does bite is across *requests*: each request drains in its own `BackgroundTask` with nothing ordering them, so "add X" and "remove X" issued by two concurrent requests can arrive in either order. That's a real property of the change rather than something I can schedule away, so it's stated in the delivery contract on `AppGroupLifecyclePluginSpec`: implement hooks as idempotent reconciliations of current state rather than applications of the delta, and take `ctx.lock` if you need mutual exclusion — it's the only mechanism that holds across workers.

Added `test_a_user_removed_and_re_added_in_one_request_replays_in_that_order` to pin the within-request half, since that's the part the code actually guarantees.

🤖 Posted by Claude Code

README.md
somethingnew2-0: Agreed, and it's already in there — `.claude/CLAUDE.md` is updated in this PR alongside the README.

The Plugin system section now carries a paragraph covering the host-side mechanics an agent needs: that request-path fires go through `defer_or_invoke_lifecycle_hook` and are recorded as plain data for a post-response replay; that the replay binds its own `_session_scope` (and why that's load-bearing for `ctx.set_group_description`); that new fires must be queued *after* the commit that makes the change durable; and that `group_deleted` is the exception and stays inline at both of its fire sites.

It also states the part that's easy to miss from the code alone — that deferral is a semantic change to the plugin interface rather than a scheduling one (hooks observe re-loaded state, are unordered across requests, and can be lost) — and points at `AppGroupLifecyclePluginSpec` as the authoritative contract, so the detail lives in one place rather than being duplicated and drifting.

`.claude/access-dev-sync.md` got the matching change on the `sync-app-groups` side, since that's where the "what recovers a lost fire" question gets answered.

🤖 Posted by Claude Code

barborico (approved):
LGTM, checking out the stacked PR now!

PR #589, opened 2026-08-21
Retry a group_deleted hook that never reached the plugin

somethingnew2-0:
# Summary

`group_deleted` is the one lifecycle hook nothing recovers when it goes missing. `sync_group` sweeps *live* app groups, so a soft-deleted one is structurally invisible to it (`AppGroup.deleted_at.is_(None)` at both `api/cli.py:412` and `:439`). If the hook never reaches the plugin, the external group and its members stay alive with nothing scanning for them — while Access shows the access as revoked. That is a divergence in the direction that matters.

So this records the delivery instead of inferring it. `DeleteGroup` writes a `PendingAppGroupDeletion` row **in the same transaction as the soft delete** — which is what makes the record survive the failure it exists for — and drops it once the hook succeeds. A new phase of `sync-app-groups` re-fires whatever is left.

**Urgency**: none, this is a durability improvement rather than a fix for a live incident.
**Expected review effort**: MEDIUM (~30 min). One new table, one new module, one CLI phase.

> ### Stack
>
> | | PR | |
> |---|---|---|
> | 1 | #585 — Defer app-group-lifecycle plugin hooks to a post-response BackgroundTask | base |
> | 2 | **this PR** — Retry a `group_deleted` hook that never reached the plugin | ← you are here |
>
> Base is `pcollins/defer-lifecycle-hooks-a41f7c`, so this diff shows only the second commit. GitHub retargets it to `main` once #585 merges. It depends on #585 for `invoke_app_group_lifecycle_hook`'s `session=` parameter and its `list[BaseException]` return.

# Motivation

#585 keeps `group_deleted` on the request rather than deferring it, precisely because "a lost fire costs latency, not correctness" is false for a delete. That removes the largest way to lose it. It does not remove every way:

- `invoke_app_group_lifecycle_hook` never propagates — by design, so a misbehaving plugin cannot fail a committed operation. A hook that raises is logged, rolled back, and gone.
- A worker that dies mid-request loses the call with the soft delete already committed.

Both leave the same orphan, and neither is hypothetical for a plugin talking to a flaky external API. This is also **pre-existing**: an inline hook that raised was already unrecoverable before #585 existed.

# Description of changes

**`PendingAppGroupDeletion`** — group id, plugin id, the member ids the delete captured, and retry bookkeeping. The FK is to `okta_group.id` rather than `app_group.id`: the base row is the one guaranteed to outlive a soft delete. Unique on `(group_id, plugin_id)`.

It carries `member_ids` rather than re-deriving them at retry time, for the same reason #588 added the argument in the first place: by then the memberships are ended, and a plugin that provisions users individually needs to know who to deprovision. Re-deriving would hand it an empty list.

**`api/operations/_lifecycle_outbox.py`** — record / settle / list-outstanding / re-load-target / record-failure. Sibling of `_lifecycle_fan_out.py` and deliberately separate: that module is about *deferring* work, this one is about *not losing* it.

**A new `sync-app-groups` phase**, running before the group sweep so an orphan is cleaned up at the earliest opportunity rather than behind every live group's reconcile. Undelivered deletes count towards the command's exit status, or a run that left an external group orphaned would look clean.

Two cases it deliberately declines, both covered by tests:

- **The group is no longer an app group.** `ModifyGroupType` hard-deletes the `app_group` row, so there is nothing to hand the hook and nothing that will change. The row is dropped.
- **The app was repointed at a different plugin.** Firing would tell the wrong plugin to delete a group it never created, and the plugin actually owed the delivery is no longer reachable through this group. Left for an operator.

**Past `MAX_DELIVERY_ATTEMPTS` (10) the row stops being retried but is deliberately *not* deleted** — it is the record that an external group may still exist, which is the thing an operator needs to see. An external API that keeps rejecting the call will not start succeeding by repetition, and retrying every run buries the signal.

# The contract change

`group_deleted` becomes **at-least-once**, so implementations must be idempotent: deleting an external group that is already gone is a success, not an error. That is stated on `AppGroupLifecyclePluginSpec` and in the README's plugin-author section.

# Scope

Only `group_deleted` gets this treatment. Every other hook is recoverable by a plugin whose `sync_group` is a full reconciliation — which is what the contract added in #585 asks for — so guaranteeing delivery of all of them would mean a DB write per fire on the request path, buying a guarantee the interface already asks plugins to provide. `ModifyGroupType`'s fire is the one delete this still cannot cover, for the reason above.

# Validation

- Full suite green: 880 passed. `ruff check`, `ruff format --check`, `ty check` clean.
- 11 new tests in `tests/test_lifecycle_delete_outbox.py`, covering the record surviving a raising hook, the membership captured before the delete rather than after, re-delivery settling the row, the retry budget and its cap, and both declined cases.
- Migration applied against SQLite locally; the `migrations (postgresql)` job covers the other backend.
- **Not done:** an end-to-end run against a real external system (killing a worker mid-delete and watching the next cronjob converge). Everything covering that path is unit/integration-level.


api/operations/_lifecycle_outbox.py
```diff
@@ -0,0 +1,158 @@
...
+    existing = (
+        await session.scalars(
+            select(PendingAppGroupDeletion)
+            .where(PendingAppGroupDeletion.group_id == group.id)
+            .where(PendingAppGroupDeletion.plugin_id == plugin_id)
+        )
+    ).one_or_none()
+    if existing is not None:
+        existing.member_ids = [member.id for member in members]
+        return
```
barborico: In this case where we're modifying an existing pending deletion and resetting the members, should we reset the attempts too?

api/cli.py
```diff
@@ -468,12 +468,83 @@ async def _sync_all_app_groups() -> int:
...
+            # The app was pointed at a different plugin after the delete. Firing would tell the
+            # wrong plugin to delete a group it never created, and the owed plugin is no longer
+            # reachable through this group, so leave the row for an operator.
+            failures += 1
+            click.echo(
+                f"  ✗ Cannot deliver group_deleted for '{group.name}': it is owed to plugin "
+                f"'{pending.plugin_id}' but the app now uses '{current_plugin_id}'",
+                err=True,
+            )
+            continue
```
barborico, line 524: Why don't we record the delivery failure here?

barborico (commented):
Approach seems sound. I'm not sure how often we'll run into the case that there's a retryable error from the group_deleted plugin which the plugin implementation itself doens't retry, but this adds some robustness without much overhead.

api/operations/_lifecycle_outbox.py
somethingnew2-0: Yes — good catch, and it was a real bug rather than a nicety.

`DeleteGroup` doesn't filter `deleted_at` when it loads the group, so deleting an already-soft-deleted one re-runs the whole operation and lands in this branch. Keeping the old attempt count meant a row that had already exhausted its budget stayed excluded from `outstanding_deletions` — so the *new* delete would never have been delivered, and nothing would have said so. That's the failure mode this table exists to prevent, reintroduced through the back door.

A second delete is a fresh obligation, so `attempts`, `last_error`, and `last_attempt_at` all reset with the new `member_ids`. Fixed in 885f846, with `test_re_deleting_a_group_resets_the_retry_budget` covering both halves: the fields reset, and the row becomes eligible for the sweep again rather than sitting silently past its budget.

🤖 Posted by Claude Code

api/cli.py
somethingnew2-0: We should, and now do — but without counting an attempt, which is the part that needed thinking about.

My original reasoning was that nothing was delivered here, so there was no attempt to record. That was half right: recording the *reason* is clearly worth it, since CLI stderr scrolls away and `last_error` on the row is what an operator actually finds later.

The catch is that `record_delivery_failure` also increments `attempts`, and rows past `MAX_DELIVERY_ATTEMPTS` drop out of `outstanding_deletions`. So recording it the ordinary way would have aged this row out of the sweep after ten runs and stopped reporting it — and unlike a flaky external API, a plugin mismatch is a misconfiguration that stays broken until a human fixes it. Silently ageing out is the wrong outcome for exactly the condition that most needs attention.

So `record_delivery_failure` grew a `count_attempt` flag: the mismatch branch records the reason and leaves the budget untouched, so the row keeps being reported every run until someone resolves it. 885f846, with the existing mismatch test extended to assert both — `last_error` is set, `attempts` stays 0.

🤖 Posted by Claude Code

somethingnew2-0:
Closing this out as I feel this addition to handle group_deleted events asynchronously is probably overkill for what we need.

PR #591, opened 2026-08-21
Add a hidden confetti cannon, revealed by the theme toggle

somethingnew2-0:
## What

https://github.com/user-attachments/assets/9e71c307-fd4b-4ce5-afa4-62255c2288da

Clicking any of the light/dark/system theme buttons three times in a row — each click within three seconds of the last — reveals a confetti button beside the toggle in the lower left of the drawer. Switching it on fires a burst from the button, and from then on every keystroke in a text field puffs a few pieces of confetti from the caret. Both the reveal and the on/off state persist in `localStorage`, so the switch stays found once discovered.

Built on [discord/confetti-cannon](https://github.com/discord/confetti-cannon) (MIT, no runtime dependencies of its own).

## Why

Surprise and delight. Access is a tool people use to ask for things and wait, and a small hidden reward costs nothing when it stays out of the way — which is the constraint that shaped most of the decisions below.

## Notes for reviewers

- **Nobody pays for it until they find it.** The overlay sits behind `React.lazy`, so `confetti-cannon` and its canvases only download once the switch is flipped; it builds as its own ~23kB (7.7kB gzipped) chunk rather than joining the main bundle.
- **`global` shim.** `confetti-cannon` scales its drawing by `global.devicePixelRatio`, a Node-ism its own webpack build shimmed away. Vite provides no `global`, so the overlay module aliases it to `globalThis` before any confetti is drawn — nothing renders at all without this. It lives in the lazy module so the rest of the app is untouched.
- **Never in the way.** The canvas is fixed, `aria-hidden`, and `pointer-events: none`, above the app bar and dialogs, so clicks pass straight through it.
- **Deliberate omissions.** Password fields and non-text inputs get no confetti, bursts are throttled so fast typing can't flood the screen, and `prefers-reduced-motion: reduce` keeps the overlay unmounted while the switch still remembers the preference for later.
- **Colors come from the MUI palette**, so it matches light and dark without hardcoding hexes.
- **One layout consequence:** with the drawer collapsed the two buttons stack, making its footer ~46px taller. The drawer paper is content-sized today, so on windows under ~595px tall the page scrolls slightly further to reach it. Happy to cap the paper at `100vh` so the drawer scrolls internally instead, if you'd prefer that here rather than as a follow-up.
- **`uuid` override.** `confetti-cannon` peer-depends on `uuid` ^9, which npm auto-installs and which carries a moderate advisory (missing buffer bounds check in v3/v5/v6 when `buf` is passed — the library only calls v4 and never passes a buffer, so it was never reachable). There is no fix inside the peer range, so the second commit pins it forward with an `overrides` entry; `npm audit` is clean again.
- **Not in the README** — documenting the gesture would spoil it. Say the word if operators should have a line anyway.

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -17,6 +17,7 @@
         "@mui/x-date-pickers": "^6.18.4",
         "@sentry/react": "^8.34.0",
         "@tanstack/react-query": "^5.101.0",
+        "confetti-cannon": "^2.3.2",
         "dayjs": "^1.11.10",
         "env-cmd": "^10.1.0",
         "react": "^18.0.0",
@@ -3636,6 +3637,13 @@
         "fsevents": "~2.3.2"
       }
     },
+    "node_modules/classnames": {
+      "version": "2.5.1",
+      "resolved": "https://registry.npmjs.org/classnames/-/classnames-2.5.1.tgz",
+      "integrity": "sha512-saHYOzhIQs6wy2sVxTM6bUDsQO4F50V9RQ22qBpEdCW+I+/Wmke2HOl6lS6dTpdxVhb88/I6+Hs+438c3lfUow==",
+      "license": "MIT",
+      "peer": true
+    },
     "node_modules/cli-cursor": {
       "version": "5.0.0",
       "resolved": "https://registry.npmjs.org/cli-cursor/-/cli-cursor-5.0.0.tgz",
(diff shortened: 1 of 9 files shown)


barborico (approved):
Essential hack week feature. :shipit:

PR #594, opened 2026-08-24
Bug fix: Cap role-derived ownerships by the membership that confers them

barborico:
# Summary

`ModifyGroupsTimeLimit.execute()` caps the ownerships a role confers using the wrong limit and under the wrong condition. Both come from one statement sitting in the ownership branch when it belongs in the membership branch. Moves it, and adds regression coverage that varies the two limits independently.

**Urgency**: 3 days
**Expected review effort**: MEDIUM — the diff is small, but it settles *which* limit governs access a role confers, and this branch contains an earlier commit that answered that question the other way.

> [!IMPORTANT]
> **The thesis of this PR changed mid-review.** The first commit fixed the crash below by switching the statement to the *ownership* limit. That is the wrong value — see Motivation — and the final commit reverses it. If you reviewed the earlier version, the disagreement is the thing worth your attention.

# Motivation

The final bulk `update(OktaUserGroupMember)` in the `if ownership_seconds_limit is not None:` branch referenced `membership_time_limit_from_now`. That value was right; its placement was not, and it bit two ways:

- **A tag setting `owner_time_limit` without `member_time_limit`** never enters the membership branch, so the name is unbound and applying the tag raises `UnboundLocalError` → HTTP 500. The statement is built unconditionally, so this fired for *any* owner-limit-only tag, not just ones on role groups.
- **A tag setting `member_time_limit` without `owner_time_limit`** never enters the ownership branch, so the statement never runs. The role's members are capped while the ownerships those memberships confer keep their old end dates — access outliving the membership it exists because of. Silent, and it returns 200.

Capping those ownerships by the owner limit instead does not hold either: The ownerships a role grants in the groups it owns are held by the role's **members**, not its owners.

The bug survived because every tag in `tests/test_time_limit_constraint.py` sets `MEMBER_TIME_LIMIT` and `OWNER_TIME_LIMIT` to the *same* value. That binds the name (no crash), makes the wrong branch indistinguishable from the right one, and makes a swap between the two values invisible — a fixture matrix that moves two knobs in lockstep is structurally blind to all three failures.

<details>

**`api/operations/modify_groups_time_limit.py`** — the derived-ownership update moves from the ownership branch to the membership branch, keyed on `membership_time_limit_from_now`. It still follows the role's *owner*-side associations, because those are the ones producing owner-side rows; what changes is which limit bounds them and when the update runs.

**`tests/test_time_limit_constraint.py`** — four tests that vary the two limits independently, which the existing suite never does:

- `test_owner_only_time_limit_leaves_role_derived_ownership_alone` — owner-limit-only tag on a role that owns another group; asserts nothing is capped, since no member limit is set. Fails pre-fix with `UnboundLocalError`.
- `test_differing_time_limits_cap_role_derived_ownership_at_member_limit` — member 7 days / owner 1 day; asserts both the derived ownership and the derived membership land at 7.
- `test_member_limit_on_a_role_reaches_the_access_it_confers` — parametrized over both association axes; asserts conferred access never outlives the membership conferring it. The member-limit-only case is the one neither `main` nor the first commit covered.
- `test_owner_limit_on_a_role_does_not_reach_the_access_it_confers` — parametrized likewise; asserts the owner limit reaches neither the role's members nor what they confer.

**`api/operations/constraints/check_for_self_add.py`** — an unrelated copy-paste slip in the same area, in its own commit:

- The owner loop's rejection message interpolated `member_group.name`, the *preceding* loop's variable. Beyond naming the wrong group, this leaked a function-scoped loop variable: a role with owner-associated groups but no member-associated ones leaves `member_group` unbound, turning a clean rejection into an `UnboundLocalError`.
- Grammar in the same message set: "is an group owner" → "a group owner", "readding" → "re-adding", and a stray double space before "has tags which restricts …" in both of `execute_for_role`'s messages.

</details>

<details>

- Each new test was confirmed to fail against the code it guards, for the predicted reason — `UnboundLocalError` for the owner-only case, and an uncapped `ended_at` for the member-only case. A regression test that passes pre-fix would prove nothing here.
- `make test` green: ruff, ty, 875 backend tests, 38 frontend tests.
- No existing test asserted on the reworded strings (`grep` over `tests/` and `src/`), so the wording commit is behavior-neutral.
- Not exercised against Postgres (`make pytest-postgres`); the change moves an existing statement between branches, so there is no dialect-specific surface.

</details>

# Guidance for Reviewers

Read just the final state (file by file).

A refactor extracting these bulk updates into composable helpers is planned to stack on top of this branch, so the structure here is not the long-term shape.

### DIFF
--- a/tests/test_time_limit_constraint.py
+++ b/tests/test_time_limit_constraint.py
@@ -1276,15 +1276,11 @@ async def test_owner_only_time_limit_caps_role_derived_ownership(
     user: OktaUser,
     url_for: Any,
 ) -> None:
-    """A tag setting only an owner time limit still caps role-derived ownerships.
+    """A tag setting only an owner time limit caps the ownerships a role grants.
 
-    The ownership branch of `ModifyGroupsTimeLimit` used to reach for
-    `membership_time_limit_from_now` when capping the ownerships a role grants
-    to its associated groups. That name is only bound by the membership branch,
-    so an owner-only tag raised `UnboundLocalError` here.
+    Memberships are left alone, since the tag carries no member limit.
     """
-    # Deliberately no MEMBER_TIME_LIMIT_CONSTRAINT_KEY -- an owner-only tag is
-    # the case that leaves the membership limit unset.
+    # Owner limit only -- the member limit is deliberately left unset.
     tag = TagFactory.build(constraints={Tag.OWNER_TIME_LIMIT_CONSTRAINT_KEY: THREE_DAYS_IN_SECONDS})
     db.session.add_all([tag, okta_group, role_group, user])
     await db.session.commit()
@@ -1304,7 +1300,6 @@ async def test_owner_only_time_limit_caps_role_derived_ownership(
     )
     assert role_owner_map_id is not None
 
-    # Baseline: nothing is time-bounded before the tag is applied.
     assert await db_count(db.session, select(OktaUserGroupMember).where(OktaUserGroupMember.ended_at.is_not(None))) == 0
 
     update_group_spy = mocker.patch.object(okta, "update_group")
@@ -1322,7 +1317,6 @@ async def test_owner_only_time_limit_caps_role_derived_ownership(
     assert rep.status_code == 200
     assert update_group_spy.call_count == 1
 
-    # The ownership the role grants `okta_group` is capped at the owner limit.
     assert (
         await db_count(
             db.session,
@@ -1336,7 +1330,6 @@ async def test_owner_only_time_limit_caps_role_derived_ownership(
         == 1
     )
 
-    # The tag sets no member limit, so the user's role membership stays indefinite.
     assert (
         await db_count(
             db.session,
@@ -1362,9 +1355,8 @@ async def test_differing_time_limits_cap_role_derived_ownership_at_owner_limit(
 ) -> None:
     """Role-derived ownerships are capped at the owner limit, not the member one.
 
-    The companion to the owner-only case above: when both limits are set but
-    differ, using `membership_time_limit_from_now` here silently capped
-    role-derived ownerships at the wrong limit instead of raising.
+    When a tag sets the two limits to different values, each side of the role's
+    grant takes its own: ownerships the owner limit, memberships the member one.
     """
     tag = TagFactory.build(
         constraints={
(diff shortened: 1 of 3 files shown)


tests/test_time_limit_constraint.py
```diff
@@ -1265,3 +1265,175 @@ async def test_time_limit_add_app_tags(
...
+    role_group: RoleGroup,
+    user: OktaUser,
+    url_for: Any,
+) -> None:
+    """A tag setting only an owner time limit still caps role-derived ownerships.
+
+    The ownership branch of `ModifyGroupsTimeLimit` used to reach for
+    `membership_time_limit_from_now` when capping the ownerships a role grants
+    to its associated groups. That name is only bound by the membership branch,
+    so an owner-only tag raised `UnboundLocalError` here.
```
matthew-bass: nit: probably good to keep docstrings relevant to just the behavior under test? This one and [the companion test](https://github.com/discord/access/blob/0a5ff6438df8c6158c100b8e84e4047b95d8acf3/tests/test_time_limit_constraint.py#L1363-L1368) narrate the old failure rather than whats there now

tests/test_time_limit_constraint.py
```diff
@@ -1265,3 +1265,175 @@ async def test_time_limit_add_app_tags(
...
+
+    role_owner_map_id = await db.session.scalar(
+        select(RoleGroupMap.id)
+        .where(RoleGroupMap.role_group_id == role_group.id)
+        .where(RoleGroupMap.group_id == okta_group.id)
+        .where(RoleGroupMap.is_owner.is_(True))
+    )
+    assert role_owner_map_id is not None
+
+    # Baseline: nothing is time-bounded before the tag is applied.
```
matthew-bass: nit: These comments mostly restate the assertions below. The same applies at [line 1325](https://github.com/discord/access/blob/0a5ff6438df8c6158c100b8e84e4047b95d8acf3/tests/test_time_limit_constraint.py#L1325), [line 1339](https://github.com/discord/access/blob/0a5ff6438df8c6158c100b8e84e4047b95d8acf3/tests/test_time_limit_constraint.py#L1339), and [lines 1415–1427](https://github.com/discord/access/blob/0a5ff6438df8c6158c100b8e84e4047b95d8acf3/tests/test_time_limit_constraint.py#L1415-L1427). Can we remove them and keep only the non-obvious shared-session note?

tests/test_time_limit_constraint.py
```diff
@@ -1265,3 +1265,175 @@ async def test_time_limit_add_app_tags(
...
+    assert update_group_spy.call_count == 1
+
+    # Ownership derived from the role is capped at the 1-day owner limit...
+    assert (
+        await db_count(
+            db.session,
+            select(OktaUserGroupMember).where(
+                OktaUserGroupMember.role_group_map_id == role_owner_map_id,
+                OktaUserGroupMember.is_owner.is_(True),
+                OktaUserGroupMember.ended_at < (datetime.now(UTC) + timedelta(days=2)),
```
matthew-bass: nit: The setup asks for a [1-day owner limit](https://github.com/discord/access/blob/0a5ff6438df8c6158c100b8e84e4047b95d8acf3/tests/test_time_limit_constraint.py#L1369-L1373), but this only checks that expiry is under 2 days. Can we add a lower bound around 1 day so an immediate or already-expired value would fail?

matthew-bass (dismissed):
looks good! just some nits that an adversarial review found for the tests that i wouldn't consider blocking

Jeff-Rowell (approved):

PR #597, opened 2026-08-24
Expire role requests and group requests, with an opt-out

barborico:
# Summary

Role requests and group requests never expired; only access requests did. This adds sweeps for both to the `access sync` cronjob, fixes a latent bug where one failing request aborted the whole pass, and lets an operator switch age-based expiration off by setting a cutoff to `never`.

**Urgency**: 5 days

# Motivation

`expire_access_requests()` has closed stale `AccessRequest` rows at the tail of the sync cronjob for a while. `RoleRequest` and `GroupRequest` had no equivalent, so a pending one stayed pending forever: it accumulated in reviewers' queues, and a role request past its `request_ending_at` remained approvable into a `RoleGroupMap` that was already expired on arrival.

Separately, the existing sweep had no error isolation. One request that failed to reject aborted expiration for every remaining request, and would fail the same way on every subsequent run, so a single poisoned row could stop expiration indefinitely.

Review feedback then asked for a way to disable expiration, which the last two commits add.

<details>

**Three focused sweeps rather than one grown function.** `expire_access_requests` keeps its name and behavior (`api/cli.py` and the existing test import it); `expire_role_requests` and `expire_group_requests` join it in the `sync` command.

| Sweep | Age cutoff | Requested-window pass |
|---|---|---|
| `expire_access_requests` | `MAX_ACCESS_REQUEST_AGE_SECONDS` | `request_ending_at < now()` |
| `expire_role_requests` | `MAX_ACCESS_REQUEST_AGE_SECONDS` | `request_ending_at < now()` |
| `expire_group_requests` | `MAX_GROUP_REQUEST_AGE_SECONDS` | none, deliberately |

**Group requests expire on the age cutoff only.** Every requested-window pass exists to stop an approval producing a dead-on-arrival grant. A role request past `request_ending_at` has that failure mode; a group request's payload is the group itself, and throwing that ask away over a stale secondary field loses more than it protects. A negative test pins this so the apparent omission does not get "fixed" later. Note that this leaves a real gap on the approve path, which #599 addresses separately; the comments here say so rather than claiming it is handled.

**Configuration.** Both `MAX_ACCESS_REQUEST_AGE_SECONDS` and the new `MAX_GROUP_REQUEST_AGE_SECONDS` are `Union[int, Literal["never"]]` with an explicit one-week default. They are **independent**; neither inherits the other, so changing one never silently moves the other. Setting either to `never` switches off its age-based sweep.

`never` rather than `None` because `Settings` reads only env vars and `.env`, where every value is a string: an omitted variable is indistinguishable from one set to nothing, and `""`, `"None"`, and `"null"` all fail int validation, so `None` is unreachable for a setting that must keep a default. A string sentinel also explains itself in a config file, where `0` would need a lookup and the natural wrong guess ("expire immediately") is the opposite of the truth.

A `field_validator` rejects ints below 1 and names `never` in the message, because `-1` is the likely Unix-habit guess for disabling and would otherwise close every pending request on the next sync. `Field(ge=...)` cannot do this: on a `Union[int, Literal["never"]]` the bound applies to the union and rejects the sentinel itself.

**Error isolation.** A shared `_expire_each` helper rejects each request in its own `try`/`except`, logging with `logger.exception` and rolling back before continuing. `RejectRoleRequest` raises `ConflictError` when a reviewer resolves a request mid-sync, which is a live race this now survives. One subtlety worth your attention: `_expire_each` reads every request id **up front**, before any rollback, and calls the reject operation with the id rather than the ORM instance, because `db.session.rollback()` expires the entire identity map and touching an attribute on a not-yet-processed instance afterward raises `MissingGreenlet`.

**No new `EventType`.** The existing `role_request_reject` / `group_request_reject` events are reused with a null actor, so operators' downstream SIEM schemas need no change.

**One intentional break:** the module-level `MAX_ACCESS_REQUEST_AGE_SECONDS` re-export in `api/config.py` is removed. Nothing in-repo read it, and the raw field is now `int | "never"` rather than a duration, so a snapshot of it would be actively misleading. An operator whose private tooling imported that name should read `settings.max_access_request_age_seconds` instead, which resolves the sentinel and returns `Optional[int]`.
</details>

<details>

- `make test` green: ruff, ty, 892 backend pytest, 37 frontend vitest.
- `tests/test_expire_role_request.py`: fresh request is a no-op; aged past the cutoff expires; past `request_ending_at` expires; `never` disables the age half while the window half still closes a lapsed request.
- `tests/test_expire_group_request.py`: fresh no-op; aged expires; **lapsed `requested_ownership_ending_at` stays PENDING** (pins the asymmetry); the group cutoff is consulted rather than the access one; `never` disables it; and disabling the *access* cutoff does not disable the group one.
- `tests/test_expire_access_request.py` gains an isolation test that injects a failure at the reject operation's own `commit`, after it has dirtied the session, and asserts the failed request is still `PENDING` — verified to fail with the rollback removed.
- `tests/test_config.py`: both defaults; independence in both directions; `never` on each; mis-cased and non-numeric values rejected; ints below 1 rejected with `never` named.
- Every claim added to the README was fact-checked against `api/config.py` and `api/syncer.py`.
- No migration, no Pydantic request/response schema change, no generated-client edit.
</details>

# Guidance for Reviewers

**Read file-by-file rather than commit-by-commit.** The branch changed its own mind partway: the config semantics introduced early (an inheriting `Optional[int]`, with `0` meaning expire-immediately) were replaced later by the independent defaults and the `never` sentinel described above. The commit messages of `d620adb` and `da4fc2a` therefore assert rationale that later commits retract, and reading them in order is misleading.

Three things most merit your eye:

1. **The first sync after deploy closes the whole existing backlog at once**, and each closure notifies the requester and every eligible approver — for a role request, that includes all Access admins. The requested-window pass for role requests is new here and is deliberately **not** covered by the disable setting, so `never` cannot make that first run quiet. Documented in the README; flagging in case you want to time the deploy.
2. **`_expire_each`'s ids-up-front contract.** The comment explains why, but a future sweep that passes it a list built somewhere other than a local query in the same function weakens the `MissingGreenlet` guarantee.
3. **`never` deliberately does not stop everything.** Access and role requests are still closed when the requester's own end date has passed. That asymmetry is intentional — the setting is named for a maximum age — and is pinned by tests on both types, but it is the thing an operator is most likely to be surprised by.

Known and accepted, not fixed here: the three sweeps are near-identical and could collapse into one parameterised helper, and none of them batch or limit. Both are worth doing; neither is new to this branch, and both would enlarge a change that is already at the edge of reviewable.

### DIFF
--- a/README.md
+++ b/README.md
@@ -288,6 +288,38 @@ the values in the default config.
 To use your custom config file, set the `ACCESS_CONFIG_FILE` environment variable to the name of your config
 override file in the project-level `config` directory.
 
+#### Request expiration
+
+The `access sync` cronjob closes pending requests that have sat too long. Two
+environment variables control it, both in seconds, both defaulting to one week:
+
+| Variable | Governs | Default |
+| --- | --- | --- |
+| `MAX_ACCESS_REQUEST_AGE_SECONDS` | access requests **and** role requests | `604800` |
+| `MAX_GROUP_REQUEST_AGE_SECONDS` | group requests | `604800` |
+
+Set either to `never` to switch off age-based expiration for the request types
+it governs:
+
+```
+MAX_ACCESS_REQUEST_AGE_SECONDS=never
+```
+
+Three things worth knowing:
+
+- **The two are independent.** Raising `MAX_ACCESS_REQUEST_AGE_SECONDS` does not
+  raise the group cutoff; set both if you want both changed.
+- **Access and role requests share one cutoff.** Disabling
+  `MAX_ACCESS_REQUEST_AGE_SECONDS` stops role requests aging out too.
+- **`never` does not mean nothing is ever closed.** Access and role requests are
+  also closed when the end date the requester asked for has already passed,
+  which is a separate check that `never` does not affect. Group requests have no
+  such check, so `never` stops their expiration entirely.
+
+A value of `0` or less is rejected at startup: it would close every pending
+request on the next sync, and is almost always a mistaken attempt to disable
+expiration. Use `never` for that.
+
 ### Sample Usage
 
 To override environment variables, create an override config file in the `config` directory. (You can name
--- a/README.md
+++ b/README.md
@@ -272,23 +272,7 @@ If you are using Cloudflare Access, ensure that you configure `CLOUDFLARE_TEAM_D
 
 Else, if you are using a generic OIDC identity provider (such as Okta), then you should configure `SECRET_KEY` and `OIDC_CLIENT_SECRETS`. `CLOUDFLARE_TEAM_DOMAIN` and `CLOUDFLARE_APPLICATION_AUDIENCE` do not need to be set and can be removed from your env file. Make sure to also mount your `client-secrets.json` file to the container if you don't have it inline.
 
-### Access application configuration
-
-_All front-end and back-end configuration overrides are **optional**._
-
-The default config for the application is at [`config/config.default.json`](config/config.default.json).
-
-The file is structured with two keys, `FRONTEND` and `BACKEND`, which contain the configuration overrides for the
-front-end and back-end respectively.
-
-If you want to override either front-end or back-end values, create your own config file based on 
-[`config/config.default.json`](config/config.default.json). Any values that you don't override will fall back to 
-the values in the default config.
-
-To use your custom config file, set the `ACCESS_CONFIG_FILE` environment variable to the name of your config
-override file in the project-level `config` directory.
-
-#### Request expiration
+##### Request expiration
 
 The `access sync` cronjob closes pending requests that have sat too long. Two
 environment variables control it, both in seconds, both defaulting to one week:
@@ -297,6 +281,9 @@ environment variables control it, both in seconds, both defaulting to one week:
 | --- | --- | --- |
 | `MAX_ACCESS_REQUEST_AGE_SECONDS` | access requests **and** role requests | `604800` |
 | `MAX_GROUP_REQUEST_AGE_SECONDS` | group requests | `604800` |
+
+These are environment variables; setting them in an `ACCESS_CONFIG_FILE` JSON
+config has no effect.
 
 Set either to `never` to switch off age-based expiration for the request types
 it governs:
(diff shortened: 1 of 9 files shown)


api/config.py
```diff
@@ -117,6 +117,16 @@ def expose_api_docs(self) -> bool:
     USER_DISPLAY_CUSTOM_ATTRIBUTES: str = "Title,Manager"
     USER_SEARCH_CUSTOM_ATTRIBUTES: Optional[str] = None
     MAX_ACCESS_REQUEST_AGE_SECONDS: int = 7 * 24 * 60 * 60
+    # Group requests get their own cutoff: their approver may need to negotiate a
+    # name and pick tags rather than just answer yes/no, which plausibly wants a
+    # longer fuse than a membership request. Unset falls back to
+    # MAX_ACCESS_REQUEST_AGE_SECONDS via `max_group_request_age_seconds`, so
+    # operators who never set it keep the behavior they already had.
```
somethingnew2-0: Let's set a default for Access 2.0, but maybe we could accept None for either MAX_ACCESS_REQUEST_AGE_SECONDS or MAX_GROUP_REQUEST_AGE_SECONDS to allow for disabling the expiration? Not sure what None looks like as an environment variable though

README.md
```diff
@@ -272,6 +272,52 @@ If you are using Cloudflare Access, ensure that you configure `CLOUDFLARE_TEAM_D
...
+- **`never` does not mean nothing is ever closed.** Access and role requests are
+  also closed when the end date the requester asked for has already passed,
+  which is a separate check that `never` does not affect. Group requests have no
+  such check, so `never` stops their expiration entirely.
+
+A value of `0` or less is rejected at startup: it would close every pending
+request on the next sync, and is almost always a mistaken attempt to disable
+expiration. Use `never` for that.
+
+**The requested-window check for role requests is new and is not covered by
```
somethingnew2-0: Let's not phrase this as what's new or not in the README.md, but what the current behavior is.

somethingnew2-0 (approved):

PR #601, opened 2026-08-25
Propagate tag constraints from tagged groups to associated roles

barborico:
# Summary

Adds `Tag.propagate_to_roles`, a single switch governing whether a tag's constraints reach roles associated with a tagged group. Propagation is computed at read time. Additionally, applicable constraints are made more visible in the UI.

**Urgency**: one week; we want to put this in place and then initiate a one-time audit in time to show off mid-September
**Expected review effort**: HIGH — breaking change, security-adjacent, 23 commits

> The tag page's "Groups with Tag" rework is **not** in this PR. It is descoped and stacked on top as
> [#602](https://github.com/discord/access/pull/602), reviewable independently once this lands.

# Motivation

Today four of the six tag constraints already reach member roles — unconditionally, invisibly, and with no way to opt out. The two **time-limit** constraints do not reach them at all. A role granting access to a compliance-scoped group means that, while the suitability of the group for the role is reviewed periodically, the suitability of the role for an individual may not be.

After this change, all six constraints propagate under one operator-visible switch, and the UI shows which constraints are in force on a group and where each one comes from.

<details>

**The propagation rule.** A role confers access upon its *members*, so propagation always feeds the role's **member-side** constraints. What differs is which key is read on the associated group:

| Association | Members of the role receive | Role's member-side constraint reads |
|---|---|---|
| Role is a **member** of group G | Membership of G | the same key on G's tags |
| Role is an **owner** of group G | Ownership of G | the **owner-side counterpart** key |

Nothing propagates to a role's own owner side — owning a role does not confer the role's grants. `OWNER_SIDE_COUNTERPART` in `api/models/tag.py` encodes both the mapping and, via its key set, which constraints propagate at all.

**Computed, not materialized.** An earlier draft materialized derived `OktaGroupTagMap` rows. That was rejected: `RoleGroupMap` is time-bounded and bulk-ended by operations that bypass `ModifyRoleGroups`, so a materialized cache would need a trigger-completeness invariant and a repair cron — and under a single gate, a missed trigger stops being "a time limit lands late" and becomes "a compliance control is silently off." A computed set cannot drift.

**Enforcement.** `effective_constraint` replaces every hand-rolled tag traversal. The associated-group loops in `CheckForSelfAdd` / `CheckForReason` are deleted; the helper covers both directions, including the **owner** loops those blocks contained, which an earlier plan would have dropped. Error messages keep naming the group that imposed the restriction, via `constraint_source_clause`.

**Time limits** are the one category that mutates state rather than vetoing, so they land twice: at grant time in `ModifyGroupUsers`, and retroactively in `ModifyGroupsTimeLimit` when a time-limited tag arrives on a group roles are already associated with. A separate `cap-role-memberships` CLI sweep covers grants predating the deploy.

**UI.** A new **Effective constraints** panel on group and role pages lists every constraint in force with its source (e.g. `SOX, via membership in App-Foo-Admin`). Tag chips are deliberately unchanged — chip styling already carries fill-vs-outline for direct-vs-app and grey for disabled, and the question users actually have is about constraints, not tags. The tag page gains a plain-language note on whether that tag reaches roles and a yes/no control to set it.

image

image

image

image

image
</details>

<details>

- `make test` green at every commit; **916 backend / 51 frontend** at head, ruff + ty clean.
- Every task was implemented TDD and independently reviewed; five needed a fix round, each re-reviewed.
- **Verified against a live local stack**, not just unit tests. Seeded a SOX tag on `App-Foo-Admin`, with `Role-Finance` a *member* of that group and `Role-Auditor` an *owner*:
  - `Role-Finance` → member time limit **90d** via `member_association`
  - `Role-Auditor` → member time limit **30d** via `owner_association` — i.e. the group's *owner* limit, read onto the role's *member* side. This is the counterpart-key rule working end to end.
  - A direct `Quarterly Renewal` tag with propagation **off** still applied its own constraints while SOX propagated — confirming the gate suppresses propagation, not direct application.
  - A group with no constraints renders no panel and no layout gap.
- Both light and dark themes checked; no console errors attributable to this branch.
- Mutation-tested where a test's value depended on it: the owner-association lookup and the `active_app_tag_mapping` eager load were each verified by removing the code under test and confirming the test fails.

</details>

# Guidance for Reviewers

**Read commit-by-commit.** Commits map one-to-one onto logical units and are individually green. Within this PR the steps depend on the one before (helpers → consumers → API → UI), so they are not further decomposable into parallel PRs.

Three places most worth your judgement:

1. **`api/models/tag.py` — the propagation rule.** Everything else calls into it. Worth checking the owner-side axis swap and that `constraint_key in OWNER_SIDE_COUNTERPART` is the right test for "does this key propagate."

2. **The breaking change (`3e52907`).** Deleting the associated-group loops means `effective_constraint` is now the *sole* enforcement path for self-add and reason on associated roles. The full suite passed with zero assertion edits, which was the parity bar. Note this also removed a latent `UnboundLocalError` — the old owner loop interpolated `member_group.name`, so an ownership-only role crashed instead of returning a clean rejection.

3. **Eager loading.** `lazy="raise_on_sql"` makes a missing loader a 500 on a live endpoint rather than a test failure, and one such bug was found and fixed during review (`d2ecb00`): `effective_constraints` raises only for groups that actually carry a tag, so an untagged smoke test proves nothing. Worth confirming no call path into `effective_constraint` lacks its loaders.

**Rollout note:** the advertised breaking change (a tag *can now stop* propagating self-add/reason) is opt-in and defaults to today's behaviour. The change every operator gets on deploy with no action is the opposite — time limits reaching roles for the first time, so role memberships begin getting shortened immediately. Worth calling out in release notes above the switch itself. `cap-role-memberships` is not wired to any cron; historical grants stay uncapped until it is run.

**Known follow-ups, deliberately out of scope:** `ModifyGroupType` doesn't reject pending `RoleRequest`s on conversion the way `DeleteGroup` does (this is the root cause that made the eager-load bug reachable); sweep-capped rows get no `ended_actor_id` / `created_reason`; the `CLAUDE.md` maintenance watermark should be refreshed once this merges.


.claude/CLAUDE.md
```diff
@@ -281,6 +281,28 @@ If you change what fields are logged or add a new `EventType`, any downstream de
 that parses these audit logs (e.g. a SIEM or a Panther detection schema) also needs to be
 updated. Those schemas typically live in the operator's own private repo, outside Access.
 
+## Comments and docstrings describe the present
```
somethingnew2-0, line 284: Thanks for adding this! 

I kept having to tell Claude to do this on my PRs 😂

api/operations/modify_role_groups.py
```diff
@@ -390,11 +390,8 @@ async def execute(self) -> RoleGroup:
             for group in groups_to_add:
                 # Handle group time limit constraints when roles are added to groups
                 # with tagged time limits as members
-                membership_ended_at = coalesce_ended_at(
-                    constraint_key=Tag.MEMBER_TIME_LIMIT_CONSTRAINT_KEY,
-                    tags=[tag_map.active_tag for tag_map in group.active_group_tags],
-                    initial_ended_at=self.groups_added_ended_at,
-                    group_is_managed=group.is_managed,
+                membership_ended_at = effective_ended_at(
```
eguerrant, line 393: This caps the new RoleGroupMap, but nothing caps the role's existing user memberships. ModifyGroupsTimeLimit only runs when tags change (modify_group_tags.py:96, modify_app_tags.py:117, modify_group_type.py:287, tags.py:173), never when an association is created. Attaching a role that has 50 indefinite members to a SOX group leaves all 50 uncapped, and renewing the role's access to the group restores every derived grant without re-reviewing role membership, which is the audit gap this PR exists to close. This recurs on every new association, it is not only a historical-grant problem, and cap-role-memberships is not on a cron, so the design's "enforcement is complete on day one" holds only for users added after the association exists. Note ModifyGroupsTimeLimit.execute() commits, so a fix has to call it once after this operation's own commit with the full groups_to_add set, not per group.

api/operations/constraints/check_for_reason.py
```diff
@@ -56,47 +56,16 @@ async def execute_for_group(self) -> Tuple[bool, str]:
...
-                require_owner_reason = coalesce_constraints(
-                    constraint_key=Tag.REQUIRE_OWNER_REASON_CONSTRAINT_KEY, tags=tags
-                )
-                if group.is_managed and require_owner_reason is True:
-                    return False, f"Reason for adding owners to {group.name} group is required due to group tags"
+                key = Tag.REQUIRE_OWNER_REASON_CONSTRAINT_KEY
+                if group.is_managed and effective_constraint(key, group) is True:
+                    clause = constraint_source_clause(key, group)
+                    return False, f"Reason for adding owners to {group.name} group is required {clause}"
             if len(self.members_to_add) > 0:
```
eguerrant, line 64: invalid_reason at line 34 uses reason.strip() == "", while the router writes body.created_reason or "" at api/routers/groups.py:707, which is truthiness-based. A whitespace-only reason is therefore rejected when this constraint applies and stored verbatim when it does not, so "provided" is judged by two different standards depending on the toggle. It puts blank-looking reasons into the audit trail.

eguerrant (changes requested):
Created a thread to discuss internally but also found some other things that needed changing. Inlined what I could but some files aren't included in the PR so I'm copy-pasting some issues below with file names and line numbers


# Not anchorable inline (files absent from the diff, post as top-level review comments)

src/helpers.tsx:73
getActiveTagsFromGroups is the shared entry point for every group-based constraint helper, minTagTimeGroups:118, requiredReasonGroups:128, and ownerCantAddSelfGroups:140, and none of them filter on propagate_to_roles. The UI still mirrors the old unconditional propagation rules in both directions.

src/pages/groups/AddUsers.tsx:135
requiredReasonGroups over the role's associated groups does not consult propagate_to_roles, so with the toggle off this dialog still marks the reason field required at line 292 while CheckForReason no longer requires it. Harmless since the value is stored either way but should be updated.

src/pages/groups/AddUsers.tsx:139
*Only possibly an issue (discussion in internal thread)*
Also in ownerCantAddSelfGroups, with the toggle off the form still blocks a role owner from self-adding while the API now allows it. That makes the toggle-off self-add bypass API-reachable rather than UI-reachable. I surfaced this issue in the thread but ideally if we decide to keep the current behavior (I don't think we should) the UI should match.

src/pages/groups/AddUsers.tsx:104
timeLimit reads only the group's own tags. The role block at 119 to 142 reassigns reason and disallow_owner_add but not timeLimit, so there is no client-side mirror for the newly propagating time limits at all. Adding a user to a role that is a member of a time-limited group offers the full unrestricted range at 147 to 167, the user picks a year, and the backend silently caps it with no error and no indication. This is the PR's headline feature being invisible exactly where the choice is made.

src/pages/groups/BulkRenewal.tsx:243
Same split on the renewal path. setRequiredReason at 338 to 351 does traverse role associations and does apply the owner-side counterpart via requiredReasonGroups(roleGroupOwnerGroups, true), so it mirrors today's backend rules correctly but is blind to propagate_to_roles. timeLimit at 243 to 253 uses only ownedGroups and memberGroups, so a renewed role membership shows no propagated cap and gets silently shortened.

Suggested fix for the frontend group

The pattern across both dialogs is consistent, the frontend already mirrors the four constraints that propagate today and has no mirror for the two that this PR adds, and it cannot see the toggle in either case. Rather than threading propagate_to_roles through getActiveTagsFromGroups and adding a role traversal for minTagTime, drive both dialogs off the API's effective_constraints. That is the "display and enforcement read the same code" property the design claims, currently achieved only for the read-only panel. EffectiveConstraints.tsx surfaces the value correctly, but it lives on the group read page and is collapsed by default, not in the dialog where the duration is chosen.

src/pages/groups/AddRoles.tsx is fine and needs no change, its minTagTime/requiredReason/ownerCantAddSelf calls at 120 to 125 read the target group's own tags, which is the role-to-group direction that correctly stays ungated.

.claude/CLAUDE.md
barborico: Same :D

api/operations/modify_role_groups.py
barborico: Confirmed, and you were right that this is not just a historical-grant problem — attaching a role with existing indefinite members left every one of them uncapped, and renewing the association rebuilt each derived grant from memberships nobody re-examined.

Fixed in #616. `ModifyRoleGroups` now calls `limit_access_conferred_by_roles` once after its own commit with the full set, per your note that `ModifyGroupsTimeLimit.execute()` commits. That helper is one of a set extracted into `api/operations/_time_limits.py` in #615, which also DRYs up the three walls of near-identical bulk updates `ModifyGroupsTimeLimit` had grown.

Two things fell out of chasing this that are worth flagging:

- The `cap-role-memberships` sweep had the same shape of bug one level down. It capped the role membership but left the rows that membership had materialized in the role's associated groups on their old end dates — so it reported a number while the access it actually confers stayed indefinite. Also fixed in #616, with a regression test.
- Propagation was not gated on the *role's* own `is_managed`, so an unmanaged role advertised constraints nothing would enforce. Fixed in the same PR.

Regression test for your case is `test_attaching_a_role_caps_its_existing_members` in `tests/test_constraint_propagation.py`, under a block that spells out why association-time capping is a separate path from tag-time capping.

api/operations/constraints/check_for_reason.py
barborico: Good catch — and the mismatch is worse than inconsistent storage, because the two standards disagree about the same string. `CheckForReason` rejects a whitespace-only reason, the router stores it verbatim when the constraint is off, so whether `"   "` counts as "provided" depends on a toggle rather than on the text.

Fixed in #613, at the boundary rather than in the constraint: `ReasonStr` is an `Annotated[Optional[str], BeforeValidator(...)]` that strips on the way in, applied to all seven inbound reason fields (`reason`, `request_reason`, `created_reason` across the request, group-member and role-member bodies). A whitespace-only reason therefore normalizes to `""` before anything downstream sees it, so `CheckForReason` and the audit trail now judge it the same way, and no call site has to remember to strip.

Doing it at the boundary rather than in `invalid_reason` also means the fix holds for the paths that never reach `CheckForReason` at all — which is where the blank-looking audit entries were coming from.

barborico (commented):
Replying here to the non-anchorable block, since it hangs together as one argument.

**You were right about the shape of the fix, and I took it.** All eight dialogs now read `effective_constraints` from the API rather than mirroring the coalescing in TypeScript, and the tag walk is deleted from `helpers.tsx`. That's #621, on top of #619 (the field on group detail) and #620 (`GET /api/constraints/effective` for the cases where there is no single group to read it off — a selection, or a group request's tags before the group exists).

Point by point:

- **`src/helpers.tsx:73`** — gone as a shared entry point. `getActiveTagsFromGroups` is now private, and the three group-level helpers survive for exactly one caller: `roles/AddGroups`, which as you noted is the role-to-group direction that correctly stays ungated. It also asks the question per option across a live search, which is not a good fit for a request per keystroke. The comment there says both things.
- **`AddUsers.tsx:104`** — the missing client-side mirror for time limits is the one I'd most wanted you to catch. It reads the resolved value now, so the picker is bounded by whatever is actually in force, propagated or not.
- **`AddUsers.tsx:135` / `BulkRenewal.tsx:243`** — both read the same resolved answer, so `propagate_to_roles` is no longer invisible to either.
- **`AddRoles.tsx`** — agreed it needed no logic change; it moved to the shared reader anyway so there is one way to ask.

**One addition to your file list:** `src/pages/roles/BulkRenewal.tsx` uses `ownerCantAddSelfGroups` in four places and had the same blindness. It also turned out to have a bug of its own — it passed the same group set to both the owner-side and member-side lookups, so a group's owner limit was being applied to rows granted as memberships. That's fixed in #621 too.

**On the internal-thread item (`AddUsers.tsx:139`):** we went with your instinct rather than making the UI match. #617 makes the combination impossible — a tag carrying either `disallow_self_add_*` key cannot have `propagate_to_roles` off, rejected at every write path, because that constraint does not merely weaken without propagation, it inverts. The owner it blocks can add themselves to a role associated with the tagged group and reach the same access by a supported path, with nothing recording that a separation-of-duties control was in play. So there is no toggle-off bypass to make API-reachable or UI-reachable; the state cannot exist.

**And on the panel being collapsed and in the wrong place** — it still lives on the group page (#622), but it is no longer carrying the "display and enforcement read the same code" claim on its own. The dialogs read the same resolved answer now, so that property holds where the choice is actually made. The panel gained provenance while I was there: each source says whether the tag is on this group, inherited from its app, or reaching it through a membership or ownership, with the group or app linked.

barborico:
Closing in favour of a stack. This branch had grown to touch enforcement, a new read API, and the whole set of access dialogs in one diff, which is not reviewable at the granularity the changes deserve — the review on it made that plain.

Nothing is dropped. Every comment here has been answered in place above, and the work is split as follows.

**On `main`, independent of each other:**

| PR | What |
|---|---|
| #612 | Steering doc: comments and docstrings describe the present |
| #613 | Normalize inbound justification text at the boundary |
| #614 | Reject explicit nulls on tag writes |

**A stack, each building on the last:**

| PR | What |
|---|---|
| #594 | Cap what a role confers by the membership that confers it |
| #615 | Extract the time-limit updates into composable helpers |
| #616 | Propagate tag constraints from tagged groups to associated roles |
| #617 | Require propagation on tags that disallow self-adds |
| #619 | Report a group's effective constraints on its detail endpoint |
| #620 | Serve effective constraints for a set of groups or tags |
| #621 | Drive the access dialogs off the effective-constraints API |
| #622 | Show a group's effective constraints, and where each comes from |

Read in that order they tell the same story this PR did, one claim at a time.

Worth knowing before reviewing them:

- **#594 changed meaning.** It originally fixed a crash by switching role-derived ownership rows to the *owner* time limit. That inverted correct semantics: what a role confers is bounded by the membership conferring it, which is member-side. Its own test docstring contradicted its code. Amended, which also left #615 a genuinely pure refactor.
- **The self-add-without-propagation question** raised in the internal thread is decided in #617, and decided your way.
- **#621 carries the frontend rework** you sketched in the review here, plus three defects found while doing it.

Anything unaddressed, comment on the relevant PR rather than here and I'll pick it up there.

api/operations/constraints/check_for_reason.py
barborico: Adjacent to this thread rather than the same issue, but it belongs here because it is the other way the audit trail was saying less than it appeared to: #626.

A row materialized because a user is in a role and the role reaches a group carried only *one* of the two reasons behind it — and which one depended on which decision happened second. `ModifyRoleGroups` stored why the role was attached, `ModifyGroupUsers` stored why the user joined the role, so a reader of the group's audit log got half the justification with nothing marking it as a half. Rows repaired by `verify_and_fix_role_memberships` carried no reason at all.

Both halves are now recorded on labelled lines, and a missing half keeps its label rather than being dropped — so a one-sided reason stays distinguishable from a row written before the change, which is the same ambiguity you were pointing at with blank-looking entries.

It branches off `main` rather than the stack; it turned out to share no lines with the propagation work despite touching the same three files.

PR #604, opened 2026-08-26
Move the example plugins to pyproject.toml, unblocking Dependabot

somethingnew2-0:
## Problem

Every Dependabot `uv` job has been failing since [#552](https://github.com/discord/access/pull/552), during **file fetching**, before any dependency resolution happens:

```
ERROR Error during file fetching; aborting: The following path based dependencies
could not be retrieved: "app_group_lifecycle_audit_logger" at /pyproject.toml
```

#552 added the in-repo audit-logger example to the `dev` group as a `[tool.uv.sources]` path dependency. Dependabot's uv file fetcher walks that table and, for each entry, unconditionally fetches `<path>/pyproject.toml` (`fetch_project_file` in `uv/lib/dependabot/uv/file_fetcher.rb`); a `DependencyFileNotFound` there is re-raised as `PathDependenciesNotReachable` and aborts the job. The plugin declared itself through a legacy `setup.py`, which `uv` is perfectly happy with but the fetcher is not. The reported location is the *declaring* manifest, hence `at /pyproject.toml` rather than the plugin directory.

This takes down the entire root `uv` ecosystem, **security updates included**. The run that surfaced it was a `cryptography` advisory bump, and the default branch is currently carrying 9 open advisories (6 high, 3 moderate) that Dependabot cannot propose fixes for. No root-uv PR has opened since the path source landed. The npm, github-actions, and docker ecosystems are separate jobs and were unaffected, which is why this went unnoticed for a month.

## Fix

The first commit adds the one `pyproject.toml` that unblocks the job. The second converts the remaining six examples, so all seven now declare packaging the way the packaging standards do (PEP 517/518 for the backend, PEP 621 for the metadata) instead of leaving one plugin looking different for reasons only a comment explained. These examples are explicitly templates ("copy the example closest to your use case"), so the form they ship is the form operators copy into their own repos.

**The conversion is metadata-preserving.** Building every example from `origin/main` and from this branch and diffing the resulting wheels, the only difference anywhere is that `Dynamic:` markers disappear from `METADATA`. Those fields were previously computed by executing `setup.py`; they are now statically declared. Names, versions, `Requires-Dist`, entry points, top-level modules, and payload file lists are byte-identical.

```
   7 < Dynamic: requires-dist
   2 < Dynamic: summary
   2 < Dynamic: author
```

The four examples that declared no `version` were being built as `0.0.0`. PEP 621 makes the field mandatory, so that value is spelled out rather than quietly promoted to a real one, keeping this a pure format change.

## Two traps worth flagging for review

**Dotted entry point groups must stay quoted.** `[project.entry-points.access.commands]` is TOML for a `commands` table nested inside `access`, which would register the health-check CLI command under a group nothing looks in. Both the file and `_load_plugin_commands`' docstring now say so.

**The Dockerfile's `rw` bind mount is still required.** Its comment justified `rw` by "every example uses a legacy setup.py", which invites deleting it now. But egg-info is written in-tree by the setuptools *backend* regardless of where metadata comes from (verified: a pyproject-only build still emits `access_notifications.egg-info/`), so the constraint outlives setup.py. Only the reasoning changed.

## Guarding the regression

New test asserting every `[tool.uv.sources]` path contains a `pyproject.toml`. This is the invariant the first commit fixed by hand, and nothing else in CI notices when it breaks: uv builds either form happily, so the first symptom is Dependabot silently declining to run. Verified it fails with a useful message when the file is removed.

The two example-plugin guards globbed `*/setup.py`, so this conversion would have dropped every plugin from their coverage **without failing**, which is precisely the silent-win failure mode `test_example_plugin_packaging.py` exists to catch. Both now read either form, and the pyproject reader rejects setuptools auto-discovery, since that would leave the top-level collision check with nothing to compare.

## Incidental

Giving the Google example a `pyproject.toml` let `ty` resolve that module for the first time, surfacing eleven `not-subscriptable` errors in its own test file. They are real unchecked assumptions on `Optional` returns, so the tests now assert the invariant rather than the ty rules being relaxed to hide it.


examples/plugins/app_group_lifecycle_google/pyproject.toml
```diff
@@ -0,0 +1,31 @@
...
+requires = ["setuptools>=77"]
+build-backend = "setuptools.build_meta"
+
+[project]
+name = "app_group_lifecycle_google"
+version = "0.1.0"
+description = "Plugin for managing Google groups linked to Access groups"
+authors = [{ name = "Discord" }]
+dependencies = [
+    "SQLAlchemy",
```
barborico, line 17: Out of scope for this PR, but I expect that we can remove this here and for the audit logger, given the app group lifecycle plugin interface refactor; the plugin implementations should no longer directly interact with SQL.

barborico (approved):

examples/plugins/app_group_lifecycle_google/pyproject.toml
somethingnew2-0: Opened https://github.com/discord/access/pull/610 to fix

PR #614, opened 2026-09-01
Reject an explicit null on a partial tag update

barborico:
# Summary

`PUT /api/tags/{id}` returns a 500 when a client sends an explicit `null` for `name` or `enabled`. Types those two fields so they cannot hold one, giving the caller a 400 naming the field — and a generated client that will not let them send it at all.

**Urgency**: 1 week
**Expected review effort**: LOW — two field declarations, plus tests.

# Motivation

`UpdateTagBody` typed every field `Optional[...]` so a partial `PUT` could omit it. That conflates two separate things: a field is omittable because it has a **default**, and nullable because of its **annotation**. Only the first is needed for a partial update.

The consequence was that an explicit JSON `null` was indistinguishable from omission by type — the two differed only by presence in `model_fields_set`, which `exclude_unset=True` preserves. The handler then wrote the `None` straight through:

```python
for key in ("name", "description", "constraints", "enabled"):
    if key in payload:
        setattr(tag, key, payload[key])
```

`name` and `enabled` back `NOT NULL` columns, so the flush failed and a client mistake surfaced as an unhandled `IntegrityError` → HTTP 500. Verified against a running instance:

| body | before |
|---|---|
| `{"name": null}` | **500** |
| `{"enabled": null}` | **500** |
| `{"description": null}` | 400 / coerced to `""` |
| `{"constraints": null}` | 200, coerced to `{}` |

Nothing invalid was persisted — the transaction fails — so this is a wrong status code and an unhelpful error, not corruption.

<details>

`name` and `enabled` are declared `str` and `bool` rather than `Optional`. Pydantic then rejects an explicit null before any validator runs, and the generated TypeScript types them as non-nullable, so a caller is stopped at compile time rather than by a round trip.

`description` and `constraints` stay nullable. Both have a meaningful empty value and the handler already coerces a null to `""` and `{}`. Treating a null there as "clear this" is reasonable; for a boolean or a name there is no equivalent, so a null is a client error rather than an instruction.

**On the defaults.** A non-`Optional` field still needs one, or it becomes *required* and partial updates break. Those defaults are never applied — the handler dumps with `exclude_unset=True`, so an omitted field is absent from the payload rather than set to its default — but Pydantic publishes them, and `default: ""` beside `minLength: 1` would tell a client that omitting the name writes an empty string. A small `json_schema_extra` hook drops `default` from the published schema for these two fields.

The result is a contract that is accurate in every respect: the field may be omitted, may not be null, and has no default.

```diff
- name?: string | null;
+ /** @maxLength 255 @minLength 1 */
+ name?: string;
- enabled?: boolean | null;
+ enabled?: boolean;
```

</details>

<details>

- Parametrized test over `name` and `enabled`: explicit null returns 400, names the field, and leaves the row untouched. Both confirmed to fail before the change with `IntegrityError` → 500.
- Two positive controls, to pin what the rejection must *not* break: `description`/`constraints` still coerce a null to their empty value, and omitting a field still leaves it unchanged.
- The published schema checked directly: no `default` on either field, `type: string` rather than `anyOf` with null, and neither field in `required`.
- `make test` green: ruff, ty, 913 backend, 38 frontend.

</details>

# Guidance for Reviewers

The design question is reject-vs-coerce. Coercing a null to the column default would silently flip `enabled` to `True` for a client that sent something meaningless, which seemed worse than an error. Rejecting also keeps omission as the single documented way to say "leave this unchanged".

`UpdateAppBody`, directly below, still has the older `Optional[...] = Field(default=None, ...)` shape for its own partial update. Left alone deliberately to keep this diff to the reported bug; worth deciding separately whether it should follow.

### DIFF
--- a/api/schemas/requests_schemas.py
+++ b/api/schemas/requests_schemas.py
@@ -400,6 +400,13 @@ class CreateTagBody(BaseModel):
         return _validate_tag_constraints(v)
 
 
+#: Fields on a tag update whose column forbids null and for which no empty
+#: value is meaningful. `description` and `constraints` are absent because they
+#: have one -- the handler coerces a null there to `""` / `{}`. Omitting a
+#: field remains the way to leave it unchanged.
+_TAG_FIELDS_REJECTING_NULL = ("name", "enabled")
+
+
 class UpdateTagBody(BaseModel):
     """Body for PUT /api/tags/{id}. All fields optional (partial update)."""
 
@@ -416,6 +423,25 @@ class UpdateTagBody(BaseModel):
             return self
         if settings.REQUIRE_DESCRIPTIONS and (self.description is None or self.description == ""):
             raise ValueError("Description is required.")
+        return self
+
+    @model_validator(mode="after")
+    def _reject_explicit_nulls(self) -> Self:
+        """Reject a null sent for a field that cannot hold one.
+
+        Every field is `Optional` so that a partial update may omit it, which
+        makes an explicit `null` indistinguishable from omission by type --
+        only by presence in `model_fields_set`. Left alone it reaches a
+        `NOT NULL` column and fails the flush, turning a client mistake into a
+        500.
+        """
+        nulled = [
+            field
+            for field in _TAG_FIELDS_REJECTING_NULL
+            if field in self.model_fields_set and getattr(self, field) is None
+        ]
+        if nulled:
+            raise ValueError(f"May not be null: {', '.join(nulled)}. Omit a field to leave it unchanged.")
         return self
 
     @field_validator("constraints")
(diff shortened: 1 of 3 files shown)


api/schemas/requests_schemas.py
somethingnew2-0: Should this be modified to not be optional instead of using a custom model_validator?

somethingnew2-0: Should this be modified to not be optional instead of using a custom model_validator?

barborico: Yeah, I confirmed that this isn't needed for partial updates. Switched both fields to plain `str` / `bool`.

The one quirk was that a non-Optional field needs a default, and Pydantic publishes it. Since `exclude_unset=True` means the default is never applied, I suppressed it from the schema with `json_schema_extra`. The contract now says exactly what's true: may be omitted, may not be null, no default.

matthew-bass (approved):

somethingnew2-0 (approved):

PR #615, opened 2026-09-01
Extract the time-limit updates into composable helpers

barborico:
# Summary

Extracts the bulk `update` statements in `ModifyGroupsTimeLimit` into three named helpers in `api/operations/_time_limits.py`. **No behavior change** — no test changed, and the same 875 pass.

**Urgency**: 3 days
**Expected review effort**: LOW — verify it is a no-op; the tests are unchanged on purpose.

> [!NOTE]
> Stacked on **#594**, which fixes the defect described below. Review that first; this PR's diff is against its branch and GitHub will retarget to `main` when it merges.

# Motivation

`ModifyGroupsTimeLimit.execute()` was three walls of near-identical bulk `update` statements, differing only in which table, which `is_owner` side, and which limit variable. Two problems with that shape:

**It caused a real defect.** The ownership branch reached for the membership branch's local, which is the bug #594 fixes. When four statements differ by one identifier each, picking the wrong identifier is invisible at review.

**It hides the rule.** What the statements collectively express — that a membership materialized from a role must never outlive either record it derives from — lived only in the shape of the SQL. There was nowhere to write it down and nothing to point a future caller at.

<details>

Three helpers, named for the question each answers:

- `limit_memberships_to_groups(group_ids, *, is_owner, end_at)` — cap access granted **to** these groups, covering both the direct user rows and the `RoleGroupMap`s.
- `limit_memberships_by_roles(role_ids, *, is_owner, end_at)` — cap access granted **by** these roles through one association side.
- `limit_access_conferred_by_roles(role_ids, *, end_at)` — the composite: cap membership of the roles and everything that membership confers, following both association sides.

Each narrows `ended_at` and never extends it — every update is guarded on the existing end date being absent or later than the limit — so they are idempotent and compose in any order. None of them commit; the caller owns the transaction.

The module docstring carries two things that previously had nowhere to live: the invariant above, and the precondition that callers pass only managed, non-deleted ids. That filter stays with the caller deliberately — re-deriving it inside every helper would add a join to each bulk update and would quietly absorb a caller's mistake rather than letting it surface.

</details>

<details>

- **No test file is touched.** That is the claim: `git diff --name-only` lists one file. The suite that covered the old shape covers the new one unchanged.
- `make test` green: ruff, ty, 875 backend, 38 frontend — the same count as the base branch.

</details>

# Guidance for Reviewers

Read this as a no-op and check that it is one. The mapping is one helper call per former statement group, so the useful review is confirming each call reproduces the `WHERE` clause it replaced — same table, same `is_owner` side, same limit.

One intentional difference: the two mid-method commits become a single commit at the end, so a failure in the ownership pass no longer leaves the membership pass durably applied. That is a change in transaction boundary, not in the rows written.

### DIFF
--- a/api/operations/modify_groups_time_limit.py
+++ b/api/operations/modify_groups_time_limit.py
@@ -1,18 +1,35 @@
 from datetime import UTC, datetime, timedelta
 
-from sqlalchemy import func, or_, select, update
+from sqlalchemy import select
+
 from api.extensions import db
-from api.models import OktaGroup, OktaUserGroupMember, RoleGroup, RoleGroupMap, Tag
+from api.models import OktaGroup, RoleGroup, Tag
 from api.models.tag import coalesce_constraints
+from api.operations._time_limits import limit_access_conferred_by_roles, limit_memberships_to_groups
 
 
 class ModifyGroupsTimeLimit:
+    """Apply the time limits a set of tags imposes on a set of groups.
+
+    Runs when a tag's constraints start applying to groups that already have
+    access granted on them -- the tag being attached, edited, or inherited from
+    an app. Existing grants are shortened to fit; nothing is ever extended.
+
+    A tag reaches access on two sides:
+
+    - Its **member** limit caps membership of the tagged groups, and where a
+      tagged group is a role, everything that membership confers in the groups
+      the role is associated with.
+    - Its **owner** limit caps ownership of the tagged groups. Owning a role
+      confers none of the role's access, so nothing follows from it.
+    """
+
     def __init__(self, groups: list[str] | set[str], tags: list[str] | set[str]):
         self.group_ids = groups
         self.tag_ids = tags
 
     async def execute(self) -> None:
-        # Only include groups that are managed
+        # Only managed groups are subject to constraints.
         groups = (
             await db.session.scalars(
                 select(OktaGroup)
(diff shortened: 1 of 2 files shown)


api/operations/modify_groups_time_limit.py
```diff
@@ -29,139 +49,30 @@ async def execute(self) -> None:
...
-                        OktaUserGroupMember.ended_at.is_(None),
-                        OktaUserGroupMember.ended_at > membership_time_limit_from_now,
-                    )
-                )
-                .values({OktaUserGroupMember.ended_at: membership_time_limit_from_now})
-                .execution_options(synchronize_session="fetch")
-            )
-            await db.session.commit()
+            end_at = datetime.now(UTC) + timedelta(seconds=membership_seconds_limit)
+            await limit_memberships_to_groups(group_ids, is_owner=False, end_at=end_at)
```
Jeff-Rowell, line 67: `limit_memberships_to_groups` gets called inside `limit_access_conferred_by_roles` with the same arguments. Can we probably remove the redundant `limit_memberships_to_groups` call?

Jeff-Rowell (approved):

api/operations/modify_groups_time_limit.py
barborico: Valid observation: `role_group_ids ⊆ group_ids`, so here the inner call tries to cap rows the outer call just set. However, `limit_access_conferred_by_roles` will eventually (later in the stack) have more callers that depend on the inner call.

FWIW there's no cost to the redundancy in this case: `limit_memberships_to_groups` guards on `ended_at > end_at OR ended_at IS NULL`, and both calls compute `end_at` from the same variable, so the second pass matches zero rows.

PR #616, opened 2026-09-01
Propagate tag constraints from tagged groups to associated roles

barborico:
# Summary

Adds `Tag.propagate_to_roles`, one switch governing whether a tag's constraints reach roles associated with a tagged group, and makes all six constraints propagate under it. Enforcement only — no new API surface, no UI.

**Urgency**: 1 week — this closes a SOX audit finding, with a one-time review wave scheduled behind it.
**Expected review effort**: HIGH — breaking change, security-adjacent, ~2000 lines.

> [!NOTE]
> Stacked on **#615** → **#594**. Review those first; this diff is against #615's branch and GitHub will retarget as they merge.
>
> Supersedes **#601**, which carried this work plus the read API and UI. Those are split out and will stack on this one.

# Motivation

A compliance tag such as `SOX` is applied to an app; its `AppGroup`s inherit it. A role that is a member of one of those groups grants its own members access to a SOX-scoped group — so that role's membership must respect the same limits, without an operator hand-tagging every such role and remembering to untag it when the role's associations change.

Today four of the six constraints already reach member roles: unconditionally, invisibly, and with no way to opt out. The two **time-limit** constraints do not reach them at all. That is the audit finding — renewing a role's access to a group restores every grant the role confers, without anyone reviewing whether those members should still be in the role.

<details>

**The switch.** One boolean column on `tag`, beside `enabled`, defaulted on via `server_default` so every existing tag keeps behaving as it does today with no backfill. Deliberately not a seventh entry in `Tag.CONSTRAINTS`: it must never be coalesced across tags (each tag decides whether *its own* constraints propagate), and its default inverts relative to the five existing booleans. As a column both problems are structurally impossible rather than prevented by a comment.

**Computed, not materialized.** `RoleGroupMap` is time-bounded and bulk-ended by `UnmanageGroup` and `DeleteGroup`, both of which bypass `ModifyRoleGroups`. A materialized cache would need a trigger-completeness invariant and a repair cron — and under a single gate a missed trigger stops being "a time limit lands late" and becomes "a control is silently off". A computed set cannot drift.

**The propagation rule.** A role confers access on its *members*, so propagation always feeds the role's member side. What differs is which key is read on the associated group:

| Association | Members of the role receive | Role's member-side constraint reads |
|---|---|---|
| Role is a **member** of G | Membership of G | the same key on G's tags |
| Role is an **owner** of G | Ownership of G | the **owner-side counterpart** key |

`OWNER_SIDE_COUNTERPART` encodes both the pairing and, by its key set, which constraints propagate at all. Nothing reaches a role's owner side — owning a role confers none of its grants.

**One enforcement path.** `effective_constraint` replaces every hand-rolled traversal; the associated-group loops in `CheckForSelfAdd` / `CheckForReason` are deleted, including the **owner** loops they contained. Error messages still name the group that imposed the restriction, via `constraint_source_clause`.

**Time limits land in three places**, because they mutate state rather than veto:

1. Grant time, in `ModifyGroupUsers`.
2. A time-limited tag arriving on a group roles are already associated with, in `ModifyGroupsTimeLimit`.
3. A role being attached to such a group, or that attachment renewed, in `ModifyRoleGroups`. Without this the gap recurs on every renewal rather than only for grants predating the deploy.

All three go through `limit_access_conferred_by_roles` (from #615), so a capped role membership never leaves behind the access it confers. The `cap-role-memberships` CLI sweep covers pre-existing grants and uses the same helper for the same reason — assigning `ended_at` directly would shorten the membership while the rows actually holding the Okta groups stayed indefinite.

**Unmanaged roles.** Propagation is gated on the role's own `is_managed` inside `_propagated_sources`. Every enforcement path already checks that separately, so gating in one place is what keeps the computed set from advertising constraints nothing applies.

</details>

<details>

- `make test` green: ruff, ty, 922 backend, 38 frontend.
- **Parity is the bar for the breaking change**: with propagation on, self-add and reason behave identically to `main`. The existing suites pass with no assertion edits except the two time-limit count tests, whose new numbers are the propagated caps.
- Both association axes covered for each constraint, plus the negative: nothing reaches a role's owner side from either direction.
- Gate coverage: off means none of the six reach associated roles while direct and app-group enforcement is untouched; and per-tag rather than global, so one opted-out tag on a group does not drag another's constraints along.
- Mutation-tested where a test's value depended on it — the unmanaged-role gate was removed and its test confirmed to fail.
- Verified against a live local stack in the previous incarnation of this work (#601): a SOX tag on `App-Foo-Admin` with `Role-Finance` a member and `Role-Auditor` an owner produced a 90-day member cap and a 30-day cap read from the group's *owner* limit respectively.

</details>

# Guidance for Reviewers

Three places most worth your judgement:

1. **`api/models/tag.py` — the propagation rule.** Everything calls into it. Worth checking the owner-side axis swap, and that `constraint_key in OWNER_SIDE_COUNTERPART` is the right test for "does this key propagate".
2. **Deleting the associated-group loops** makes `effective_constraint` the *sole* enforcement path for self-add and reason on associated roles. The full suite passing with no assertion edits was the parity bar.
3. **Eager loading.** `lazy="raise_on_sql"` makes a missing loader a 500 on a live endpoint rather than a test failure, and `effective_constraint` raises only for groups that actually carry a tag — so an untagged smoke test proves nothing. Worth confirming no call path into it lacks its loaders.

**Rollout note:** There's both a new opt-in capability (a tag *can now stop* propagating self-add and reason) and opt-out breaking behavior change: time limits reach roles for the first time, so new role memberships begin getting shortened immediately. Capping existing memberships requires manual action, as `cap-role-memberships` is deliberately not wired to any cron.

### DIFF
--- a/api/integrity.py
+++ b/api/integrity.py
@@ -2,13 +2,13 @@ import logging
 from datetime import UTC, datetime, timedelta
 
 from sqlalchemy import func, or_, select, update
-from sqlalchemy.orm import selectinload
 from api.extensions import db
-from api.models import OktaGroup, OktaGroupTagMap, OktaUserGroupMember, RoleGroup, RoleGroupMap, Tag
+from api.models import OktaGroup, OktaUserGroupMember, RoleGroup, RoleGroupMap, Tag
 from api.models.tag import effective_constraint
+from api.operations import UnmanageGroup
 from api.operations._derived_reason import role_derived_reason
 from api.operations._time_limits import limit_access_conferred_by_roles
-from api.operations import UnmanageGroup
+from api.routers._eager import effective_constraint_options
 from api.services import okta
 
 logger = logging.getLogger(__name__)
@@ -190,17 +190,7 @@ async def cap_role_memberships(dry_run: bool = False) -> int:
     roles = (
         await db.session.scalars(
             select(RoleGroup)
-            .options(
-                selectinload(OktaGroup.active_group_tags).joinedload(OktaGroupTagMap.active_tag),
-                selectinload(RoleGroup.active_role_associated_group_member_mappings)
-                .joinedload(RoleGroupMap.active_group)
-                .selectinload(OktaGroup.active_group_tags)
-                .joinedload(OktaGroupTagMap.active_tag),
-                selectinload(RoleGroup.active_role_associated_group_owner_mappings)
-                .joinedload(RoleGroupMap.active_group)
-                .selectinload(OktaGroup.active_group_tags)
-                .joinedload(OktaGroupTagMap.active_tag),
-            )
+            .options(*effective_constraint_options())
             .where(RoleGroup.deleted_at.is_(None))
             .where(RoleGroup.is_managed.is_(True))
         )
@@ -227,11 +217,6 @@ async def cap_role_memberships(dry_run: bool = False) -> int:
         ).all()
         capped += len(over_long)
         if not dry_run and len(over_long) > 0:
-            # Go through the shared helper rather than assigning `ended_at`
-            # here: capping the membership alone leaves the rows it
-            # materialized in the role's associated groups on their old end
-            # dates, so the sweep would shorten the membership without
-            # shortening the access it actually confers.
             await limit_access_conferred_by_roles([role.id], end_at=limit_from_now)
         if len(over_long) > 0:
             logger.info(
--- a/api/integrity.py
+++ b/api/integrity.py
@@ -1,9 +1,13 @@
 import logging
+from datetime import UTC, datetime, timedelta
 
 from sqlalchemy import func, or_, select, update
+from sqlalchemy.orm import selectinload
 from api.extensions import db
-from api.models import OktaGroup, OktaUserGroupMember, RoleGroupMap
+from api.models import OktaGroup, OktaGroupTagMap, OktaUserGroupMember, RoleGroup, RoleGroupMap, Tag
+from api.models.tag import effective_constraint
 from api.operations._derived_reason import role_derived_reason
+from api.operations._time_limits import limit_access_conferred_by_roles
 from api.operations import UnmanageGroup
 from api.services import okta
 
@@ -174,3 +178,70 @@ async def verify_and_fix_role_memberships(dry_run: bool = False) -> None:
                         # Remove user from okta group owners
                         # https://help.okta.com/en-us/Content/Topics/identity-governance/group-owner.htm
                         await okta.remove_owner_from_group(active_role_group_map.group_id, user_id)
+
+
+async def cap_role_memberships(dry_run: bool = False) -> int:
+    """Cap existing role memberships against propagated time limits.
+
+    The enforcement path (grant time) is complete on its own; this is the
+    one-off sweep over grants that predate the feature. Delta-based, so it is
+    safe to re-run: a membership already at or under its limit is skipped.
+    """
+    roles = (
+        await db.session.scalars(
+            select(RoleGroup)
+            .options(
+                selectinload(OktaGroup.active_group_tags).joinedload(OktaGroupTagMap.active_tag),
+                selectinload(RoleGroup.active_role_associated_group_member_mappings)
+                .joinedload(RoleGroupMap.active_group)
+                .selectinload(OktaGroup.active_group_tags)
+                .joinedload(OktaGroupTagMap.active_tag),
+                selectinload(RoleGroup.active_role_associated_group_owner_mappings)
+                .joinedload(RoleGroupMap.active_group)
+                .selectinload(OktaGroup.active_group_tags)
+                .joinedload(OktaGroupTagMap.active_tag),
+            )
+            .where(RoleGroup.deleted_at.is_(None))
+            .where(RoleGroup.is_managed.is_(True))
+        )
+    ).all()
+
+    capped = 0
+    for role in roles:
+        seconds_limit = effective_constraint(Tag.MEMBER_TIME_LIMIT_CONSTRAINT_KEY, role)
+        if seconds_limit is None:
+            continue
+        limit_from_now = datetime.now(UTC) + timedelta(seconds=seconds_limit)
+        over_long = (
+            await db.session.scalars(
+                select(OktaUserGroupMember)
+                .where(OktaUserGroupMember.group_id == role.id)
+                .where(OktaUserGroupMember.is_owner.is_(False))
+                .where(
+                    or_(
+                        OktaUserGroupMember.ended_at.is_(None),
+                        OktaUserGroupMember.ended_at > limit_from_now,
+                    )
+                )
+            )
+        ).all()
+        capped += len(over_long)
+        if not dry_run and len(over_long) > 0:
+            # Go through the shared helper rather than assigning `ended_at`
+            # here: capping the membership alone leaves the rows it
+            # materialized in the role's associated groups on their old end
+            # dates, so the sweep would shorten the membership without
+            # shortening the access it actually confers.
+            await limit_access_conferred_by_roles([role.id], end_at=limit_from_now)
+        if len(over_long) > 0:
+            logger.info(
+                "cap_role_memberships: role=%s limit=%ss over_long=%d dry_run=%s",
+                role.name,
+                seconds_limit,
+                len(over_long),
+                dry_run,
+            )
+
+    if not dry_run:
+        await db.session.commit()
+    return capped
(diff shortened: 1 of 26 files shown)


api/routers/role_requests.py
```diff
@@ -185,16 +186,30 @@ async def list_role_requests(
...
+                                    # depend on which constraint key happens to be passed.
+                                    selectin_polymorphic(OktaGroup, [AppGroup, RoleGroup]),
+                                    selectinload(RoleGroup.active_role_associated_group_member_mappings)
+                                    .joinedload(RoleGroupMap.active_group)
+                                    .selectinload(OktaGroup.active_group_tags)
+                                    .joinedload(OktaGroupTagMap.active_tag),
+                                    selectinload(RoleGroup.active_role_associated_group_owner_mappings)
+                                    .joinedload(RoleGroupMap.active_group)
+                                    .selectinload(OktaGroup.active_group_tags)
+                                    .joinedload(OktaGroupTagMap.active_tag),
```
barborico: These appear to duplicate the `effective_constraint_options()` added above. Am I misreading? If not, can we use that helper to DRY it up?

Same for the two other big options additions in this file.

api/integrity.py
```diff
@@ -164,3 +168,70 @@ async def verify_and_fix_role_memberships(dry_run: bool = False) -> None:
...
+            .options(
+                selectinload(OktaGroup.active_group_tags).joinedload(OktaGroupTagMap.active_tag),
+                selectinload(RoleGroup.active_role_associated_group_member_mappings)
+                .joinedload(RoleGroupMap.active_group)
+                .selectinload(OktaGroup.active_group_tags)
+                .joinedload(OktaGroupTagMap.active_tag),
+                selectinload(RoleGroup.active_role_associated_group_owner_mappings)
+                .joinedload(RoleGroupMap.active_group)
+                .selectinload(OktaGroup.active_group_tags)
+                .joinedload(OktaGroupTagMap.active_tag),
```
barborico: Can we use the helper here too?

api/integrity.py
```diff
@@ -164,3 +168,70 @@ async def verify_and_fix_role_memberships(dry_run: bool = False) -> None:
...
+                )
+            )
+        ).all()
+        capped += len(over_long)
+        if not dry_run and len(over_long) > 0:
+            # Go through the shared helper rather than assigning `ended_at`
+            # here: capping the membership alone leaves the rows it
+            # materialized in the role's associated groups on their old end
+            # dates, so the sweep would shorten the membership without
+            # shortening the access it actually confers.
```
barborico: This comment seems superfluous, focused on a change from prior behavior
```suggestion
```

tests/test_constraint_propagation.py
```diff
@@ -0,0 +1,661 @@
...
+#
+# The retroactive block above runs when a time-limited *tag* lands on a group.
+# The other way an association starts being governed is the association itself
+# being created, and that path never triggered a cap: a role with existing
+# indefinite members attached to a time-limited group left every one of those
+# memberships uncapped. The role's access to the group is bounded (the
+# `RoleGroupMap` is capped at creation, and derived rows take the minimum), but
+# membership *of the role* is never forced through review -- so renewing the
+# role's access rebuilds every derived grant from memberships nobody
+# re-examined. This recurs on every new association, not just historically.
```
barborico: This comment describes past behavior, but it should limit itself to present behavior/intent.

api/routers/role_requests.py
barborico: Not misreading — the helper was dead code. Nothing called it, and it also omitted the group's own tags despite claiming to cover everything `effective_constraint` reads.

Folded `active_group_tags` into `effective_constraint_options()` and used it at all three sites here. The `selectin_polymorphic` stays inline at each one because it is what makes the `RoleGroup` paths resolvable on a polymorphic `OktaGroup` query — `cap_role_memberships` selects `RoleGroup` directly and needs no such pairing, so it does not belong inside the helper. Noted that split in the docstring.

One real change fell out: these sites had `joinedload(RoleGroupMap.active_group)` while the helper declares `selectinload`. Now uniform, which also means a future query combining this with `role_group_map_options()` cannot hit `Loader strategies ... conflict`. Covered by `test_role_request_assignee_filter_excludes_propagated_self_add_role_target` and the admin-branch test below it; full suite still 922.

2c9f3da

api/integrity.py
barborico: Yes — done. `select(RoleGroup)` needs no polymorphic pairing, so it is just `.options(*effective_constraint_options())` now. 2c9f3da

api/integrity.py
barborico: Removed. 2c9f3da

tests/test_constraint_propagation.py
barborico: Rewritten to state what the tests below require rather than what used to happen. 2c9f3da

Jeff-Rowell (commented):
Claude review is calling out duplicate code that already lives in [effective_constraint_options()](https://github.com/discord/access/blob/f0de3fd62e619d1f253d6f547bab3c8c977d8cc7/api/routers/_eager.py#L126-L158). Not sure if this is intentional, should we de-duplicate these?
1. [api/operations/constraints/check_for_self_add.py](https://github.com/discord/access/blob/f0de3fd62e619d1f253d6f547bab3c8c977d8cc7/api/operations/constraints/check_for_self_add.py#L40-L48)
2. [api/operations/constraints/check_for_reason.py](https://github.com/discord/access/blob/f0de3fd62e619d1f253d6f547bab3c8c977d8cc7/api/operations/constraints/check_for_reason.py#L42-L50)
3. [api/operations/modify_group_users.py](https://github.com/discord/access/blob/f0de3fd62e619d1f253d6f547bab3c8c977d8cc7/api/operations/modify_group_users.py#L75-L88)

Otherwise LGTM!

Jeff-Rowell (dismissed):

api/integrity.py
```diff
@@ -174,3 +178,55 @@ async def verify_and_fix_role_memberships(dry_run: bool = False) -> None:
                         # Remove user from okta group owners
                         # https://help.okta.com/en-us/Content/Topics/identity-governance/group-owner.htm
                         await okta.remove_owner_from_group(active_role_group_map.group_id, user_id)
+
+
+async def cap_role_memberships(dry_run: bool = False) -> int:
```
matthew-bass, line 183: Would it be cleaner to put this write path behind an operation, or do you think `api.integrity` is the right home for a direct database mutation and commit?

api/operations/modify_role_groups.py
```diff
@@ -655,6 +650,38 @@ async def execute(self) -> RoleGroup:
...
+            # `propagating_seconds_limit` states -- it does not re-derive it.
+            propagated_limits = [
+                await propagating_seconds_limit([g.id for g in groups_to_add], Tag.MEMBER_TIME_LIMIT_CONSTRAINT_KEY),
+                await propagating_seconds_limit(
+                    [g.id for g in owner_groups_to_add], Tag.OWNER_TIME_LIMIT_CONSTRAINT_KEY
+                ),
+            ]
+            seconds_limit = min((limit for limit in propagated_limits if limit is not None), default=None)
+            if seconds_limit is not None:
+                await limit_access_conferred_by_roles(
```
matthew-bass: Was sitting with a codex issue finding that seems reasonable, might take some extenuating circumstances  to trigger the race, but might be worthwhile to fix if it's easy to do

> The role association and derived grants have already been committed before this cap runs. If this step fails, the request errors but leaves the association durable while the role's existing memberships remain uncapped.

matthew-bass (commented):
couple small comments

one other nit i noticed is that the comments / docstrings are pretty long and expose a lot of the "incident/change history" that might be better suited to the pr description (or just cut)

api/operations/modify_role_groups.py
barborico: Seems easy enough to move them into the same transaction! Fixed in https://github.com/discord/access/pull/616/commits/ccf939f01fde78b58ed16d9e2c07a68a11480a9e.

api/integrity.py
barborico: Since this is intended to be an uncommon (likely one-time) repair rather than a standard operation, I think that this is the right home and matches precedent.

barborico:
> one other nit i noticed is that the comments / docstrings are pretty long and expose a lot of the "incident/change history" that might be better suited to the pr description (or just cut)

Took a pass at cleaning this up throughout the stack, and added some more steering to my CLAUDE.md. I'm not opposed to lengthy comments per se, but agree that they should clearly and succinctly reflect the present intent only.

matthew-bass (approved):

PR #617, opened 2026-09-01
Require propagation on tags that disallow self-adds

barborico:
> **Stacked on #616** (`brynna/propagate_tag_constraints_to_roles`). Review only the last commit; GitHub retargets this to `main` when #616 merges.

**Urgency**: 3 days

## Why propagation and self-add prohibitions can't be independent

The reason and time-limit constraints degrade gracefully when a tag stops reaching roles — they still bind on the direct path, and a role path still records *some* reason and *some* duration. `disallow_self_add_*` does not degrade, it **inverts**.

An owner of G blocked from adding themselves directly can add themselves to a role associated with G and arrive at exactly the same access by a fully supported path, with nothing recording that a separation-of-duties control was ever in play. So a tag carrying either self-add key must propagate.

## Enforcement

`validate_constraint_propagation` in `api/models/tag.py` rejects the combination at **every** tag write:

- `PUT /api/tags/{id}` validates the **merged** result, not the body. This is a partial update, so either half of the conflict may be the one arriving while the other is already stored — a `PUT` carrying only `propagate_to_roles: false` is coherent in isolation. It runs before the `setattr` loop so a rejected edit leaves the session unmutated.
- `CreateTag.__init__` validates too, so CLI and seed callers are held to the same invariant as the router.

The message states the invariant rather than which half the write happened to change, since either half may be the one arriving. It names *every* offending key — removing one of two would still leave the tag invalid.

## UI

- A **Propagate these constraints to roles?** toggle on the tag form, with a tooltip distinguishing it from disabling the tag (which turns off enforcement everywhere).
  image
- A sentence on the tag page saying whether constraints reach roles.
  image
- An **inline conflict message**, so an admin sees which restriction is at fault next to the control instead of bouncing off a 400 after submit.
  image
  image

<details>
<summary><h3>The `deps` bit</h3></summary>

The conflict rule lives on `propagateToRoles` but reads `ownerAdd`/`memberAdd`. React Hook Form re-runs a field's rules only when *that field* changes, so without help the message never appears when the restriction is what changed, and stays on screen after the admin fixes it. The two self-add toggles carry `validation={{deps: ['propagateToRoles']}}`, which makes RHF `trigger` the conflict rule after either of them validates.

Both prefill paths use `?? true`, not `Boolean(...)`: the field is optional in the generated type and the server default is `true`, so coercing `undefined` would show — and on save, silently write — the opposite of what an older tag actually does.
</details>

<details>

`propagationRules.ts` and `PropagationNoteView` are unit-tested. The form itself can't be: this repo's MUI style engine is `@mui/styled-engine-sc`, aliased in `vite.config.ts` for the app build only, and that alias doesn't reach vitest's resolution — importing a real MUI component blows up at module load. Hence the screenshots above, captured against a local backend.

The merged-partial-update paths were checked over HTTP:

| Request | Stored state | Result |
|---|---|---|
| `{"constraints": {"disallow_self_add_membership": true}}` | `propagate_to_roles: false` | **400** |
| `{"constraints": {...true}, "propagate_to_roles": true}` | — | **200** |
| `{"propagate_to_roles": false}` | constraint already set | **400** |

937 backend / 45 frontend passing; `ruff`, `ty`, and `tsc` clean.
</details>

### DIFF
--- a/api/routers/tags.py
+++ b/api/routers/tags.py
@@ -20,6 +20,7 @@ from api.operations import CreateTag, DeleteTag
 from fastapi_pagination.ext.sqlalchemy import apaginate
 
 from api.pagination import Page, validated
+from api.models.tag import validate_constraint_propagation
 from api.routers._eager import group_tag_map_options
 from api.schemas import (
     TagListItem,
@@ -158,6 +159,15 @@ async def put_tag(
         payload["constraints"] = {}
     if "description" in payload and payload["description"] is None:
         payload["description"] = ""
+    # Validate the merged result, not the body: this is a partial update, so
+    # either half of the conflict may be arriving now while the other is
+    # already stored. Runs before the setattr loop so a rejected edit leaves
+    # the session unmutated.
+    validate_constraint_propagation(
+        payload["constraints"] if "constraints" in payload else tag.constraints,
+        payload["propagate_to_roles"] if "propagate_to_roles" in payload else tag.propagate_to_roles,
+    )
+
     for key in ("name", "description", "constraints", "enabled", "propagate_to_roles"):
         if key in payload:
             setattr(tag, key, payload[key])
--- a/api/routers/tags.py
+++ b/api/routers/tags.py
@@ -123,8 +123,19 @@ async def put_tag(
 
     from api.operations import ModifyGroupsTimeLimit
 
+    # Locked for the rest of this transaction, because the validation below
+    # reads whichever half of the tag this request is not sending. Two updates
+    # arriving together -- one turning propagation off, one switching a
+    # self-add restriction on -- would each read the other's stale half, both
+    # pass, and store the pair neither is allowed to create. Released by the
+    # commit below.
     tag = (
-        await db.scalars(select(Tag).where(Tag.deleted_at.is_(None)).where(or_(Tag.id == tag_id, Tag.name == tag_id)))
+        await db.scalars(
+            select(Tag)
+            .where(Tag.deleted_at.is_(None))
+            .where(or_(Tag.id == tag_id, Tag.name == tag_id))
+            .with_for_update()
+        )
     ).first()
     if tag is None:
         raise HTTPException(404, "Not Found")
--- a/src/pages/tags/CreateUpdate.tsx
+++ b/src/pages/tags/CreateUpdate.tsx
@@ -15,6 +15,7 @@ import EditIcon from '@mui/icons-material/Edit';
 import FormControl from '@mui/material/FormControl';
 import Grid from '@mui/material/Grid';
 import IconButton from '@mui/material/IconButton';
+import Tooltip from '@mui/material/Tooltip';
 
 import {ToggleButtonGroupElement, FormContainer, TextFieldElement} from 'react-hook-form-mui';
 
@@ -27,6 +28,7 @@ import {
   TagByIdPutVariables,
 } from '../../api/apiComponents';
 import NumberInput from '../../components/NumberInput';
+import {MEMBER_SELF_ADD_LABEL, OWNER_SELF_ADD_LABEL, propagationConflictMessage} from './propagationRules';
 import {OktaUserDetail, TagDetail} from '../../api/apiSchemas';
 import {isAccessAdmin} from '../../authorization';
 import accessConfig, {requireDescriptions} from '../../config/accessConfig';
@@ -61,6 +63,7 @@ interface CreateTagForm {
   memberReason?: string;
   ownerAdd?: string;
   memberAdd?: string;
+  propagateToRoles: string;
 }
 
 interface TagDialogProps {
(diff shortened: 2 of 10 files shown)


src/pages/tags/CreateUpdate.tsx
```diff
@@ -356,12 +371,63 @@ function TagDialog(props: TagDialogProps) {
...
+              </FormControl>
+            </Grid>
+          </Grid>
+          <Grid container spacing={1}>
+            <Grid item xs={12}>
+              <FormControl fullWidth sx={{marginTop: '18px'}}>
+                <Tooltip
+                  title={
+                    'When yes, these constraints also apply to any role that is a member or owner of a group ' +
+                    "carrying this tag \u2014 the role's own members must satisfy the same time limits, reason " +
```
barborico:
```suggestion
                    "carrying this tag: The role's own members must satisfy the same time limits, reason " +
```

tests/test_tag_propagate_to_roles.py
```diff
@@ -94,3 +94,179 @@ async def test_post_tag_defaults_propagate_to_roles_when_absent(
...
     loaded = (await db.session.scalars(select(Tag).where(Tag.id == response.json()["id"]))).one()
     assert loaded.propagate_to_roles is True
+
+
+# --- Self-add constraints require propagation ------------------------------
+#
+# `disallow_self_add_*` is a separation-of-duties control, and unlike the
+# reason and time-limit constraints it does not merely weaken when propagation
+# is off -- it inverts to permitted. An owner of G blocked from adding
+# themselves directly can add themselves to a role associated with G and arrive
```
barborico:
```suggestion
# is off -- it inverts to permitted: An owner of G blocked from adding
# themselves directly could add themselves to a role associated with G and arrive
```

src/pages/tags/CreateUpdate.tsx
barborico: Applied. The `—` escape was an artifact of how I wrote the file; your version drops it entirely, which is better than the escape either way. 7a3c759

tests/test_tag_propagate_to_roles.py
barborico: Applied. 7a3c759

Jeff-Rowell (dismissed):

api/routers/tags.py
```diff
@@ -158,6 +159,15 @@ async def put_tag(
         payload["constraints"] = {}
     if "description" in payload and payload["description"] is None:
         payload["description"] = ""
+    # Validate the merged result, not the body: this is a partial update, so
+    # either half of the conflict may be arriving now while the other is
+    # already stored. Runs before the setattr loop so a rejected edit leaves
+    # the session unmutated.
+    validate_constraint_propagation(
```
matthew-bass, line 177: I think there's a race here where two concurrent requests could override the expected writes of the other.

> If there's two requests where one turns propagation off and another enables a `disallow_self_add_*` constraint, then they persist different changed columns and leave the forbidden combination stored. Can we make the validation and write atomic against concurrent updates?

Seems super rare in practice so considering a nit, but thought worth noting

src/pages/tags/CreateUpdate.tsx
```diff
@@ -335,12 +344,18 @@ function TagDialog(props: TagDialogProps) {
...
                   name="ownerAdd"
                   enforceAtLeastOneSelected
                   exclusive
                   required
+                  // The conflict rule lives on `propagateToRoles`, and RHF
+                  // re-runs a field's rules only when that field changes. `deps`
+                  // makes this toggle re-trigger it, so switching a self-add
+                  // restriction on surfaces the conflict and switching it back
+                  // off clears the message.
+                  rules={{deps: ['propagateToRoles']}}
```
matthew-bass: I think there might be some UX weirdness here with surfacing the errors when filling out this form if the propagation constrant is disabled and self-add restriction is enabled

If that's the case, it seems like the form should explain that this combination is invalid as soon as the second toggle is selected. 

> With `FormContainer`’s default `onSubmit` mode, these dependencies do not run the propagation validation before the first submit, so the user sees no conflict until clicking Create or Update. Could we validate this combination on change or blur and cover the pre-submit behavior with a form interaction test?

matthew-bass (dismissed):
Approving! Two comments I wouldn't consider blocking

also nit: similar to the other PR, seems like there might be some wordy comments/docstrings on this one that the guidance docs try to avoid having

api/routers/tags.py
barborico: Seems like a remote edge case but also a super easy fix: https://github.com/discord/access/pull/617/changes/deba10a7665c646a2bf1427154e82454ec16da76

src/pages/tags/CreateUpdate.tsx
barborico: Good call. Addressed in https://github.com/discord/access/pull/617/changes/3570ba7ab2e5228c0ecec6e1f4c84e2ded4f34c3.

matthew-bass (dismissed):

matthew-bass (dismissed):

PR #619, opened 2026-09-01
Report a group's effective constraints on its detail endpoint

barborico:
> **Stacked on #617** (`brynna/require_propagation_for_self_add_tags`). Review only the last commit; GitHub retargets this as the stack merges.

**Urgency**: 3 days

## Motivation

Make the constraints in force per group obvious in the UI, without having to replicate constraint coalescing logic in the frontend.

## What's Changing 

Which constraints bind a group, and why, was only ever answered *inside* enforcement. Anything wanting to show it had to re-derive coalescing and propagation for itself — and any drift between the two surfaces as the UI promising something enforcement does not do.

`effective_constraints(group)` in `api/models/tag.py` returns every constraint in force, with its coalesced value and the sources that produced it. It is built on the same `constraint_sources` walk enforcement already uses, so the two cannot disagree. `GET /api/groups/{id}` serves it as `effective_constraints`.

A source carries the tag plus **how the tag reached the group**:

| `origin` | Meaning |
|---|---|
| `direct` | The tag sits on this group |
| `app` | Inherited from the group's `App` |
| `member_association` | Propagated onto this role from a group it is a member of |
| `owner_association` | Propagated onto this role from a group it owns |

That distinction is the point: it separates a constraint the reader can lift by untagging this group from one they cannot.

Backend only. No UI reads it yet — that lands in the dialogs and the group-page panel further up the stack.

<details>
<summary><h3>Two shape decisions worth a look</h3></summary>

**`source_id` / `source_name`, not `source_group_*`.** An `app` origin's source is an `App`, not a group. Naming the fields after the origin keeps one shape across all four cases; a `direct` origin leaves both `None`, having no source beyond the group itself.

**Populated in the handler, not by `from_attributes`.** `effective_constraints` is derived, not an ORM attribute, so `validate_python(..., from_attributes=True)` cannot pick it up. `get_group` sets it after validating. `post_group`/`put_group` leave it at `[]` — the group page re-fetches the detail after a mutation, so the read is where it is populated.

Also worth flagging for the reviewer: the plural `effective_constraints` reads one relationship more than the singular `effective_constraint` — `OktaGroupTagMap.active_app_tag_mapping`, which is what tells `direct` from `app`. That path belongs to `group_tag_map_options()`, which detail routes already nest under `active_group_tags`, so `groups.py` gets it by composing the two helpers. `effective_constraint_options()`'s docstring now says so, since a caller that took only that helper and then asked for provenance would hit `raise_on_sql`.
</details>

<details>

Ten new tests. Six in `tests/test_effective_constraints.py` cover the helper directly: unset constraints omitted, association source reported, empty when nothing applies, and one per origin branch — `direct`, `app`, and a soft-deleted app.

That last one is deliberately loaded with `selectinload` rather than the production `joinedload`. `AppTagMap.active_app` is `innerjoin=True`, so under `joinedload` a deleted app collapses the whole `active_app_tag_mapping` chain to `None` and the source degrades to `direct` — meaning the unguarded `.name` read is not reachable through today's loader. `selectinload` issues a separate SELECT per relationship, so the mapping survives while `active_app` comes back `None`: exactly the shape the guard exists for, and the shape any future switch to `selectinload` would produce.

Four in `tests/test_effective_constraints_api.py` go through `GET /api/groups/{id}`: a propagated constraint appears, an untagged group reports `[]`, and the `direct` / `app` origins round-trip through the schema.

Adding `*effective_constraint_options()` to `DEFAULT_LOAD_OPTIONS` sits alongside an existing `selectinload(OktaGroup.active_group_tags)`, so it was worth proving no `Loader strategies ... conflict` at query-compile time — the 427 group-touching tests pass.

`src/api/apiSchemas.ts` regenerated (`ConstraintOrigin` as a literal union, the two new types, the field on all three group-detail variants). `apiComponents.ts` regenerated to no new content — a full regen also collapses a number of unrelated object literals that the committed file has expanded, which the `openapi-client-drift` check treats as drift — so it is reverted.

947 backend / 45 frontend passing; `ruff`, `ty`, and `tsc` clean.
</details>

### DIFF
--- a/api/models/tag.py
+++ b/api/models/tag.py
@@ -390,12 +390,30 @@ def constraint_source_clause(constraint_key: str, group: OktaGroup) -> str:
 
 def _constraint_entry(
     constraint_key: str, constraint: TagConstraint, sources: list[ConstraintSource]
-) -> dict[str, Any]:
-    """One constraint's coalesced value and the sources that produced it."""
+) -> Optional[dict[str, Any]]:
+    """One constraint's coalesced value and the sources that produced it.
+
+    A source setting a flag to `False` imposes nothing -- the tag form writes
+    all four boolean keys on every save, so most tags carry several -- and a
+    constraint every source turns off is not in force at all. Both are dropped,
+    so a reader is never told a restriction applies when it does not, and a tag
+    is never named as the reason for one it explicitly declines to impose.
+    `constraint_source_clause` filters the same way for the same reason.
+
+    Discriminated on `is not False` rather than on truthiness: only a boolean
+    flag can be switched off, and a falsy *number* is the tightest possible
+    limit rather than the absence of one.
+
+    Returns:
+        The entry, or None when nothing is left contributing.
+    """
+    contributing = [source for source in sources if source.value is not False]
+    if not contributing:
+        return None
     return {
         "constraint": constraint_key,
         "name": constraint.name,
-        "value": _fold(constraint, sources),
+        "value": _fold(constraint, contributing),
         "sources": [
             {
                 "tag_id": source.tag.id,
@@ -404,7 +422,7 @@ def _constraint_entry(
                 "source_id": source.source_id,
                 "source_name": source.source_name,
             }
-            for source in sources
+            for source in contributing
         ],
     }
 
@@ -425,8 +443,10 @@ def effective_constraints(group: OktaGroup) -> list[dict[str, Any]]:
         name), `value` (coalesced across every source under that constraint's
         own rule), and `sources`. A source names the tag, how it reached the
         group (`origin`), and the app or group it came from -- `source_id` and
-        `source_name`, both None for a `DIRECT` origin. Constraints nothing
-        sets are omitted, so an untagged group returns an empty list.
+        `source_name`, both None for a `DIRECT` origin. A constraint nothing
+        sets, and a flag every tag setting it turns off, are both omitted --
+        so an untagged group returns an empty list, and so does one whose only
+        tag declines every constraint.
 
     Raises:
         InvalidRequestError: If a relationship this reads was not eager-loaded.
@@ -437,7 +457,7 @@ def effective_constraints(group: OktaGroup) -> list[dict[str, Any]]:
     entries = []
     for constraint_key, constraint in Tag.CONSTRAINTS.items():
         sources = constraint_sources(constraint_key, group, include_provenance=True)
-        if not sources:
-            continue
-        entries.append(_constraint_entry(constraint_key, constraint, sources))
+        entry = _constraint_entry(constraint_key, constraint, sources)
+        if entry is not None:
+            entries.append(entry)
     return entries
--- a/api/models/tag.py
+++ b/api/models/tag.py
@@ -404,10 +404,22 @@ def _constraint_entry(
     flag can be switched off, and a falsy *number* is the tightest possible
     limit rather than the absence of one.
 
+    Sources are ordered by ascending value, then tag name, then source name.
+    Ascending value puts the tag that actually decided the answer first, which
+    for a `min` constraint is the shortest limit; the reader wants to know which
+    tag is binding them, not which happened to be traversed first. Flags all
+    tie -- every one left is `True` -- so they fall through to the name keys.
+    `bool` subclasses `int`, so one comparator covers both kinds of constraint.
+    The two name keys are what make the order independent of traversal, since
+    one tag can reach a role from two different groups and tie on the first two.
+
     Returns:
         The entry, or None when nothing is left contributing.
     """
-    contributing = [source for source in sources if source.value is not False]
+    contributing = sorted(
+        (source for source in sources if source.value is not False),
+        key=lambda source: (source.value, source.tag.name, source.source_name or ""),
+    )
     if not contributing:
         return None
     return {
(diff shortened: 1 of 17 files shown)


api/schemas/core_schemas.py
```diff
@@ -291,6 +292,29 @@ class RoleGroupMapDetail(BaseModel):
...
 
 
+class EffectiveConstraintSourceDetail(BaseModel):
+    model_config = ConfigDict(from_attributes=True)
+    tag_id: str
+    tag_name: str
+    origin: ConstraintOrigin
+    # Deliberately not `source_group_*`: an "app" origin's source is an App,
+    # not a group. Both fields are None for a "direct" origin, which has no
+    # source beyond the group itself.
```
barborico:
```suggestion
    # The site where the source tag was directly applied, from which the
    # constraint is inherited: an app, another group, or None if applied
    # directly.
```

api/models/tag.py
```diff
@@ -386,3 +386,41 @@ def constraint_source_clause(constraint_key: str, group: OktaGroup) -> str:
...
+
+
+def effective_constraints(group: OktaGroup) -> list[dict[str, Any]]:
+    """Every constraint in force on `group`, with its coalesced value and sources.
+
+    Backs the API response the UI reads, so display and enforcement answer from
+    the same code. Unlike `effective_constraint`, this reads
+    `OktaGroupTagMap.active_app_tag_mapping` for provenance -- callers must
+    eager-load it (`group_tag_map_options()` in `api/routers/_eager.py` does).
+    """
```
barborico, line 469: This docstring is missing some info; see https://github.com/discord/access/pull/612.

api/schemas/core_schemas.py
barborico: Applied. aa86187

api/models/tag.py
barborico: Right — public, so it owes Args/Returns/Raises. Added all three, following the `constraint_sources` docstring just above it. The Returns section now spells out the entry shape (which keys, `Tag.CONSTRAINTS` ordering, that unset constraints are omitted so an untagged group gives `[]`), and Raises carries the eager-load note that used to be loose prose in the summary — the interesting part being that provenance costs one relationship more than `effective_constraint` needs.

`_constraint_entry` is private, so it keeps its one-liner.

I also audited the rest of what this stack adds against that rule. Two real gaps, both in already-open PRs rather than here: the four `_time_limits.py` helpers (#615) and `propagating_seconds_limit` (#616) have summaries and Args but no Raises. Happy to fix them in place if you would rather not carry it — say the word and I will push to those branches. `coalesce_constraints` and `coalesce_ended_at` have no docstring at all, but both predate this work on `main`. aa86187

api/models/tag.py
```diff
@@ -386,3 +386,78 @@ def constraint_source_clause(constraint_key: str, group: OktaGroup) -> str:
...
+            {
+                "tag_id": source.tag.id,
+                "tag_name": source.tag.name,
+                "origin": source.origin,
+                "source_id": source.source_id,
+                "source_name": source.source_name,
+            }
+            for source in contributing
+        ],
+    }
```
Jeff-Rowell: The RFC design has the winning tag listed first in sources [here](https://app.notion.com/p/discordapp/Propagating-Access-SOX-constraints-from-tagged-groups-to-roles-3b3f46fd48aa80ecae80e202247e6524?source=copy_link#070e46a5718841ea869a828d25ded7a4). Are these sources sorted somewhere else before being displayed in the UI in a later PR in the stack?

Jeff-Rowell (commented):
Not sure how critical sorting the sources is, but seems like there might be a little drift from the RFC's desired design for how these constrain sources are presented in the UI

api/models/tag.py
barborico: Fixed in https://github.com/discord/access/pull/619/changes/081b973853d56d0163ab041c7fed4404d0e00bc8!

tests/test_effective_constraints.py
```diff
@@ -278,3 +285,242 @@ async def test_nothing_propagates_onto_an_unmanaged_role(db: Db) -> None:
...
+    tag name. The source name is the tiebreak that keeps the order stable."""
+    strict_group = OktaGroupFactory.build(name="Aardvark-Group")
+    lenient_group = OktaGroupFactory.build(name="Zebra-Group")
+    role = RoleGroupFactory.build()
+    tag = TagFactory.build(name="SOX", constraints={Tag.REQUIRE_MEMBER_REASON_CONSTRAINT_KEY: True})
+    db.session.add_all([strict_group, lenient_group, role, tag])
+    await db.session.commit()
+    for group in (lenient_group, strict_group):
+        db.session.add(OktaGroupTagMapFactory.build(group_id=group.id, tag_id=tag.id))
+        db.session.add(RoleGroupMap(group_id=group.id, role_group_id=role.id, is_owner=False))
```
matthew-bass: should this use `RoleGroupMapFactory`?

api/models/tag.py
```diff
@@ -386,3 +386,91 @@ def constraint_source_clause(constraint_key: str, group: OktaGroup) -> str:
...
+
+
+def effective_constraints(group: OktaGroup) -> list[dict[str, Any]]:
```
matthew-bass: anyway we can narrow this type further? seems like it might be worthwhile to do

api/models/tag.py
```diff
@@ -386,3 +386,91 @@ def constraint_source_clause(constraint_key: str, group: OktaGroup) -> str:
...
+
+    Raises:
+        InvalidRequestError: If a relationship this reads was not eager-loaded.
+            Provenance means this reads one relationship more than
+            `effective_constraint` does -- `OktaGroupTagMap.active_app_tag_mapping`,
+            supplied by `group_tag_map_options()` in `api/routers/_eager.py`.
+    """
+    entries = []
+    for constraint_key, constraint in Tag.CONSTRAINTS.items():
+        sources = constraint_sources(constraint_key, group, include_provenance=True)
```
matthew-bass, line 509: didn't realize Access could have "unmanaged groups", but here's a codex finding where it seems like the UI will show tags, but they wouldn't be enforced for unmanaged groups.

> Unmanaged groups skip tag enforcement, but this still reports their direct and App-inherited tags as effective. A previously tagged group that becomes unmanaged would therefore display controls Access no longer applies.

tests/test_effective_constraints.py
barborico: Fixed in https://github.com/discord/access/pull/619/commits/801da9fb38feea6c302e93e69d9caa9917de3b73

api/models/tag.py
barborico: Fixed in https://github.com/discord/access/pull/619/commits/0f6439209e7048b3466fc5d06ffe19e5ce078cec

api/models/tag.py
barborico: Fixed in https://github.com/discord/access/pull/619/commits/9ebe6c482f1eaf4c273cfb712bb8a0cffb97859a

matthew-bass (dismissed):

PR #620, opened 2026-09-01
Serve effective constraints for a set of groups or tags

barborico:
> **Stacked on #619** (`brynna/report_effective_constraints_on_groups`). Review only the last commit.

**Urgency**: 3 days

## Motivation

Delegate all constraint coalescing from the frontend to the API to avoid logical drift between pre- and post-submit validation.

## What's Changing 

`GET /api/constraints/effective` — one place to ask "what constraints apply here?".

The dialogs where someone picks an access duration or writes a justification each need this. Answering it client-side means transcribing `api.models.tag` into TypeScript — which tags are enabled, which reach a role through its associations, and the minimum or logical OR across them — once per dialog, and every copy is then free to drift from the rule it mirrors. #619 made the answer available for one group; this makes it available for a *selection*.

**Two modes, exactly one of which you name:**

| Mode | Question | Returns |
|---|---|---|
| `group_ids` | What binds the things I have selected? | `coalesced` **and** `by_group` |
| `tag_ids` | What would these tags impose? | `coalesced` only |

Tag mode covers the group that does not exist yet — approving a group request means choosing tags for a group about to be created, so there is nothing to compute propagation or app inheritance from and every source is `direct`.

Still backend only. The dialogs and the group-page panel come next.

<details>
<summary><h3>Design decisions worth a look</h3></summary>

**Both shapes, not one.** `coalesced` bounds a single shared control (a bulk dialog's one duration picker); `by_group` answers per row (which groups block a self-add, which picker options to disable). Neither derives from the other without re-implementing `coalesce` in the client, which is the duplication this endpoint exists to remove — so it returns both rather than making the caller fold.

**Both id lists, or neither, is a 400.** The two modes answer different questions; combining them would silently invent a third.

**Stale ids are dropped, not fatal.** A dialog's selection is built from a list render that may predate a deletion, and one stale id should not blank the answer for the rest. An entirely stale selection therefore yields an empty answer rather than an error.

**Capped at 200 ids.** The eager loads scale with the requested set, so it is bounded rather than left to the caller — well above any real dialog selection. Rejected by the query model, so it arrives as a validation error: 400 with an `errors` list, per this API's problem-detail contract.

**Not a field on the list endpoints.** `effective_constraints` needs the association and tag eager loads for every group it touches. Hanging it off a paginated list pays that for every row of every page render; here it is paid once per dialog interaction, for the ids actually selected.

**The handler's docstring is written for an API consumer, not a Python reader.** FastAPI publishes it as the endpoint's OpenAPI description, so it reaches the generated TypeScript client and any docs page. My first draft was the full Google-style block the #612 rule asks for, and the generated client came out carrying `db: Request-scoped session` and the private constant name `_EFFECTIVE_CONSTRAINTS_MAX_IDS` in its public JSDoc. Notes for a reader of the function are in comments instead, and the interface documentation for the actual work lives on the `api.models.tag` helpers, which is where another module imports from.
</details>

<details>

Eleven tests in `tests/test_effective_constraints_endpoint.py`: a group's own constraints; coalescing across groups while `by_group` keeps them separate; booleans folding by OR; constraints propagated onto a role; the propagation gate respected; unknown ids omitted; tag mode coalescing and ignoring disabled tags; and the three input-validation rejections (neither, both, over the cap).

`api/models/tag.py` gains `effective_constraints_across_groups` and `effective_constraints_for_tags` — both public, so both carry Args/Returns/Raises per #612. The former folds every group's sources under one `coalesce` rather than merging per-group answers; a tag reaching two selected groups appears once per group, since the reader needs to know which of their selections carries it.

**On the regenerated client**: `apiComponents.ts` is spliced, not wholesale. A full regen here also collapses eleven unrelated object literals (`/api/healthz`, `/api/requests/{id}`, and nine others) that the committed file has expanded. This is not a Node version difference; it reproduces on the same Node 24 CI uses. So the diff is the two new-content hunks only, leaving the file at **+137/−0**. `apiSchemas.ts` is a clean regen.

958 backend / 45 frontend passing; `ruff`, `ty`, `tsc`, and prettier clean.
</details>


api/routers/constraints.py
```diff
@@ -0,0 +1,106 @@
...
+    # written for a caller of the HTTP endpoint. Notes for a reader of this
+    # function belong in comments like this one, and the interface
+    # documentation for the work itself lives on the `api.models.tag` helpers.
+    if bool(q_args.group_ids) == bool(q_args.tag_ids):
+        raise HTTPException(400, "Provide exactly one of group_ids or tag_ids")
+
+    if q_args.tag_ids:
+        tags = list(
+            (await db.scalars(select(Tag).where(Tag.id.in_(q_args.tag_ids)).where(Tag.deleted_at.is_(None)))).all()
+        )
```
matthew-bass, line 78: Tag mode only evaluates the explicitly selected tags, so the AppGroup preview omits tags it will inherit from the selected App when the group is created. This can let the UI offer an option such as indefinite ownership that creation later silently coerces to the inherited limit. 

Could the UI show the inherited App tags so the UI applies the same effective constraints before the user actually submits?

barborico: The bug that you're describing is preexisting; constraints on new group requests are evaluated only against directly requested tags today and in #621. But what you're asking for should be a small bugfix that can go on top of this stack.

matthew-bass (dismissed):

api/routers/constraints.py
barborico: Fixed in https://github.com/discord/access/pull/645

matthew-bass (dismissed):

PR #621, opened 2026-09-01
Drive the access dialogs off the effective-constraints API

barborico:
> Stacked on #620. The base branch carries `GET /api/constraints/effective`; this PR is the frontend that uses it.

**Urgency**: 3 days
**Expected review effort**: HIGH - a lotta churn, and it's React which is not as much our forte

## Motivation

Delegate all constraint coalescing from the frontend to the API to avoid logical drift between pre- and post-submit validation.

## What's Changing 

Eight dialogs each decided for themselves what constrained a selection, by walking tags in TypeScript. For a role that meant reading its associated groups and folding their tags by hand — once per dialog, with no knowledge of `propagate_to_roles`. So the UI could offer a duration or waive a justification that the backend would then refuse, and each copy was free to drift from `api/models/tag.py` independently.

They now read what the API resolved, through a new `src/constraints.ts`. The tag walk is deleted from `helpers.tsx`.

| Dialog | Where the answer comes from |
|---|---|
| `groups/AddUsers`, `groups/AddRoles` | `effective_constraints` already on the group detail — no request of its own |
| `requests/Create`, `requests/Read`, `role_requests/Create`, `role_requests/Read` | the endpoint, one group |
| `groups/BulkRenewal`, `roles/BulkRenewal` | the endpoint, the whole selection |
| `group_requests/Read` | the endpoint in tag mode — the group does not exist yet |

<details>
<summary><h2>How the reader behaves, and why</h2></summary>

The answer now arrives over the network, so a dialog has to cope with not having it yet. Four things follow from that, and they are the parts most worth a careful read:

- **The gates assume a restriction until the answer is in hand.** `isReasonRequired` and `isSelfAddDisallowed` report `true` while a request is in flight, if it failed, or if the payload never carried an answer. A permissive default shown early invites an action the backend then refuses; the cost of being briefly over-strict is much lower.

- **Absent is not empty.** `[]` means "nothing applies"; `undefined` or `null` means nobody has said. Only `GET /api/groups/{id}` resolves `effective_constraints`: an audit group reference has no such field, and the create/update group responses send `null` (#619). An id the endpoint dropped, because the group no longer exists, gets no `by_group` entry. All three are unknown, not unrestricted.

- **A time limit cannot fail closed, so submit is disabled instead.** There is no safe stand-in for "how long", so dialogs offering a duration disable submit while the answer is unknown, and show `ConstraintsUnavailableAlert` if the request failed — queries do not retry, so that state otherwise persists with no explanation.

- **Selections over the endpoint's 200-id cap are split, not truncated.** A role can be associated with more groups than one request may name. Folding the batches back together is a batching concern rather than the tag coalescing this file exists to avoid: it combines answers the API already computed, and picks min or OR from the value's type, never from knowledge of what a constraint means.
</details>
<details>
<summary><h2>Bugs fixed</h2></summary>

### Array query parameters went out comma-joined

`resolveUrl` in `src/api/apiFetcher.ts` built its query string with the `URLSearchParams` record constructor, which stringifies each value. `group_ids: ['a','b']` therefore went out as one parameter named `a,b`; FastAPI read that as a single id, matched no group, and answered `{coalesced: [], by_group: {}}`.

`/api/constraints/effective` is the first endpoint in the generated client to take an array parameter, which is why this had never surfaced. A one-element array survives stringification intact, so every single-group dialog was correct while the multi-id ones were not:

| Caller | Consequence |
|---|---|
| `groups/BulkRenewal`, `roles/BulkRenewal` | offered Indefinite on limited groups, and waived a required justification |
| `roles/BulkRenewal` row marks | flagged *every* row as self-add restricted |
| `group_requests/Read` with two or more tags | unbounded ownership picker |

Worth noting which way each half failed. An empty `coalesced` is indistinguishable from "nothing is constrained", so the roll-up gates failed **open**; a missing `by_group` entry reads as unknown, so the per-row gates failed closed. `resolveUrl`'s parameter was typed `Record<string, string>`, so `tsc` had nothing to object to either. Arrays now expand to repeated keys.

### The approval forms defaulted to Indefinite

`requests/Read` and `role_requests/Read` own their form so the constraint effect can move the `until` field. `useForm` snapshots `defaultValues` on the mounting render, which here is the render where the request itself is still loading; the snapshot therefore came from an empty request and read as `indefinite`.

The effect only ever *narrowed*, so two of its three cases never fired:

- On a group with **no** time limit, nothing ran. Approving a 30-day request defaulted to indefinite access.
- On a group **with** one, where the request already fit inside it, nothing ran either. The field held an option the narrowed list no longer contained, so the picker rendered blank and submitted `indefinite`, which the backend clamped to the group maximum rather than to what was asked for.
- Only a request *longer* than the limit behaved, moving to the longest duration still on offer.

The effect now owns the starting value in both directions. This is the same duration-picker bug an earlier revision of this PR reported as fixed; that fix covered the narrowing half only.

The decision behind the write lives in `approvalUntilDefault` in `src/constraints.ts`, shared by both pages rather than duplicated. Neither dialog can be rendered under vitest, which is exactly how a half-fix passed review the first time; as a pure function it is directly testable.
</details>
<details>

**32 new frontend tests.** `constraints.test.ts` covers the pure readers and `approvalUntilDefault`. `constraintsHooks.test.tsx` drives the hooks against a stubbed client through a real `QueryClientProvider`: gates closed in flight, gates closed on failure, an absent list treated as unknown, an id the response did not answer for treated as unknown, no request issued for an empty selection (and not blocked by it), batching at 200 with the answers folded, and blocked while any one batch is still outstanding.

`apiFetcher.test.ts` asserts the URL the fetcher actually builds, through a stubbed `fetch`: an array becomes repeated keys, a one-element array is unchanged, scalars are untouched, an empty array is omitted rather than sent empty, and path parameters still substitute. The hook tests mock `./api/apiComponents` wholesale, so they never exercise this layer; that is why it needs its own file.

Fail-closed logic passes just as well when it does nothing, so the tests are mutation-checked: removing the fail-closed guard fails 3, breaking either half of the batch fold fails 1 each, reverting the query-string fix fails 2, and making `approvalUntilDefault` narrow-only fails 3. The batching fixture has both batches report both keys, so every entry must be folded rather than merely inserted.

**Verified in a browser**, since this repo's MUI style engine cannot be loaded under vitest and the dialogs themselves are therefore not unit-testable:

- On a group with a 3-day owner limit, `GET /api/constraints/effective` fires on selection, the caption reads *"Ownership of this group is limited to 3 days"*, and **For how long?** settles on *12 Hours* rather than the unconstrained default.
- On an approval page for a request that asked for indefinite access to a group limited to 7 days, **For how long?** settles on *5 Days*.
- On an approval page for a 30-day request against a group with no limit, **For how long?** settles on *30 Days* rather than *Indefinite*.
- `groups/AddUsers` renders the limit with no constraints request at all, reading the value off the group detail.

944 backend / 77 frontend passing; `ruff`, `ty`, `tsc`, and prettier clean.
</details>


src/constraints.ts
```diff
@@ -0,0 +1,189 @@
...
+
+// Seconds, or null when no limit applies. Compared against `undefined` rather
+// than tested for truthiness: a `0`-second limit is a real constraint, and
+// treating it as absent would offer an unbounded duration.
+export function effectiveTimeLimit(constraints: Constraints, owner: boolean): number | null {
+  const value = valueOf(constraints, side(TIME_LIMIT_KEYS, owner));
+  return typeof value === 'number' ? value : null;
+}
+
+export function effectiveRequiredReason(constraints: Constraints, owner: boolean): boolean {
```
barborico: This function name does not make it clear that a boolean is returned, or what "owner" means.
```suggestion
export function isReasonRequired(constraints: Constraints, isOwner: boolean): boolean {
```
Same with `effectiveOwnerCantAddSelf` below.

src/pages/groups/AddUsers.tsx
```diff
@@ -100,47 +85,14 @@ function AddUsersDialog(props: AddUsersDialogProps) {
...
-      ownerCantAddSelfGroups(active_groups_owners ?? [], true);
-  }
+  // `effective_constraints` is resolved server-side by the same code that
+  // enforces these: a role's own tags plus anything reaching it through its
+  // associations, gated on `propagate_to_roles`. It arrives on the group the
+  // page already fetched, so reading it costs no request of its own.
+  const constraints = props.group.effective_constraints;
+  const timeLimit = effectiveTimeLimit(constraints, props.owner); // in seconds
+  const reason = effectiveRequiredReason(constraints, props.owner);
+  const disallow_owner_add = effectiveOwnerCantAddSelf(constraints, props.owner);
```
barborico: We should be using camelCase, not snake_case.

src/constraints.ts
barborico: Applied, along with the parameter rename. The two predicates are now `isReasonRequired(constraints, isOwner)` and `isSelfAddDisallowed(constraints, isOwner)`, and the reader's methods follow — `reader.isReasonRequired(true)`, `reader.forGroup(id).isSelfAddDisallowed(false)`.

Left `effectiveTimeLimit` under its own name since it returns `number | null` rather than a boolean, but its parameter is `isOwner` too now. 41aa078

src/pages/groups/AddUsers.tsx
barborico: Fixed — `disallow_owner_add` is now `selfAddDisallowed`.

Three snake_case locals in `roles/BulkRenewal.tsx` are untouched by this PR (`role_memberships`, `groups_cant_add_self_owner`/`_member`, `display_owner_add_constraint`); renaming them means touching every usage and would bury the real change here, so I have left them and filed the sweep separately. Say the word if you would rather have it in this PR. 41aa078

src/pages/group_requests/Read.tsx
```diff
@@ -188,7 +201,12 @@ export default function ReadGroupRequest() {
   const [tagSearchInput, setTagSearchInput] = React.useState('');
   const [ownershipUntil, setOwnershipUntil] = React.useState<string | null>(null);
 
-  const ownershipTimeLimit = React.useMemo<number | null>(() => minTagTime(selectedTags, true), [selectedTags]);
+  // The group being tagged does not exist yet, so there is no id to resolve
+  // constraints against — only the tags the approver has picked. That is what
+  // the endpoint's tag mode is for, and it keeps the coalescing on the server
+  // here too.
+  const tagConstraints = useConstraintsForTags(selectedTags.map((tag) => tag.id));
```
matthew-bass, line 232: Requested tags are resolved through only the first 100 global tag records before reaching this state. If a request has multiple tags and only some are on that first global page page, `selectedTags` becomes a non-empty partial list, so approval silently drops the missing requested tags.

src/pages/group_requests/Read.tsx
```diff
@@ -91,20 +93,31 @@ const ROLE_GROUP_PREFIX = 'Role-';
...
+        ? [[], accessConfig.DEFAULT_ACCESS_TIME]
+        : [UNTIL_OPTIONS, accessConfig.DEFAULT_ACCESS_TIME];
     }
     const [lastId, filtered] = filterUntilLabels(ownershipTimeLimit);
     return [filtered, lastId];
-  }, [ownershipTimeLimit]);
+  }, [ownershipTimeLimit, constraintsBlocked]);
 
   React.useEffect(() => {
     if (ownershipUntil == null || ownershipUntil === 'indefinite' || ownershipUntil === 'custom') return;
```
matthew-bass: I tested this locally with a local access stack and it seemed to be true, but considering a nit because i think it's blank anyways and the backend would correctly enforce whatever ownership limit is needed by the tag.

> If the request starts with indefinite ownership and the approver selects a tag with an ownership limit, the dropdown removes Indefinite but this leaves the form value unchanged. Approval still submits indefinite ownership, and the backend silently shortens it instead of the UI showing what will actually be created.

src/pages/groups/BulkRenewal.tsx
```diff
@@ -212,151 +205,83 @@ function BulkRenewalDialog(props: BulkRenewalDialogProps) {
...
+    setLabels(
+      Object.entries(Object.assign({}, filteredUntil, {custom: 'Custom'})).map(([id, label]) => ({
+        id: id,
+        label: label,
+      })),
+    );
+    // `until` is read but deliberately not a dependency: this reacts to the
+    // limit changing, not to the user picking a duration.
+    // eslint-disable-next-line react-hooks/exhaustive-deps
+  }, [timeLimit]);
```
matthew-bass: seems like there's an possible UI bug here where if you switch from a "constrained" time row to a non constrained group, the duration menu remains incorrectly filtered

> After switching from constrained rows to unconstrained rows, `timeLimit` is null both while loading and after success. Only `constraintsBlocked` changes, so this effect does not rerun and the duration menu can remain incorrectly narrowed. The role-renewal dialog has the same behavior.

barborico: Fixed in https://github.com/discord/access/pull/621/commits/f14d1a4840447de5673edd005e13ee8b292bb49d

src/pages/group_requests/Read.tsx
barborico: Good find. Looks like this was a preexisting bug, although probably unlikely to hit unless people have created a lot of tags as semantic labels (vs applying constraints). Fixed in https://github.com/discord/access/pull/621/commits/f50ccd4673a0d29189fad038516cdeb240731b67.

src/pages/group_requests/Read.tsx
barborico: Fixed in https://github.com/discord/access/pull/621/commits/925a04ae03c04672319aa6edd56c7a9e83f85bf2

matthew-bass (dismissed):
LOOKS GOOD TO ME

PR #626, opened 2026-09-02
Record both halves of why a role-derived grant exists

barborico:
**Urgency**: 5 days
**Expected review effort**: LOW, mostly tests

## Motivation

Create a more robust audit trail.

When user U has access to group G because U is a member of role R and R is associated with G, the `OktaUserGroupMember` row materialized in G is not the record of a decision anyone made about U and G. It exists because of **two** decisions: someone put U in R, and someone attached R to G.

The stored `created_reason` carried only one of them — and which one depended on which decision happened second. `ModifyRoleGroups` wrote why the role was attached; `ModifyGroupUsers` wrote why the user joined the role. So a reader of G's audit log saw half the justification, with nothing marking it as a half, and no way to tell which half they were looking at.

## What's Changing

Both halves are recorded, on their own labelled lines:

```
User in role because: needs ledger access for the Q3 close
Role {in|owns} group because: finance job function covers payment config
```

Three sites materialize these rows — `ModifyGroupUsers`, `ModifyRoleGroups`, and `verify_and_fix_role_memberships` — and each holds a *different* half while being able to reach the other. Composing in one helper is what keeps the stored shape identical across all three, including repaired rows, which previously defaulted to blank.

image

<details>
<summary><h2>Details worth a look</h2></summary>

- **A missing half still gets its label.** `(no reason given)` stands in, rather than dropping the line. Otherwise a one-sided reason is indistinguishable from a row written before this existed, which is exactly the ambiguity being fixed.
- **Both halves blank produces an empty string**, not two placeholder lines — nothing was recorded, and saying so twice is noise.
- **Truncation is budget-aware.** The limit is read off `OktaUserGroupMember.created_reason` rather than restated, so widening the column widens what fits without a second edit. Each half gets whatever slack the other leaves, so a short half never forces a long one into an even split; only when both exceed their share is the budget divided.
- The two frontend components that render a stored reason handle the multi-line shape.

</details>

<details>

`tests/test_role_derived_reason.py` covers the composition directly and through all three writers: both halves present, either half missing, both missing, the exact stored shape, and the truncation cases including the asymmetric-slack behaviour and the boundary where both halves exceed their share.

921 backend / 38 frontend passing; `ruff`, `ty`, `tsc`, and prettier clean.

</details>

### DIFF
--- a/api/operations/_derived_reason.py
+++ b/api/operations/_derived_reason.py
@@ -1,0 +1,91 @@
+"""The two-part justification behind a role-derived grant.
+
+When user U has access to group G because U is a member of role R and R is
+associated with G, the `OktaUserGroupMember` materialized in G is not the
+record of a decision anyone made about G directly. It exists because of two
+separate decisions: someone put U in R, and someone attached R to G. Either one
+alone is half the story, and a reader of G's audit log could not tell which
+half they were looking at.
+
+Each of the three sites that materializes such a row holds one half directly
+and can reach the other, but each holds a *different* half -- `ModifyRoleGroups`
+knows why the role was attached, `ModifyGroupUsers` knows why the user joined
+the role, and `api.integrity` is repairing rows it did not create. Composing
+here keeps the stored shape identical across all three.
+"""
+
+from __future__ import annotations
+
+from typing import Optional
+
+from api.models import OktaUserGroupMember
+
+USER_IN_ROLE_PREFIX = "User in role because: "
+ROLE_IN_GROUP_PREFIX = "Role in group because: "
+
+#: Stands in for a half that was never recorded. A blank half is itself worth
+#: seeing -- dropping the label instead would make a one-sided reason
+#: indistinguishable from a row written before this composition existed.
+MISSING_REASON_PLACEHOLDER = "(no reason given)"
+
+_ELLIPSIS = "…"
+
+# Read off the column rather than restated, so widening `created_reason` widens
+# what fits here without a second edit. `TypeEngine.length` is Optional in
+# SQLAlchemy's typing and the column declares `Unicode(1024)`, so the fallback
+# is unreachable in practice -- it is there so an unbounded column would not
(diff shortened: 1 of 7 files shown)


api/operations/_derived_reason.py
```diff
@@ -0,0 +1,91 @@
...
+"""
+
+from __future__ import annotations
+
+from typing import Optional
+
+from api.models import OktaUserGroupMember
+
+USER_IN_ROLE_PREFIX = "User in role because: "
+ROLE_IN_GROUP_PREFIX = "Role in group because: "
```
barborico, line 24: This doesn't read right when the role *owns* the group. Let's set it to "Role owns group because: " in that case.

barborico: Fixed in 5f5869e — `Role owns group because: ` when the association makes the role an owner.

The side is taken from the `RoleGroupMap` at each of the four sites that compose a reason, since that is what decides whether the materialized row is a membership or an ownership. The user half stays `User in role because: ` in both cases: only a role's *members* receive what it confers, so that half is always a membership.

There was already a test for the ownership branch and it was pinning the wrong copy — it now asserts the new label, plus two unit tests: one for the label selection, one for the truncation budget, since the owner label is longer and a composition that just fits as a membership has to trim as an ownership.

api/operations/_derived_reason.py
```diff
@@ -0,0 +1,104 @@
+"""The two-part justification behind a role-derived grant.
+
+When user U has access to group G because U is a member of role R and R is
+associated with G, the `OktaUserGroupMember` materialized in G is not the
+record of a decision anyone made about G directly. It exists because of two
```
barborico:
```suggestion
record of a decision anyone made about U and G directly. It exists because of two
```

Jeff-Rowell (approved):

somethingnew2-0 (approved):

PR #627, opened 2026-09-02
Explain each tag constraint, and what propagation does to it

barborico:
> Stacked on #622. Last PR in this series.

**Urgency**: none at all
**Expected review effort**: MEDIUM - the code diff is massive and mostly React (ew), but toodling around in the app locally is probably a better way to judge if you like it than reading the code 

## Motivation

`propagate_to_roles` gave the tag form a seventh toggle whose effect on the other six was left to the reader to work out. Only the toggle itself carried a tooltip.

Review feedback on #617:

> I do agree that any new toggles we add will potentially provide opportunity for confusion, so we should try to do whatever is necessary in the dialog and documentation to reduce that. It could be that we add some dynamic helper text below each of the other constraints describing the behavior of the tag based on when the propagate to roles behavior is checked or unchecked.

## What's changing

Each constraint now carries help text, and a front end module standardizes the copy. The **form** describes both cases, so an admin choosing whether to propagate sees what it means for the constraint in front of them. The **tag page** describes only the behaviour in force, read off that tag's own propagation setting and value.

Two consequences the copy expresses:
- an **owner** limit reaches roles that **own** the group and lands on those roles' **membership**, since membership of the role is what confers the ownership
- the self-add pair requires propagation

### Tag edit view

Propagation must be on when self-adds are prohibited:
image

Example constraint help text:
image

Disallow self-adds is blocked when propagation is off:
image

A blast radius warning shows when tightening constraints on an in-use tag:
image

### Tag read view

With constraints and some help text:
image

With no constraints:
image

### Effective constraints panel

Help text in the constraint view:
image

### Home page docs

image

## Impact beyond the copy

- **The two views now name constraints the same way.** They disagreed on all six (`Member Time Limit` vs `Member time limit`, and so on), which is its own source of confusion. The self-add labels also become *"Disallow adding oneself as…"*: adding yourself to a role requires owning the **role**, so the restriction reaches people who own neither it nor the tagged group, and naming an owner of the tagged group was simply wrong.
- **The tag page lists only constraints in force.** The form writes all four boolean keys on every save, so a typical tag showed three or four rows reading "No". Those are not constraints; a tag that has none drops the table entirely and says so in a sentence.
- **Time limits in the Value column are rounded.** A one-hour limit read `0.041666666666666664 days`.
- The unused constraint name/description in the backend were removed.

### DIFF
--- a/src/pages/tags/ConstraintHelpText.tsx
+++ b/src/pages/tags/ConstraintHelpText.tsx
@@ -1,0 +1,32 @@
+import React from 'react';
+
+import Box from '@mui/material/Box';
+import Typography from '@mui/material/Typography';
+
+import type {HelpParagraph} from './constraintHelp';
+
+/**
+ * Renders constraint help text inside a `Tooltip`.
+ *
+ * A paragraph's `lead` is emphasised and its `text` is not, so the two
+ * propagation cases stand out from the prose around them. Split rather than
+ * marked up because MUI's `title` takes a ReactNode and would render markdown
+ * literally; keeping the copy as data also lets `constraintHelp.ts` stay a pure
+ * module that vitest can test. Same approach as `PropagationNoteView`.
+ */
+export default function ConstraintHelpText({paragraphs}: {paragraphs: HelpParagraph[]}) {
+  return (
+    <>
+      {paragraphs.map((paragraph, index) => (
+        <Typography key={index} variant="body2" sx={{marginTop: index === 0 ? 0 : '8px'}}>
+          {paragraph.lead != null && (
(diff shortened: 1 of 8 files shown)


somethingnew2-0:
I feel like Tooltips are a bit difficult to discover and maybe not the correct component to display help text. I wonder if Claude Design could help come up with a more intuitive layout that also allows us to display dynamic help text in the dialog? I think the "Propagate to roles" toggle should probably go above all the other toggles since it affects their behavior? Maybe it should be above the "Optional Constraints" section as propagating to roles is not necessarily a constraint, but a modifier of the constraints?

PR #628, opened 2026-09-02
[api] Reduce Google group lifecycle alert noise

matthew-bass:
## Summary

- retry transient Cloud Identity group API failures with randomized exponential backoff
- parameterize actionable reconciliation errors so recurring failures can share a Sentry issue
- remove the plugin's redundant generic exception event while preserving the actionable error

`HttpRequest.execute()` defaults to a single attempt when `num_retries` is omitted. The Cloud Identity client now uses a request builder that gives every generated request three retries by default, while individual operations remain unaware of the policy. The shared Access dispatcher still logs raised hook failures, and failures still count toward the sync command's non-zero exit status. Further aggregation and Discord notification policy remain in Sentry rather than introducing reporting state into the plugin interface.

## Testing

- `uv run pytest` (923 passed)
- `uv run ty check .`
- `uv run ruff check .`
- `uv run ruff format --check .`

### DIFF
--- a/examples/plugins/app_group_lifecycle_google/plugin.py
+++ b/examples/plugins/app_group_lifecycle_google/plugin.py
@@ -24,6 +24,7 @@ from api.plugins.app_group_lifecycle import (
     AppGroupLifecyclePluginStatusProperty,
     DanglingPushMappingError,
     MissingOktaTargetError,
+    ReportedPluginError,
     UnresolvableOktaTargetError,
     hookimpl,
 )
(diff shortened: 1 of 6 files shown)


examples/plugins/app_group_lifecycle_google/plugin.py
somethingnew2-0: Instead of configuring retries on the individual request, should we configure it globally in the Google API client?
```suggestion
        self._groups_api = build("cloudidentity", "v1", credentials=credentials, num_retries=GOOGLE_API_NUM_RETRIES).groups()
```

barborico (dismissed):
+1 to Peter's comment which IMO would be a non-blocking improvement.

examples/plugins/app_group_lifecycle_google/plugin.py
matthew-bass: (codex initially posted this)

But looks like there's some nuance with that code where it doesn't set a global retry count for all requests from the client.

It looks like doing it that way only sets the retries when "fetching the API call descriptions to construct the client"

To get around this, LLM updated the client to use a custom requestBuilder that defaults to three retries

barborico (approved):
LGTM! Probably worth adding some agent steering in a follow on PR about using format strings for logging instead of f-strings.

PR #635, opened 2026-09-08
Render only the first paragraph in inline markdown descriptions

barborico:
# Summary
`MarkdownDescription`'s `inline` mode (list-view table cells) renders only the first paragraph of a description now, appending ` ...` when more follows. Previously, it ran together multiple paragraphs on a single line.

**Urgency**: none

# Motivation
Nothing in the app inherently produces two-paragraph group descriptions today, although it's possible. A follow-up PR lets app owner groups carry a second paragraph with non-boilerplate description, which makes this behavior more salient.

<details>

The fix truncates the *source markdown* to its first paragraph before handing it to `ReactMarkdown`, rather than adding `p` to the allowed list, which would break the single-line clamp inline mode depends on.

`firstParagraph` is exported as a pure string function and unit-tested on its own; the component just calls it. Two details are load-bearing and are pinned by tests:

- **Only whole blank lines are stripped from the front.** A `trimStart()` here would be wrong: four leading spaces are an indented markdown code block, and this helper truncates source markdown rather than rendering it. A test asserts the indentation survives, and it was mutation-checked — it fails if the regex is replaced with `trimStart()`.
- **A description that opens with a blank line** would otherwise produce a bare ` ...` with no visible text, so leading blank lines are skipped before the paragraph break is located.

CRLF is normalized first, so a blank line submitted by a browser textarea counts as a paragraph break.

Block mode is untouched and still renders every paragraph.
</details>

<details>

- Full frontend suite green at this branch's head: **9 files, 56 tests passed** (`npm test -- --run`).
- `npm run tsc -- --noEmit` clean.
- 12 unit tests on `firstParagraph` plus 3 rendering tests covering inline truncation, the single-paragraph case (no spurious ellipsis), and block mode still rendering every paragraph.
- The indentation-preservation test was verified to discriminate: substituting `trimStart()` makes it fail, and only it, then reverting restores green.

TODO: no screenshots attached; the visible effect is confined to list-view cells for descriptions that have more than one paragraph, which no current data produces.
</details>

# Guidance for Reviewers
Review file-by-file; the three commits are a helper, its wiring, and a follow-up regression test, and only the final state is interesting.

The judgement worth a second opinion is the choice to truncate the markdown source rather than allow `p` in `INLINE_ELEMENTS`. Allowing `p` would render both paragraphs and let the existing `WebkitLineClamp: 1` hide the overflow, which is arguably simpler, but the clamp then depends on block layout inside a table cell and the ellipsis would be the browser's rather than an explicit signal that content was dropped.

### DIFF
--- a/src/components/MarkdownDescription.test.tsx
+++ b/src/components/MarkdownDescription.test.tsx
@@ -38,6 +38,13 @@ describe('firstParagraph', () => {
   it('ignores leading blank lines and still marks that more follows', () => {
     expect(firstParagraph('\n\nSecond para\n\nThird')).toBe('Second para ...');
   });
+
+  it('preserves leading indentation, which is meaningful markdown', () => {
+    // Only blank lines are stripped. Leading spaces on a non-blank line are
+    // meaningful in markdown (an indented code block). A `trimStart()` here would
+    // incorrectly remove them.
+    expect(firstParagraph('    curl -X GET /api\n\nSee the docs')).toBe('    curl -X GET /api ...');
+  });
 });
 
 describe('MarkdownDescription inline rendering', () => {
--- a/src/components/MarkdownDescription.test.tsx
+++ b/src/components/MarkdownDescription.test.tsx
@@ -1,0 +1,32 @@
+import {describe, it, expect} from 'vitest';
+
+import {firstParagraph} from './MarkdownDescription';
+
+describe('firstParagraph', () => {
+  it('returns a single-paragraph description unchanged, with no ellipsis', () => {
+    expect(firstParagraph('Owners of the Zendesk application')).toBe('Owners of the Zendesk application');
+  });
+
+  it('keeps only the first paragraph and marks that more follows', () => {
+    expect(firstParagraph('Owners of the Zendesk application\n\nAlso grants billing access')).toBe(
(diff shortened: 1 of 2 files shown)


barborico:
Closing. This was a prerequisite for #637, which is also closed in favour of a smaller design that no longer depends on it.

Two corrections worth recording for anyone who revisits this:

1. The bug is milder than this PR described. Unwrapping `<p>` leaves a newline between the paragraphs, which HTML collapses to a space, so a two-paragraph description renders as `Owners of the Foo application Also grants billing access` — a space-joined run-on, not the letter-jammed text claimed above.
2. The stated rationale for truncating the source rather than allowing `p` doesn't hold. Measured in a browser: with `p` allowed and `margin: 0`, a short first paragraph followed by a second still gets the browser's ellipsis, and every cell renders at an identical height. The two approaches are behaviourally equivalent, and the literal ` ...` here is clamped away whenever the first paragraph fills the line.

Multi-paragraph descriptions are already possible for every other group type, so this remains a real but pre-existing cosmetic issue. If it's worth fixing, the minimal fix is adding `'p'` to `INLINE_ELEMENTS` with a margin reset — less code than this PR.

PR #636, opened 2026-09-08
Dedupe approver lookups so a double-owner gets one notification, not two

Jeff-Rowell:
# Summary
A group/app owner who holds two concurrent active ownership rows for the same group (a direct grant plus one granted via a role mapping) was being returned twice by the approver lookup helpers, so they received duplicate DMs for a single access request.
**Urgency**: 3

# Motivation
Reported in an internal thread: a user's notification count climbed rapidly, with pairs of identical "has requested membership to [App-X]" DMs for the same request. Traced to `get_group_managers`, `get_app_managers`, and `get_access_owners` returning the same `OktaUser` more than once when that user has multiple active `OktaUserGroupMember` rows granting ownership of the same group (e.g. one direct row plus one via a role mapping) — a state the data model explicitly allows and never collapses (see the access-grant-precedence rules). `CreateAccessRequest` passes that list straight through to the notification hook with no dedup, so the notification plugin sends one message per list entry.

# Description of Changes
- Added `.distinct()` to the `OktaUser` queries in `get_group_managers` (`api/models/okta_group.py`) and `get_app_managers`/`get_access_owners` (`api/models/app_group.py`) so a user with multiple active ownership rows for the same group is returned once.
- Added regression tests covering all three lookups with a user granted ownership both directly and via a role.

# Validation of Changes
- Added `tests/test_group_owner_lookups_dedup.py`, verified it fails on `main` (each lookup returns the user twice) and passes with the fix.
- Full test suite passes (`uv run pytest`).
- `ruff check`, `ruff format --check`, and `ty check` pass on the changed files.

# Guidance for Reviewers
Single logical change across two files plus one test file — fine to review as one diff. The core question is whether `.distinct()` on `select(OktaUser)...` is the right fix point vs. deduping at each call site (e.g. `create_access_request.py`); I chose the query-level fix so every current and future caller of these three helpers gets the same guarantee.

### DIFF
--- a/api/models/app_group.py
+++ b/api/models/app_group.py
@@ -17,8 +17,13 @@ async def get_app_managers(app_id: str) -> List[OktaUser]:
 
     if ((await db.session.scalar(select(func.count()).select_from(owner_app_groups_stmt.subquery()))) or 0) > 0:
         owner_app_group_ids = [ag.id for ag in await db.session.scalars(owner_app_groups_stmt)]
+        # A user can hold two concurrent active OktaUserGroupMember owner rows
+        # for the same app-owner group at once (a direct grant plus one via a
+        # role mapping, per the access-grant-precedence rules) -- distinct()
+        # so such a user is returned once, not once per row.
         result = await db.session.scalars(
             select(OktaUser)
+            .distinct()
             .join(OktaUser.all_group_memberships_and_ownerships)
             .where(OktaUserGroupMember.group_id.in_(owner_app_group_ids))
             .where(OktaUserGroupMember.is_owner.is_(True))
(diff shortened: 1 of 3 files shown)


api/models/app_group.py
```diff
@@ -17,8 +17,13 @@ async def get_app_managers(app_id: str) -> List[OktaUser]:
 
     if ((await db.session.scalar(select(func.count()).select_from(owner_app_groups_stmt.subquery()))) or 0) > 0:
         owner_app_group_ids = [ag.id for ag in await db.session.scalars(owner_app_groups_stmt)]
+        # A user can hold two concurrent active OktaUserGroupMember owner rows
+        # for the same app-owner group at once (a direct grant plus one via a
+        # role mapping, per the access-grant-precedence rules) -- distinct()
```
barborico, line 22: Multiple roles can grant ownership, so it's possible for there to be more than two total rows here.
```suggestion
        # A user can hold multiple concurrent active OktaUserGroupMember owner
        # rows for the same app-owner group at once (a direct grant plus any
        # number of role-based grants) -- distinct()
```

barborico (approved):
Thanks for jumping on this! The root cause and fix look solid. Just one suggested tweak to the comments/docstrings.

PR #637, opened 2026-09-08
Allow appending a custom description to app owner groups

barborico:
# Summary
An app owner group's description can now carry free text below the fixed `Owners of the {app name} application` line, rendering as a second markdown paragraph. No new column, no migration, no API field.

**Urgency**: 5 days

# Motivation
Owner groups sometimes carry app-level permissions beyond "owns this app", and there was nowhere to record that: the description was composed at app creation and rewritten on every app rename, and `PUT /api/groups/{id}` rejected every non-tag edit for owner groups.

Stacked on #635, which fixes inline markdown rendering. That PR must merge first: this one is what starts producing two-paragraph descriptions, and without it list views render the two paragraphs run together.

<details>

The composed string stays in `okta_group.description` — it is already stored, already synced to Okta, and already rendered everywhere. Two ideas carry the design:

**One invariant, enforced in one place.** A description is either exactly the base line, or the base line + `\n\n` + free text. The check lives in `ModifyGroupDetails.execute()` rather than a router, because every writer already routes through it — `put_group` and the plugin `set_group_description` capability both do — so one site covers the API, plugins, and any future caller, and lands as a 400 through the existing `except ValueError`.

**One reseat rule, shared by both readers.** A description starting with the expected base line splits into base + remainder; one that does *not* becomes the remainder in its entirety. Both the edit form (seeding its field) and the app-rename path (recomposing) use it, so a description written before this convention is repaired rather than discarded, and repaired identically by either.

The rename path refuses with a 400 when recomposition would exceed the 1024-character column, checked in a **pre-flight before its loop**: `ModifyGroupDetails` pushes to Okta and commits per call, and the app name is assigned before the loop, so a check inside it would leave earlier groups renamed and the app half-renamed. `CreateApp` carries the same guard for the same reason.

**On the frontend duplicating a backend format string.** The form builds the expected prefix from the app name rather than splitting on the first blank line. Splitting on structure cannot tell a conforming description from a divergent one, so it would show wrong text as immutable and leave a divergent description unrepairable through the UI. The duplication is contained by a test that reads the TypeScript source and asserts it agrees with the Python — for both the format string and the length budget. This is also not unprecedented: `CreateUpdate.tsx` already keeps `APP_GROUP_PREFIX` and `ROLE_GROUP_PREFIX` as client-side copies.

Nothing changes for non-owner app groups, role groups, or plain Okta groups; `name`, `type`, `app_id`, and plugin config remain rejected for owner groups.

TODO: screenshots of the edit dialog not yet attached.
</details>

<details>

- Backend **984 passed**; frontend **82 passed**; `make ruff`, `make ty`, `tsc --noEmit` all clean.
- `tests/test_spa.py::test_missing_asset_returns_404_not_the_spa_shell` fails on this branch **and at its base commit**; it needs a frontend build. This branch touches neither `tests/test_spa.py` nor `api/app.py`.
- Every fix in this branch was mutation-verified: the fix was made inert, the covering test confirmed to fail, then restored. That includes the rename pre-flight (moving the check into the loop fails the overflow test), the indentation handling on both sides, and the guard against structural edits.
- A differential fuzz of 940 inputs compared the Python and TypeScript trimming; a 7-case HTTP round trip (store → GET → seed → submit → validate → store → rename → store) confirmed free text is byte-identical throughout.

**Reviewer-facing note on one gap:** SQLite does not enforce the description column width, so an overflow stores silently in tests and would surface as a 500 only on PostgreSQL. The two overflow guards are therefore load-bearing in production in a way the suite alone would not demonstrate.
</details>

# Guidance for Reviewers
Read file-by-file; the 14 commits include review-fix passes and do not reflect final state cleanly.

Three things most merit a second opinion:

1. **`api/routers/apps.py`** — the rename pre-flight's position. It is correct only because it runs before the loop; the reasoning is in the comment and I would like it checked.
2. **`api/routers/groups.py`** — the owner-group guard compares structural fields by *value* against `original_app_id` captured at load, not against the live attribute, because an earlier block in the same handler mutates `group.app_id` in memory first. An early draft compared against the live value and would have silently permitted rebinding an owner group to another app.
3. **`api/operations/create_app.py`** — promotion of a pre-existing group now reformats its description. This fixes a silent, permanent failure in the in-repo Google lifecycle plugin: its adoption branch fired on an empty owner-group description, the new constraint raised, a failing hook rolls back the transaction, and the error was logged and swallowed, so the group would never adopt and would retry identically forever.

Accepted and not fixed, all deliberate: `rstrip`/`trimEnd` disagree on a few exotic Unicode whitespace characters (cosmetic, no rejection possible either way); `ModifyGroupDetails` does not trim, so a stored value is not canonical until a later rename normalizes it; a combined rename-plus-description call issues two identical `App` lookups; and the form now submits a composed description on every owner-group save, so a tags-only edit issues an Okta update it previously did not and silently reseats a divergent description onto the base line.

### DIFF
--- a/api/models/app_group.py
+++ b/api/models/app_group.py
@@ -1,9 +1,24 @@
+import re
 from typing import List
 
 from sqlalchemy import func, or_, select
 
 from api.extensions import db
 from api.models.core_models import App, AppGroup, OktaGroup, OktaUser, OktaUserGroupMember
+
+_LEADING_BLANK_LINES_RE = re.compile(r"^(?:[ \t]*\n)+")
+
+
+def _trim_free_text(text: str) -> str:
+    """Strip leading blank lines and trailing whitespace, preserving indentation on the first line of real content.
+
+    Leading spaces/tabs on a non-blank line are meaningful markdown (an indented list item or
+    code block), so unlike `str.strip()` this does not treat a run of leading whitespace and
+    newlines as one unit to discard -- only whole blank lines ahead of the first real content are
+    removed. Trailing whitespace carries no such meaning and is stripped in full. Mirrors
+    `firstParagraph`'s leading-blank-line handling in `src/components/MarkdownDescription.tsx`.
+    """
+    return _LEADING_BLANK_LINES_RE.sub("", text).rstrip()
 
 
 async def get_app_managers(app_id: str) -> List[OktaUser]:
@@ -91,7 +106,7 @@ def app_owners_group_description(app_name: str, additional_description: str | No
         The base line alone, or the base line, a blank line, and the free text.
     """
     base = f"Owners of the {app_name} application"
-    additional = (additional_description or "").strip()
+    additional = _trim_free_text(additional_description or "")
     if not additional:
         return base
     return f"{base}\n\n{additional}"
@@ -118,5 +133,5 @@ def app_owners_group_description_remainder(description: str, app_name: str) -> s
     if normalized.strip() == base:
         return ""
     if normalized.startswith(base):
-        return normalized[len(base) :].strip()
-    return normalized.strip()
+        return _trim_free_text(normalized[len(base) :])
+    return _trim_free_text(normalized)
--- a/api/models/app_group.py
+++ b/api/models/app_group.py
@@ -19,6 +19,10 @@ def _trim_free_text(text: str) -> str:
     `firstParagraph`'s leading-blank-line handling in `src/components/MarkdownDescription.tsx`.
 
     `\\r\\n` is normalized to `\\n` first, since a browser textarea submits CRLF line endings.
+
+    Mirrored in the frontend by `trimFreeText`, a private helper in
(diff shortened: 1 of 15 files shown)


barborico:
Closing in favour of a much smaller design.

Rather than keeping the boilerplate as an enforced invariant (validation in `ModifyGroupDetails`, a reseat rule shared by the form and the rename path, overflow pre-flights, and a cross-language pin test), we're making app owner group descriptions **fully editable like any other group**, with the boilerplate as the seeded default at creation.

That deletes most of this PR and removes a whole bug class with it: the 1024-character overflow only existed because a rename recomposed and could *grow* a stored string. With no compose step, the seeded value is bounded at ~281 characters and nothing can overflow.

Two genuine pre-existing bug fixes from this branch are being carried forward into the replacement:
- the rebind guard being defeated by an earlier in-memory mutation of `group.app_id`
- `plugin_data` edits on owner groups returning 200 while silently discarding the write

PR #644, opened 2026-09-09
Add a CLI command to prune redundant direct access

barborico:
# Summary

Adds an `access prune-redundant-direct-access` CLI command that removes a user's direct grant to a group when a role that user holds already grants the same access. It reports without changing anything unless `--apply` is passed, and by default only removes a grant that cannot outlive the role coverage behind it.

**Urgency**: LOW - 2 weeks
**Expected review effort**: MEDIUM — 570 lines of production code (the rest is tests), but the command deletes access grants in bulk, so the selection rules and the safety defaults deserve real scrutiny.

# Motivation

This is an idea that's been mentioned infrequently for years, so here's an AI pass.

Access exists to make role-based access the natural default: a role represents a job function, and attaching it to a group grants that group's access to everyone in the role. A direct grant covering access a role already provides works against that in two ways. It obscures *why* a user has access, which is the transparency the role model exists to provide. And it survives the user leaving the role, so role membership stops being a reliable lever for granting and revoking.

These grants accumulate naturally — a user is usually added directly first and joins the covering role later — and nothing today finds them. This command does, and removes them, leaving the role as the single legible source of the access.

<details>

**`api/redundant_access.py`** (new) holds the whole engine, following how `api/integrity.py` backs `fix-role-memberships`:

- `OktaUserGroupMember.role_group_map_id` is NULL for a direct grant and non-NULL for a role-derived one. A `(user, group, access type)` triple holding both active is the redundancy. One join pairs the two sides, so the database returns only genuine candidates along with both end dates.
- End dates aggregate per triple rather than per row, because `ModifyGroupUsers` ends *every* direct row for a triple in one statement — the decision has to be all-or-nothing or acting on one row could destroy a sibling that would have failed the guard. Aggregation happens in Python because `bool_or` is Postgres-only and SQLite's `max()` drops the `DateTime` type.
- The **duration guard** removes a direct grant only when it ends no later than the role coverage, reading `NULL` as indefinite. `--allow-shortening` lifts it.
- Filters resolve to id sets before the sweep. Group and app filters union; user filters intersect with the result. An unresolvable value raises rather than narrowing to nothing, so a typo cannot read as a clean run.
- The driver removes one group at a time so a failed group cannot strand the ones after it, holding plain dataclasses rather than ORM instances across the loop — a rollback expires the identity map, and re-reading an expired attribute on an `AsyncSession` raises `MissingGreenlet`.

**`api/cli.py`** gains the Click wrapper; the callback parses flags, calls the module, prints, and sets the exit status.

**Writes route through `ModifyGroupUsers`** rather than a bulk `UPDATE`, which is what makes the command correct rather than merely convenient. That operation ends only rows with a NULL `role_group_map_id`; it calls Okta only when no other active row survives; it reports lost access to app-group-lifecycle plugins only when a user loses *all* access; and it emits the existing `GROUP_MODIFY_USER` audit event. Since the role-derived row always survives here, **no Okta call is made and no lifecycle hook fires** — the change is invisible outside Access's own records, and no new `EventType` is needed.

**Known gaps, deliberately left for follow-ups:**

- `--user`/`--group` resolution is LIKE-based (`or_(id == value, email.ilike(value))`), so `_` and `%` act as wildcards and a multi-match resolves to an arbitrary record. This matches the id-or-email idiom at ~10 existing call sites; this command is the first destructive consumer. A characterization test pins the behavior.
- No test pins the cross-module invariant that a `RoleGroup` can never be a `RoleGroupMap` target. This command's safety leans on it, and the consequence if it broke is severe, but the invariant lives in `ModifyRoleGroups` / `ModifyGroupType` and belongs in a test file for those.
- The driver's `except Exception: rollback()` would mis-report an already-committed removal as failed if `ModifyGroupUsers` ever raised after its first internal commit. No such path exists today.

**Also in this PR, per review feedback:** a `### CLI commands` README section documenting the rest of the `access` CLI, which the README previously named only in passing under Kubernetes cronjobs. Commands are grouped by when you reach for them — setup, scheduled jobs, repair and cleanup — with the operationally sharp edges called out: the two authoritative `sync` flags reverse the source of truth, `notify` covers three separate audiences, and `sync-app-groups` exits non-zero on a failed group. Landing it here rather than separately since the rebase already requires a fresh approval.
</details>

<details>

- 67 new tests in `tests/test_prune_redundant_direct_access.py`; full suite 1125 passing (1051 for `tests/` on PostgreSQL), re-run after the rebase onto main.
- Run against **both engines**: SQLite (default) and a native PostgreSQL 15.6 cluster, since CI runs both and this code was at risk of dialect-specific SQL.
- `ruff check`, `ruff format --check`, and `ty check` clean.
- Coverage includes: the duration guard at every boundary (both indefinite, one indefinite, equal dates, longer/shorter each way); candidate selection (expired on either side, ownership vs membership, multiple covering roles, multiple direct rows); filter resolution including soft-deleted records and the `None`-vs-empty-set contract; dry run performing no writes; the Okta client never being called even for a managed group; no lifecycle hook firing when pruning an app group with a plugin configured; the `GROUP_MODIFY_USER` audit event; the surviving role-derived row being left untouched; and per-group failure isolation.
- The `ilike` characterization test was verified to actually fail under exact matching, rather than passing vacuously.

Not run: `make pytest-postgres`, which needs Docker. The native cluster above is the equivalent.
</details>

# Guidance for Reviewers

Commit-by-commit — the history is ordered so each commit carries its own tests and builds on the last.

Most of all, is this the UX you'd want available to you as an Access admin? I'm thinking we'd probably wanna run it across all users and groups without allowing shortening, and then maybe even with that in some cases. A related idea is to support capping direct individual access separately from role-based access, under the assumption that the former goes stale sooner.

**Design choice: the dry-run-by-default inversion.** `fix-unmanaged-groups` and `fix-role-memberships` apply by default with an opt-in `--dry-run`; this command does the reverse. The reasoning is that those two perform idempotent repair toward a state the data model already implies, while this one deletes grants an operator may have meant to keep. Worth confirming you agree the inconsistency is justified.


README.md
```diff
@@ -406,6 +406,31 @@ access init <YOUR_OKTA_USER_EMAIL>
 
 Visit [http://localhost:3000/](http://localhost:3000/) to view your running version of Access!
 
+### Maintenance commands
```
somethingnew2-0, line 409: Don't want to scope creep this PR too much, but maybe we should document the other existing cli maintenance commands in the README.md as well?

somethingnew2-0 (dismissed):

README.md
barborico: Done!

api/redundant_access.py
somethingnew2-0, line 1: Instead of a file in the top-level api directory should this be a new operation? Seems like the single `prune_redundant_direct_access` entrypoint would fit the Operation pattern well

barborico: I'm only expecting this to get called via the CLI, and it seems like operations are mostly structured around operating on passed entities vs conducting big sweeps. So I was thinking of this as more similar to the other fixup functions, e.g. in integrity.py. What do you see as the advantages of refactoring into an operation?

somethingnew2-0: Adding a single-purpose file to the top-level api directory feels like unnecessary clutter. I think it would be best fit in the operations directory as operations themselves are written to be used in an interactive shell.

PR #647, opened 2026-09-09
Render MUI through Emotion so styled() interpolations may call hooks

somethingnew2-0:
Opening any Bulk Review dialog crashes the page with [React error #300](https://react.dev/errors/300), "Rendered fewer hooks than expected". The Data Grid is the trigger, but the cause is the style engine, so the fix is here rather than in the dialogs.

[`GridRootStyles`](https://github.com/mui/mui-x/blob/v7.29.13/packages/x-data-grid/src/components/containers/GridRootStyles.tsx) calls `useGridPrivateApiContext()` and `useGridSelector()` from *inside* its `styled()` interpolation, to decide one border radius. [Emotion](https://emotion.sh/docs/introduction) re-evaluates interpolations on every render, so the hook order holds. styled-components v6 memoizes the interpolation in a ref and serves the cached result whenever theme, stylesheet and props are unchanged, so on the second render those hooks are never called and React throws. The grid's own resize handling supplies that second render immediately after mount, which is why it crashes on open in a browser but never under jsdom.

Emotion is [Material UI's default engine](https://mui.com/material-ui/integrations/styled-components/) and the one MUI X is written against, so switching removes the mismatch instead of working around it. I scanned the dependency tree for other interpolations that call hooks: `GridRootStyles` is the only one across `@mui/material`, `@mui/base`, `@mui/lab`, `@mui/x-data-grid` and `@mui/x-date-pickers`. Nothing prevents the next one, though, and MUI X still ships this pattern on its latest major, so pinning the engine is worth more than patching the one call site.

**No application code changes.** Only four modules call `styled()`, all through `@mui/material/styles`, and none of styled-components' own API (`keyframes`, `css`, `createGlobalStyle`, `StyleSheetManager`, transient `$` props) is used anywhere. The theme augmentation in `src/mui.d.ts` already targets `@mui/material/styles`, which is engine-agnostic. There is no global CSS to fight over injection order, and nothing references a generated `sc-*` class name.

Dropping the alias also retires the Vitest `mainFields` override, which existed only to keep MUI on its ESM build so the `@mui/styled-engine` alias would apply. Under Emotion the CJS path resolves correctly on its own.

`src/styleEngine.test.tsx` pins the one property of the engine the app depends on for correctness. Nothing else in the suite would notice the engine changing, so a future alias could quietly reintroduce the crash. I verified the test earns its place by restoring the alias and reinstalling styled-components: it fails with exactly `Rendered fewer hooks than expected`.

**For review:** class-name generation and stylesheet injection order differ between the two engines, so a visual pass over the app is the real check here. The surface is small given the above, but it is not zero, and no test covers it. Bundle drops 6.2 kB raw and 1.9 kB gzipped.

Separately, this leaves `@mui/x-data-grid` at 16% of the bundle for three dialogs that use pagination and little else. Replacing it with the plain MUI `Table` the rest of the app already uses would save ~97 kB gzipped, but that is a UI change worth its own review and is not needed for correctness once this lands.

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -8,11 +8,12 @@
       "name": "access",
       "version": "1.0.0",
       "dependencies": {
+        "@emotion/react": "^11.14.0",
+        "@emotion/styled": "^11.14.1",
         "@mui/base": "^5.0.0-beta.28",
         "@mui/icons-material": "^6.5.0",
         "@mui/lab": "^6.0.1-beta.36",
         "@mui/material": "^6.5.0",
-        "@mui/styled-engine-sc": "^6.4.9",
         "@mui/x-data-grid": "^7.29.13",
         "@mui/x-date-pickers": "^7.29.4",
         "@sentry/react": "^8.34.0",
@@ -25,8 +26,7 @@
         "react-hook-form": "^7.51.2",
         "react-hook-form-mui": "^7.6.2",
         "react-markdown": "^10.1.0",
-        "react-router-dom": "^7.18.2",
-        "styled-components": "^6.1.13"
+        "react-router-dom": "^7.18.2"
       },
       "devDependencies": {
         "@openapi-codegen/cli": "^3.1.0",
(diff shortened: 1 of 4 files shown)


Jeff-Rowell (approved):
Added https://github.com/discord/access/pull/648 as a follow up

PR #653, opened 2026-09-09
Label managed Google groups as Workspace security groups

somethingnew2-0:
[Google Workspace security groups](https://knowledge.workspace.google.com/admin/groups/control-access-to-sensitive-data-with-security-groups) are the groups you are allowed to name in a policy that grants access to sensitive data. An operator running app groups through the example Google plugin has to go label them by hand in the
Admin console, and keep doing it for every new group. That is the kind of manual step the plugin
exists to remove, and skipping it means an app group that looks like it gates sensitive data cannot
actually be used to.

A security group is an ordinary Google Group carrying an [extra label](https://docs.cloud.google.com/identity/docs/groups#group_labels), so this adds a `security_group` group configuration and a `require_security_groups` app mandate, both reconciled by patching that label onto the linked group. The [plugin README](https://github.com/discord/access/blob/pcollins/google-groups-lifecycle-plugin-c00710/examples/plugins/app_group_lifecycle_google/README.md#security-groups) documents the behaviour.

## Two properties of the label drove the design

**It is one-way.** [Google never converts a security group back](https://docs.cloud.google.com/identity/docs/how-to/update-group-to-security-group),
so the configuration is a floor rather than a switch. Access raises the label and never attempts a
downgrade, and a group Google already reports as a security group has its configuration *restored*
to match, so the checkbox keeps agreeing with what Google will enforce instead of sitting unchecked
next to a group that is one regardless. This is also why the property is not declared `immutable`:
promoting an existing group is a legitimate operation Google supports, and `immutable` would block
it to prevent an un-set that is already a no-op.

The app mandate ORs with the group setting for the same reason: an app can only raise the
requirement. That lets a mandate cover groups predating it without rewriting each one's stored
config, and group validation rejects an explicit opt-out while it is set. Since nothing in the
plugin interface fires on an app configuration change, a newly-set mandate reaches existing groups
on their next `sync-app-groups` sweep or group update.

**Google enforces [membership requirements](https://docs.cloud.google.com/identity/docs/how-to/update-group-to-security-group),**
and rejects the label outright when they do not hold. That is recorded as a sync *error*, not a
skip, which is deliberately the louder of the plugin's two tiers: Access is otherwise displaying a
security group it has not actually got, so the divergence should fail the `sync-app-groups` run
even though the group's owner, not an admin, is usually who resolves it. Worth a reviewer's opinion
if you would rather it route quietly to owners instead.

## Reviewer notes

The label patch is a separate Groups API call from the display-name/description one, so a refusal
is attributable and does not block metadata converging. The patch replaces the label map rather
than merging into it, hence re-sending the group's existing labels alongside
`groups.discussion_forum` (dropping that one is how a group stops being a Google Group).

No frontend change: the generic plugin config form already renders `boolean` properties as a
checkbox and resolves `default_value`.

### DIFF
--- a/examples/plugins/app_group_lifecycle_google/plugin.py
+++ b/examples/plugins/app_group_lifecycle_google/plugin.py
@@ -293,7 +293,8 @@ class GoogleGroupManagerPlugin:
             "Label the linked Google group as a Workspace security group so it can grant access to "
             "sensitive data. Google does not convert a security group back to an ordinary group, so "
             "clearing this later has no effect and it is restored on the next sync. A security group "
-            "may contain only users, service accounts, and other security groups in this domain."
+            "may contain users and service accounts with Google accounts from any domain, plus other "
+            "security groups from this one."
         )
         if app_requires_security_groups:
             security_group_help = (
@@ -1025,10 +1026,10 @@ class GoogleGroupManagerPlugin:
 
         Google Workspace treats the label as one-way: a security group cannot be converted back to
         an ordinary Google Group. So the configuration is a floor, not a two-way switch -- Access
-        raises the label and never tries to lower it. A group Google already reports as a security
-        group has its Access-side configuration corrected to match, so the checkbox keeps telling
-        the truth about what Google will enforce, rather than sitting unchecked next to a group
-        that is a security group regardless.
+        raises the label and never tries to lower it. Once the group carries the label, this group's
+        own setting is written to match, so the checkbox keeps telling the truth about what Google
+        will enforce rather than sitting unchecked next to a group that is a security group
+        regardless.
 
         Args:
             ctx: The plugin capability context.
@@ -1041,10 +1042,15 @@ class GoogleGroupManagerPlugin:
         """
         labels = google_group.get("labels") or {}
         already_labeled = GOOGLE_GROUP_LABEL_SECURITY in labels
-        configured = self._get_configured_security_group(ctx, group)
+        # The group's own setting, read separately from the effective value below: whether to label
+        # is the OR of group and app, but the value that has to stay converged with Google is this
+        # one. Keying the writes on the effective value would leave a mandated group's stored False
+        # untouched forever, and the config form submits that stale False straight back into the
+        # validation error for opting out of the mandate.
+        opted_in_by_group = bool(ctx.get_config(group, CONFIG_SECURITY_GROUP, False))
 
         if already_labeled:
-            if not configured:
+            if not opted_in_by_group:
                 logger.info(
                     f"The Google group for {group.name} is already a security group, which Google cannot "
                     f"undo; restoring its '{CONFIG_SECURITY_GROUP}' configuration to match."
@@ -1052,7 +1058,7 @@ class GoogleGroupManagerPlugin:
                 ctx.set_config(group, CONFIG_SECURITY_GROUP, True)
             return None
 
-        if not configured:
+        if not self._get_configured_security_group(ctx, group):
             return None
 
         logger.info(f"Labeling the Google group for {group.name} as a Workspace security group...")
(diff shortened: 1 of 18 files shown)


examples/plugins/app_group_lifecycle_google/plugin.py
```diff
@@ -898,6 +1015,70 @@ async def _adopt_or_enforce(
...
+        already_labeled = GOOGLE_GROUP_LABEL_SECURITY in labels
+        configured = self._get_configured_security_group(ctx, group)
+
+        if already_labeled:
+            if not configured:
+                logger.info(
+                    f"The Google group for {group.name} is already a security group, which Google cannot "
+                    f"undo; restoring its '{CONFIG_SECURITY_GROUP}' configuration to match."
+                )
+                ctx.set_config(group, CONFIG_SECURITY_GROUP, True)
```
Jeff-Rowell: An app mandate makes `configured` true even when the stored group value is false or absent, so this never restores that stale value. The form then stays unchecked and submits a value that validation rejects on later edits. Could we inspect the group-level value separately while retaining the effective OR for the label decision below?

```suggestion
        configured_by_group = bool(ctx.get_config(group, CONFIG_SECURITY_GROUP, False))
        configured = self._get_configured_security_group(ctx, group)

        if already_labeled:
            if not configured_by_group:
                logger.info(
                    f"The Google group for {group.name} is already a security group, which Google cannot "
                    f"undo; restoring its '{CONFIG_SECURITY_GROUP}' configuration to match."
                )
                ctx.set_config(group, CONFIG_SECURITY_GROUP, True)
```

examples/plugins/app_group_lifecycle_google/plugin.py
```diff
@@ -898,6 +1015,70 @@ async def _adopt_or_enforce(
...
+            # security group it has not actually got, which is worth failing the batch sync over
+            # even though the group's owner (not an admin) is usually the one who can fix it.
+            reason = (e.reason or "").strip() or str(e)
+            return (
+                f"Google refused to label the linked Google group as a security group: {reason}. A security "
+                "group may contain only users, service accounts, and other security groups in "
+                f"{self._domain}, and only a Workspace Super Admin or Groups Admin may apply the label. "
+                f"Remove any other members from the group, or clear its '{CONFIG_SECURITY_GROUP}' "
+                "configuration to stop trying."
+            )
```
Jeff-Rowell, line 1088: Clearing the group setting cannot stop retries when the app mandate is enabled: the effective value remains true and validation rejects an explicit false. This wording keeps the remediation valid for either source of the requirement (and clarifies that only nested security groups must share the domain).

```suggestion
            return (
                f"Google refused to label the linked Google group as a security group: {reason}. Security groups may "
                f"contain users and service accounts, plus security groups from {self._domain}. Only a Workspace "
                "Super Admin or Groups Admin may apply the label. Fix the group's membership or caller permissions, "
                "or turn off the group or app configuration requiring the security label."
            )
```

Jeff-Rowell (changes requested):

examples/plugins/app_group_lifecycle_google/plugin.py
somethingnew2-0: Good catch, and it was worse than the checkbox display: the config form only substitutes `default_value` for an *absent* value, so a stored `false` came back unchecked and submitted straight into the opt-out validation error on an edit that never touched the field. Both writes now key on the group-level value (`opted_in_by_group`) while the label decision keeps the effective OR, and I also record it right after a successful patch so convergence does not wait for the next sweep. Four new assertions in `test_reconcile_labels_every_group_of_an_app_that_mandates_security_groups` and `test_reconcile_converges_a_mandated_group_that_google_already_labeled` fail against the old logic.

examples/plugins/app_group_lifecycle_google/plugin.py
somethingnew2-0: Both points taken. The message now names whichever of the group `security_group` or the app `require_security_groups` setting is asking for the label, since clearing the group one is a no-op under a mandate. And you are right on the domain scoping: per Google's [requirements](https://docs.cloud.google.com/identity/docs/how-to/update-group-to-security-group) only nested security groups are domain-bound, while users and service accounts need a Google account but no particular domain. I corrected that in the message, the group config help text, and the README, which all repeated the same wrong claim.

Jeff-Rowell (dismissed):

PR #662, opened 2026-09-22
Dismiss an action's label when it opens a dialog

barborico:
**Urgency**: none
**Expected review effort**: LOW — one small component and three call sites

## Motivation

Hovering the edit or delete icon on a tag, group or app shows its label. Clicking it opens a dialog, and the label stays: it hangs over the dialog for as long as the dialog is up, and is still there once the dialog closes.

image

## Cause

A tooltip closes when its anchor sees a `mouseleave`, and this anchor never does. The dialog covers it and marks the whole app root `aria-hidden`, so no pointer event reaches it again. Moving the mouse into the dialog does not help either: the anchor was already occluded when the pointer left it, so no `mouseleave` was ever generated.

Measured through one open-and-close, before the change:

| step | tooltips on screen | anchor |
|---|---|---|
| hover the icon | 1 | visible |
| click, dialog opens | 1 | inside `aria-hidden` subtree |
| move pointer into the dialog | 1 | inside `aria-hidden` subtree |
| Escape, dialog closes | 1 | visible |

## What's changing

`ActionTooltip` closes on mousedown and refuses to reopen while a dialog is present. Both halves are needed: the press focuses the button, and MUI treats a focused control as a reason to show its label, so closing alone puts it straight back.

Applied to the three headers that pair an icon with a dialog — tag, group and app. The audit links beside them keep their plain tooltips; they navigate rather than open anything, so their anchor is never covered.

The `PopperProps={moveTooltip}` offset each call site passed by hand moves into the component, which is why those local constants go away.

## A behaviour left alone

Once the dialog closes, focus returns to the button and its label appears again. That is ordinary MUI focus behaviour — the same thing you get tabbing to the control — and it clears on the next pointer move. Suppressing it would cost keyboard users the hint, so it stays.

## Validation of changes

Driven with real pointer events against a local instance, hovering and then pressing without moving the mouse, which is how a person opens the dialog and the state the label used to survive. After the change the count is **0** while the dialog is open, where every step above read 1.

Frontend suite passes (97 tests on this branch), `tsc` and prettier clean.


eguerrant (approved):
Comments are maybe a little verbose but lgtm otherwise

